PR #564, opened 2015-01-14
Fixing up the sheet notes in the json file.  URL to autocalc script was ...

DigitalPagan:
...not showing underscores. Escaped to fix, hopefully that works.  Also updated the damage bonus reference to include 1/2 db and fixed a spelling typo.


DigitalPagan:
Attn: admins, please verify that escaping underscores fixes URL display in site.  I don't have a way to test if that actually works myself.

PR #573, opened 2015-01-20
Update 2

CoalPoweredPuppet:
Correct sheet names; updated sheet; brought sheet title inline with
other nWoD sheets already on Github


SteveKoontz:
You're changes are conflicting with the changes I made to get the sheets working. You need to pull down the current version and apply your changes to that. On top of which you need to make sure when you're renaming your CSS/HTML/Preview files that you change the names to correctly match in your .json file.

CoalPoweredPuppet:
...ugh.  I understand what you are saying, and at the same time, its
confusing.

Don't do nuttin', I'll fix it.  You've helped more than you should have
anyhow.  Thanks for the update.

CoalPoweredPuppet:
So...little confused on a few things.  Exactly which sheet did I screw up?  Iron Kingdoms, Vampire the Requiem, or Eclipse phase?  Or is it all of them?

PR #577, opened 2015-01-22
Update charsheet_3-5.html

felipearmat:
Changed all camps with {value="class"} to {value="1"}, this can make the macros work with a binary answer to classskill checks.
Stopped doing changes on skillcheck macros at the diplomacy skill (line 972), must kepp it on until last skill.


SteveKoontz:
I'm not comfortable merging in changes to only part of the sheet. Also, I'm not sure about your interpretation of the rules. You should reach out to the sheet's origin author and compare notes.

PR #578, opened 2015-01-22
Fix character info section not saving

beeke01:
The items in the Character Info section of the sheet were not defined as type="text" and did not have the attr_ for each of the attributes.


SteveKoontz:
These changes are incompatible with other pull requests made to this sheet.

PR #579, opened 2015-01-23
Updated Character Info section

finderski:
The input fields in the Character Info section were missing the prefix "attr_". Added that to the fields so input would be saved.


SteveKoontz:
These changes are incompatible with other changes already merged into this sheet.

PR #581, opened 2015-01-27
Fixing Minor Issues

CoalPoweredPuppet:
Added powerfields
Added buttons for 'jacks
Added buttons NPC
Fixed Unarmed Combat weapons
Fixed NPC Power 
Minor Cosmetic Changes


SteveKoontz:
John, I know this must be getting frustrated but this isn't right still. You have to follow the instructions given on the top of the repository. 
https://github.com/Roll20/roll20-character-sheets/blob/master/README.md

For your Vampire sheet.json file, it doesn't pass the JSON validator
http://jsonlint.com/

None of your HTML, CSS, and preview files actually match the names in the sheet.json file. All of this will cause the sheet to not load onto the site.

Also, you need to be very careful with the characters you're using in your instructions. Symbols that look fine to you can throw errors. For example a ( ’ and ' ) are different characters and can cause errors in parsing the .json file. Also a tab and spaces aren't the same, tabs will cause errors.

Let me know if you have any questions.

PR #583, opened 2015-01-28
Create Minor Update to HTML

Acyd69:
I would like to propose this minor update to the Palladium Megaverse character sheet. The following is an itemized list of the updates/changes I entered.

-Changed the symbol after Megaverse from a diamonded questions mark to © (I think that is the most appropriate for copyrighted material)
-Added text inputs for Level and Experience on the first tab in the Character Statistics area. I think that text input will be best since the number input seems to be view limited at 3 characters and the text fields have a more pleasing look.
-Added a roller button on the Skill tab (fourth tab). This universal roller requests a skill name and rating and provides a success/failure result.

Thank you for your consideration of these changes to this character sheet.

### DIFF
--- a/Palladium Megaverse/Minor Update to HTML
+++ b/Palladium Megaverse/Minor Update to HTML
@@ -217,8 +217,8 @@ Palladium Megaverse© Character Sheet <br>
                 <tr>
                     <td><h5>Run</h5></td>
                     <td><input class="attribute" type="number" value="round(@{spd}x120/1760)" disabled="true" name="attr_run_mph" /></td>
-                    <td><input class="attribute" type="number" value="round(@{spd}x5/3)" disabled="true" name="attr_run_ft_melee" /></td>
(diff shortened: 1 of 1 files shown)


Acyd69:
I thought it might be useful if I annotated which lines the changes and additions occurred at.

© symbol can be found in lines 1 and 1089
Level and experience additions can be found at lines 33 through 40
Roller has been added at line 493

Thank you for your consideration

Acyd69:
More updates since nothing has processed yet.

- Added calculations to the values of the run blocks to save time. This can been seen in lines 219 through 223. 

P.S. I was unable to determine exactly what "Cruise" was so no calculations have been added yet.

Acyd69:
After talking with John W. I have realized that what I have done here is propose a complete rewrite of the character sheet HTML. I'm not sure if that is an issue or not. If it would be better for me to redo the pull request, please let me know here. Otherwise, I will leave things the way they are and wait to present any other changes/additions later.

PR #630, opened 2015-02-14
Added even more functionality - Microlite20

hylianux:
The changelog says it all.  I added more roll buttons, and also put restrictions in to add extra bonuses if you chose human as your race and/or chose fighter as your class.  I removed the spell section, since the rules of microlite state that you have access to ALL spells at your given level.  The only way to make use of that section would be to painstakingly add every single spell to the section, effectively making a really really long character sheet.  Not really all that useful.  There's a table in the rulebook that shows you how much each spell level costs, but I wasn't sure if I'd be allowed to include such a table in the character sheet.

Also added Saving Throw rolls, and added in the strength bonus to melee damage rolls.  Modified the attack rolls to include light weapon melee rolls (they use DEX instead of STR) and added the level modifier to each attack roll.

I think that's every possible roll button covered for Microlite20 :)

### DIFF
--- a/Microlite20/microlite20-test.html
+++ b/Microlite20/microlite20-test.html
@@ -26,8 +26,8 @@
                     </label>
                     <div class="col-sm-4">
                         <select class="form-control" name="attr_gender">
-                            <option value="1">Female</option>
-                            <option value="2">Male</option>
+                            <option value="1">Male</option>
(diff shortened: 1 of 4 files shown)


hylianux:
I won't be adding anymore commits.  Consider this a new milestone.  A baseline.

PR #651, opened 2015-02-26
Microlite20 - New layout and styling. Also new sheet for Toon

hylianux:
Overhauled the look of the sheet.  The old one looked plain, bare, and required a lot of scrolling.  The new one has grouping, columns, and a nice texture background.  None of the functionality was changed at all.


hylianux:
Lesson learned... Make a branch first before you do anything, then commit that branch.  Don't just work in the head branch all the time.  I made a new character sheet for Toon: The Cartoon Roleplaying Game, but couldn't get it on a separate pull request.  So It'll have to be included as part of this one.

PR #655, opened 2015-02-28
Adding New Character Sheet: Toon

hylianux:
This sheet is to be used with Toon: The Cartoon Roleplaying Game.

I accidentally included my changes to another sheet with this release (currently in a separate pull request), so I reverted it back (hence the additional commits and revert).  Still learning about branches in github :/


hylianux:
Have to close this out because this will remove my changes for another sheet.  Lesson learned... Branch first, THEN commit...

PR #657, opened 2015-03-03
Fugue 2.0 Changes

herrozerro:


SteveKoontz:
There is a merge conflict with your code. Somewhere in the process of your commits something changed and we didn't get the correct oder of changes. You'll need to save aside your changes, pull from the repository and then submit a new pull request with your final changes added back in.

PR #658, opened 2015-03-04
Created pt-br version of dnd sheet with translated macros and values.

danielfigueiredo:
Hello roll20 team, this new sheet is a translated and macro modified version of the D&D_3-5 sheet.
There are many colleges in Brazil that aren't familiar with English and having all D&D books translated in our country they hardly ever know the English terms. Everyone would be very happy if they can select this pt-br version.

I was invited to a table using the roll20 system and after I saw what you guys have done and, as a 6 + year experienced web developer, I have to say that its a very nice job and the product is very good.

This is a very easy pull since all files are new, I think I got the right roll20UserId, either way my subscription email is danielfigueiredoc@gmail.com

Let me know if there is something missing.

Best


SteveKoontz:
Sounds great. Thank you for doing the translation.

danielfigueiredo:
Good, will it be available at the website for anyone who creates a campaign?

SteveKoontz:
Yes, it will be available with the next update, which is a really big one, the Update of Holding.

PR #661, opened 2015-03-06
Update and rename Call_of_Cthulhu.html to Elric.html

Krometti:
An Elric version based on CoC


SteveKoontz:
I don't understand what you're trying to do here. I get that you want to make an Elric sheet but why would you delete the Call of Cthulhu sheet to do that and leave the new Elric sheet in the Call of Cthulhu folder. If you want to make a new sheet it needs to be in it's own folder, with it's own sheet.json file.

PR #692, opened 2015-03-14
added optional tie breaker for initiative based on PF's "Higher modifier wins"

mlenser:
This adds a an optional modifier so the initiative tracker can figure out who goes first in the case of a tie. It uses 3.X's system of:

"If two or more combatants have the same initiative check result, the combatants who are tied act in order of total initiative modifier (highest first)"

It takes the whole initiative and divides it by 100. So if you roll a 15 and your initiative mod is 3 your initiative is 15+3+(3/100) = 18.03. Negative makes it lower. The initiative tracker handles it all in the end.

I added it on the PC and NPC sheet both.


mlenser:
Meant to send this to Actoba. Will go through him.

PR #694, opened 2015-03-14
Align columns for Skills

mlenser:
Align skills so that the columns align across normal skills, custom skills, and untrained skills. Also made the Prof and Bonus/Misc fields slightly smaller font-size so the important info when rolling is bigger in comparison.

Also committed to Actoba's branch here: https://github.com/Actoba/roll20-character-sheets/pull/83

### DIFF
--- a/DnD_5e/DnD_5e.css
+++ b/DnD_5e/DnD_5e.css
@@ -312,7 +312,7 @@
 	font-weight: bold;
 	border-radius: 4px;
 
-	width: 79px;
+	width: 78px;
     height: 20px;
     cursor: pointer;	
 	position: relative;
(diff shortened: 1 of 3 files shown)


mlenser:
I intended to keep these 2 commits separate pull requests, but it automatically adds them..

SteveKoontz:
Seeing as how Actoba rejected these changes on his fork I'm not comfortable merging these changes in. You should reach out to Actoba and resolve whatever conceptual vision differences you two have for the direction of the sheet.

It's Roll20's policy to error towards the vision of the original sheet author. If you feel like your sheet is different enough to warrant being it's own sheet, stylistically different with a different vision (take Savage Worlds and the Savage Worlds Tabbed sheets for an example) let us know and we can consider a different take on the D&D5E sheet.

mlenser:
Seems he is unwilling to receive contribution. I'll do my own and work on the NPC parts.

PR #701, opened 2015-03-16
Microlite20 - Reordered positioning, and Cure for Walls of Text

hylianux:
I somehow missed adding a note to the initiative roll template explaining how it was calculated.


hylianux:
... Which then lead me to all sorts of UI changes... :D

PR #717, opened 2015-03-19
D&D 5e Shaped Character Sheet

mlenser:
Notable changes from Actoba's version are in the ReadMe.md.

See the screenshot for the visual changes.

I plan to expand probably to NPCs from here.


SteveKoontz:
For reference sake please change the folder name for this sheet to something like D&D_5e_Shaped so that it's easy to tell the difference between the existing sheet and the new take on it.

PR #725, opened 2015-03-23
D&D 5e Shaped Sheet

mlenser:
Renamed as requested.


SteveKoontz:
I'm a big fan of providing options to players, especially with a game as popular as 5E. I also like the look of what you've done, it's attractive. That said I have a question and a concern. 

You say you're going to keep the sheet running parallel to the existing sheet. Does that mean you intend to keep the sheets compatible (using the same attribute names)? I think it would be valuable if players could make an easy transition between sheets if they desired.

Secondly, I think Actoba should be credited as a sheet author since the code base is his.

mlenser:
1. I have kept nearly every attribute the same if  possible. I have renamed armor to match the dnd spelling and not the British spelling, so that would be lost. Besides that only the NPC attributes will change as I've mostly merged the PC and NPC sheet together. Where possible I've kept the existing attribute. If you have suggestions for how to do this better I'm all ears. I was thinking of writing a conversion script.

2. I agree with giving him author. I assumed he'd be against the idea based on our conversations.

mlenser:
An extra note on 1: over time it's guaranteed that Actoba's sheet and mine will differ. I'd love to work with him for things like naming, but that requires cooperation.

Plus it's difficult if we have different functionality. For instance I have range on ranged weapons and he doesn't. Or my ac auto calculates the highest and his doesn't.

Long story short: I will continuously pull bug fixes and new functionality from his sheet, though I don't think the same will be true in reverse.

SteveKoontz:
Actually Actoba was quite supportive of your efforts. He just has a different vision. I don't think it's out of the realm of possibility for you two to collaborate for the good of both sheets but I'll leave that up to you two.

PR #785, opened 2015-04-19
Simple Save button added

CoalPoweredPuppet:
Simple save button added per request, stats rolls added, added hit location roll to simple weapon roll, cleaned up roll template

### DIFF
--- a/Cyberpunk 2020/Cyberpunk2020.html
+++ b/Cyberpunk 2020/Cyberpunk2020.html
@@ -1,5 +1,5 @@
 <!-- Cyberpunk 2020 Character Sheet -->
-<!-- Made by Tony R. and edited by CoalPoweredPuppet -->
+<!-- Made by Tony R. -->
 <!-- With lots of help from Chris N. -->
 <!-- Inspired by Diana P.'s d&d 3.5 character sheet and Alicia G's Star Wars Saga Edition character sheet -->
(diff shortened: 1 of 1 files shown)


TonyRettich:
Hi, first off, thanks for beautifying this sheet! It looks sexy.

There is a change that you made to it that breaks the stats so they do not work as intended.

On page 103 of the Cyberpunk 2020 rulebook, it states “At Critical wound level, the character's REF, INT and CL are automatically reduced by half (round up).” and “If mortally wounded, the character's stats are reduced to 1/3rd normal (divide by 3, rounding up).” You currently have serious reducing Reflex by 2 (which it should do, but something it should stop doing once a character has moved on to critical), Critical reducing a character's Int, Ref and Cool by 2 (it should be reducing those stats by half) and mortal doing nothing (it should be reducing Int, Ref and Cool to 1/3rd)

In your previous edit, and this edit, you have changed it so that check boxes in the damage table affect base stats, instead of affecting the save, and made it so that the radio buttons in the damage table affect the saves, instead of affecting the base stats. I had the check boxes subtracting from the base save, and the radio buttons affecting the stats.  If you could leave this part of the sheet as it was, I would really appreciate it. Thanks!

CoalPoweredPuppet:
First: thank you.  I appreciate you taking the time to talk about this.

I am trying to figure out a way to get around the two extra wounds boxes in the serious level of damage.  Since you have balked at the change, I'm gonna assume my attempt was 'not good'.  I'll look into othe methods.

I am closing this pull request.  I will use the other damage track in my next pull request.

PR #800, opened 2015-04-25
(d&d5e_shaped) Add tooltips for skill checks (copied from basic rules)

Rouby:
There is probably some need to adjust stylings to fit in and rename the css-classes to fit a more general scheme

### DIFF
--- a/D&D_5e_Shaped/D&D_5e.css
+++ b/D&D_5e_Shaped/D&D_5e.css
@@ -3625,3 +3625,36 @@ input {
   max-width: 100%;
 }
 /* End custom Pure grid css (24 cols) */
+
+
+.sheet-row.sheet-row-relative
+{
+    position:relative;
+}
+
+.sheet-row-relative .sheet-tooltip-wrapper
+{
+    letter-spacing:0em;
+}
+
+.sheet-row-relative .sheet-tooltip-content
+{
+    width: 270px;
+    padding: 6px;
+}
+
+.sheet-tooltip-ul
+{
+    margin: 0 0 0 20px;
+}
+
+.sheet-tooltip-li
+{
+    line-height: 14px;
+}
+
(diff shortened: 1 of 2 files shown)


mlenser:
Hey Rouby,

It would be easier to merge these changes into my fork instead of putting it directly into roll20's base branch.
https://github.com/mlenser/roll20-character-sheets/tree/master/D%26D_5e_Shaped is my fork.

Helping add some documentation as tooltips is a great idea! Though I'm curious why we need a new class "sheet-row-relative". Did you make that to have the content be wider? We can do that by adding some widths on the tooltip class instead.

If you'd like to discuss this please pm me on Roll20 as Github does not have private messaging. We can chat on changes over roll20 or I can give you my skype/gchat. https://app.roll20.net/users/277007/mark

Thanks again for contributing. Please follow up. :)

Rouby:
Hi,

Just forked your repo, I will issue a pull request as soon as I solved the conflicts ;).

I looked to contact you but since Github has removed the private messages I didnt know exactly where to messasge you.

I created the class 'sheet-row-relative' because the tooltip-button needs to be in a relative-container (because the absolute positioning). I am not that familiar with all the other classes and maybe there is a better way to position the tooltip-button (I am sure there is, because right now it looks kinda squashed).

Also I would like to translate this sheet to german. What do you think would be the best approach to this?

mlenser:
Message me on roll20 so I can give you my info and we can discuss. I have already setup some tooltip examples for checks and skills that you can follow.

Translating: I would like to do this too. Roll20 doesn't support it very well, but we can do some css and html to show certain text. See https://app.roll20.net/forum/post/1837912/d-and-d-5e-slash-slash-spanish-character-sheet-hojas-de-personaje-en-espanol

I'd happily set an example up if you want to translate all the text into German.

PR #818, opened 2015-05-04
New Microlite20 Character Sheet - Beacon

hylianux:
This character sheet is a Microlite20 variant.  Here's the main home site for it: http://beacond20.blogspot.com/

This character sheet is intended to be placed under Microlite20, if possible (in the same way that multiple versions of Dungeons and Dragons all fit under the main Dungeons and Dragons category).


SteveKoontz:
Hylianaux, 

This sheet isn't very different from the existing Microlite20 sheet. Yours just has more fields. We normally only allow a 2nd sheet if the style is substantially different (take the 5E and 5E shaped sheets for example).

My suggestion would be to touch base with the original author about updating the existing sheet and submit a pull request to overwrite the existing sheet with your improved one.

hylianux:
I think the point of confusion here is that Beacon and Microlite20 are the same game, and they're not.

It may look like Microlite20, but that's because the rules are based on Microlite20, and expanded/changed (much like AD&D1 vs AD&D2).  There are SOME similarities, but the rules are different enough to the point where it needed its own sheet.  It may look similar, but I can assure you the calculations are quite different here.  I'd love to make a microlite20 sheet that implements the rules of all the variant games based on it (e.g. spylite, microlite74), but until these sheets enable me to have full javascript control, I'm sorta stuck :)

I also opted for the similar visual style because it helps with familiarity.  If you've played Microlite20, and wish to try Beacon for a different kind of game, it's helpful to know where everything is.  I've actually got pending changes to Microlite20 for this same tabular style as well.  I was going to check 'em in after this change got approved.

Does each game submitted into the character sheet repository have to look drastically different in design in order for it to be considered?

SteveKoontz:
I see your point. I was under the impression that this was just a different Microlite20 sheet. I'll go ahead and merge it in since you say that it handles different mechanics.

hylianux:
Awesome.  Thanks man :)

PR #826, opened 2015-05-12
Pathfinder-necros: Major update including pathfinder unchained

jarquafelmu:
* Finished implementing conditions that the original Sam had started.
* Implemented energy drain
* Implemented wound threshold
* Implemented nested inline rolling.
* Added tabination to skills, spell classes and spells
* Updated the footer to include myself, Vince and Nibrodooh as maintainers as well as updating the last updated date.
* Surrounded almost every case of a check or radio box and name with a label tag to facilitate clicking. I did not do this with skills because of the misc skills not having a name.
* Changed the styling of labels found underneath sheet-table-data classes to look like the normal text found in those classes. This allows for normal HTML best practice coding by allowing the name of a check or radio box to be selectable as well.
* Finished implementing the conditions for the NPC tab.
* Added nested inline support and tried to remove as many redundant nested inlines as possible to reduce calculation lag.
* Added the sheet-center class to wounds labels
* The diplomacy skill check didn't have the name attribute which caused it to fail when pulled to the macro quick bar.
* Added adventure, background and consolidated skill groups in addition to the 
* Added the new config page
* Added tabination to the skills, spells and spell class page. Should help with scrolling.
* Added a new skill tab to the skill section that contains all the skills in their original alphabetical format.
* Changed the unchecked version of the tabs to be a light blue instead of black.
* Changed the checked version of tabs to be black instead of light gray
* Moved the concentration portion of the spell class up a line to join the caster level


jarquafelmu:
Hey @SteveKoontz I updated my fork and made a new pull request

jarquafelmu:
no longer needed because Vince's pull request already merged my changes.

PR #859, opened 2015-05-24
NUD10 major update

allenmaher:
added roll templates, inline rolls, tabbed interface, autocalculations more info in the readme


allenmaher:
A patches to D&D BECMI seem to be included as well in this request... though it was not intended to be in there... the patch adds NPC fields and toggles to the sheet (it is ready for use).

PR #874, opened 2015-06-02
Update small error in the json file.

allenmaher:
Hopefully this is in better shape.  Not sure how I managed to get my fork so out of sync last time.


Rhallinan:
Looks good now! Not sure what happened either

PR #884, opened 2015-06-10
Roll templates

CoalPoweredPuppet:
Roll templates and slight adjustment to weapons area.  All names stayed the same so there should be no effect on current data.


CoalPoweredPuppet:
Kkyk in

PR #888, opened 2015-06-12
Adding my Shadowrun5e char sheet

Gerdofal:
This is a new addition to the list, my Shadowrun5e character sheet. Makes full use of roll templates, roll buttons, and as much automatic calculation as possible. I did my best in the json file, and I hope my markup of the instructions looks the way I hope it will look. Gerdofal (Chad on roll20)


Rhallinan:
I opened up our current sheet for shadowrun 5e and this is pretty similar to the one we have. We only want one community sheet per game unless a very different approach is taken. I'm not an expert on Shadowrun but can you explain the main differences between the two. Are these changes something that could be applied to the existing sheet?

Gerdofal:
It's based on the existing sheet. I contacted the author of the existing
sheet and then took it over. At first it was just tweaks but at this point
I've completely re-written it from the ground up to use roll-templates and
have many more useful roll buttons, which the existing sheet did not have.
I also made a ton of updates to fields and formulas to make it match 5th
edition rules more closely.

At this point it's pretty much a complete re-write, a massive expansion on
the existing. That's why I put it in as a separate sheet, the two are
different enough that characters that are currently in the old sheet would
end up needing a ton of manual changes before they would really work in the
new. In reality, this is a complete replacement for the old sheet, but it
wouldn't be a seamless transition. There were just too many things that
needed updates and tweaks.

Chad





*Chad Parker*
gerdofal@gmail.com

"The greater danger for most of us is not that our aim is too high and we
miss it, but that it is too low and we reach it." - Michelangelo

Rhallinan:
So what changes did you make that would be incompatible with the old one. What would prevent them from being implemented into the existing sheet? We love seeing new features and improvements for sheets but we try to keep them all on one sheet unless it is drastically different. Again, I'm not a shadowrun expert so I just want to make sure I understand what the differences are.

Gerdofal:
Hi there,

Well, it's not so much changes that are incompatible. I didn't change any
attribute names except in a couple cases where I absolutely had to due to
rules changes. The biggest two changes are the implementation of the new
game rule in 5th edition: "limits" (which limit the number of successes you
can have on a particular roll) and some standardization to how the skills
are setup. I also made a bunch of changes to the spells section to reduce
the insane number of rolls sometimes required for spell casting. All of
these would have the same impact on an existing character.

Let's take the active skills section. With the addition of limits and sheet
options to control those, none of the active skills would function until
those choices were made. So if I just updated the sheet, people wouldn't be
able to make any rolls using the skills section until they made the
drop-down choices for the skill choosing how the limit is calculated for
that particular skill.

The other changes are pretty much the same. Existing information from the
current sheet may carry over, but choices would need to be made and
additional information given before rolls would work. In spells, for
example, drain calculations would need to be entered so that drain could be
calculated for the spell.

My sheet should just be uploaded as a corrected and upgraded version of the
old sheet, but I was a little concerned that if we did that some people
might find they aren't ready for a game until they finish filling in the
new information. That would annoy me as a player and I didn't want others
to have to go thru that.

It would be different if there could be some warning to users of the old
sheet that the new one was changed, but I would personally hate to come
online for a 5pm game and find I needed to spend 30+ minutes updating my
sheet before I could roll using the sheet.

Rhallinan:
Thanks for the explanation. I'm adding your new sheet as our Shadowrun 5e sheet and will be keeping the old one but marking it as hidden so existing users can keep using it but new ones will use your new updated sheet. 

Again, thanks for explaining the differences to me. I just wanted to make sure I understood the differences before I pushed it out.

Gerdofal:
Thanks, It doesn't appear to have made it in yet, any eta? People on the
forums are asking.


*Chad Parker*
gerdofal@gmail.com

"The greater danger for most of us is not that our aim is too high and we
miss it, but that it is too low and we reach it." - Michelangelo

PR #891, opened 2015-06-18
Changed Unnatural Attributes to Bonus.

Craftomega:
40k rpgs have two systems for using characteristics; one uses the raw with misc modifiers. The other uses the characteristic bonuses (First Number). The problem is how this sheet is done does not take into consideration what those bonuses do. EX: Strength bonus for a Space Marine is Strength x2 +20. But it reads off the raw characteristic not allowing for modifiers of those bonuses. The changes here allow the player to input those bonuses and make the system that use them read off of that.

### DIFF
--- a/Deathwatch/Deathwatch.html
+++ b/Deathwatch/Deathwatch.html
@@ -113,9 +113,9 @@
                         <button name="roll_WS" type="roll" value="/me Weapon Skill Target [[(@{WeaponSkill}+?{Modifier|0})]] Roll: [[1d100]]"><label>Weapon Skill (WS)</label></button>
                     </div>
                     <div class="sheet-col">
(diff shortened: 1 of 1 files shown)


Rhallinan:
Are the name changes on attributes necessary? I'm worried it could mess up the sheet for people already using it.

Craftomega:
Yes, if someone else is to come in and look at the naming scheme it won’t make sense to have it as unnatural. Another coder might not make the connection that it is in fact the bonus, which is a very different concept. It might hurt some people but in the long run it will be beneficial.

Rhallinan:
Ok, I usually don't like unnecessary name changes but after reading up on Deathwatch a bit I can see where you are coming from on this one. I'll merge the changes through. The amount of people that will be affected should be pretty minimal

PR #896, opened 2015-06-23
Massive Update

alicia86:
Mainly formatting update to facilitate the inclusion of more show/hide options and all relevant fields.

Big change that will affect users is the conversion of the Defense Level/Armor fields into an autocalc with a show-able field to manually put in the math, (Level put in by default, which is good for most characters)


Rhallinan:
The change looks good. I'll pull the code in to test myself and will merge if everythign runs fine.

PR #937, opened 2015-07-14
fix hit location values in Dark Heresy 1e and Rogue Trader

jfredett:
The numbers in the paper doll are not correct. I suspect this is true of (at least) Deathwatch and the other 40K systems, but I don't have the sourcebooks for them.


Rhallinan:
Good catch. I only have the book for Dark Heresy and Rogue Trader too but i'll look into if it is the same.

PR #944, opened 2015-07-18
Typos Corrected

AndrewTBP:
I'm interested in forking this character sheet for the new Fantasy Age RPG. 
I felt I should correct a few typos first.


Rhallinan:
Sounds good. I haven't had a chance to look at Fantasy AGE yet but i've heard they are pretty similar so it shouldn't be too difficult.

PR #954, opened 2015-07-21
Stars Without Number -- Ship Sheet Implementation

jfredett:
I won't claim it's the best in the world, but it'll do.

In this PR:

* a multi-tab implementation
* a tab with a basic implementation of ship character sheets. Including:
    - All basic data (Name, Hull Type, AC, Armor, Crew Min/Max, HP, etc)
    - A section for additional notes
    - Repeating fields for Weapons, Defenses, and Fittings.
    - A second section for more notes (you can never have enough space for arbitrary notes)
* an updated changelog + sheet.json (I forgot to update it last time)

What it doesn't contain:

* Roll macros for any of the ship stuff (that would require some way of syncing the ship between character's, so you could use the right gunnery stat, etc). Fortunately, there isn't a great deal needed there besides damage; so it could be done.
* Great CSS. The repeating fields don't break with convention, and remain divs-full-of-divs. I think tables might make more sense for rigid layouts like these. I might refactor them when I go into refactoring the culture/profession/etc sections into repeating fields. I did the best I could, but I'm no CSS wizard.

Paging @kevinsearle -- What do you think about converting to tables for layout? I spent a fair while getting all the divs to line up, and they still aren't great.

Next up will be roll templates, I think.

### DIFF
--- a/Stars_Without_Number/ReadMe.md
+++ b/Stars_Without_Number/ReadMe.md
@@ -4,6 +4,21 @@ This sheet is created for use in Stars Without Number on Roll20. Based on the ch
 
 ### Changelog
 
+#### 0.0.7
+
+* Add ship character sheet tab
+
+#### 0.0.6
+
+* Roll buttons for all the listed skills
+* Roll buttons for all saves
+* Auto-sets the Max Strain to Unaltered CON
+* Boxes to store each Attribute's bonus value
+* A Box to store your current AC (to allow target-macros + for easy reference)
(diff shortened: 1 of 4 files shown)


kevinsearle:
Looks awesome! Wish I had time to have not let things lapse for so long.

While in theory I would want to avoid tables for layout, I think I've come around to using them for character sheets. The nested div hell is just too difficult to manage. So you've got my backing on that.

Rhallinan:
Glad both parties here like it. I'll merge it in and test it out.

PR #958, opened 2015-07-21
Fix #957 - Attack Rolls not working (SWN)

### LINKED ISSUE #957: Stars Without Number: Attack roll buttons not working
Reported from a user:

> you embedded roll buttons for attacking do not work. The rolls turn out to have waaay to many addition signs, just straight up breaking the roll.

jfredett:
This removes the `@{target|AC}` element, replacing it with a simple query, and adds some defaulting for the weapon AB and regular AB sections. Should resolve the problems people have been experiencing w/ rolls not working (if AB or Weapon AB was empty, then the roll wouldn't work).

Fixes #957


jfredett:
I folded this into #954, to avoid merge conflicts later.

PR #959, opened 2015-07-22
SWN: added credit to sheet author

kevinsearle:
What it says on the tin.

### DIFF
--- a/Stars_Without_Number/ReadMe.md
+++ b/Stars_Without_Number/ReadMe.md
@@ -4,6 +4,17 @@ This sheet is created for use in Stars Without Number on Roll20. Based on the ch
 
 ### Changelog
 
+#### 0.0.6
+
+* Added generic attack roll button.
+* Added attack and damage rolls for weapons.
+* Added initiative button.
+* Added roll buttons for skills.
+* Skills now default to -1.
+* Added AC field.
+* Added attribute bonus field.
+* Set maximum System Strain to unaltered Constitution.
+
 #### 0.0.5
 
 * Fixed Skill rows not aligning.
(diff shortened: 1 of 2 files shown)


jfredett:
@kevinsearle We've got 3 SWN related PRs open right now, 2 from me and this from you, I suspect we're going to run into some conflicts; Do you have a fork we can work off of, merge down all our changes, then submit one common PR so as to avoid conflicting? Right now I know this PR and my Ship-sheet one will collide. This one also doesn't update the version number on the sheet itself, by the looks of it.

kevinsearle:
My changes are based off the latest Roll20:master and should only conflict if you updated those meta-info files. I'm intentionally holding off on doing any more changes (although I don't have any in the pipeline) until your Ship update as I figure that's going to be a major one.

Are you including the switch to a table layout with this coming PR?

As far as the version number, it had already been upped to 0.0.6 but without a corresponding entry in the changelog. I thought that meant you did that in one of your updates but maybe it was a mistake on my part from an earlier change.

kevinsearle:
Ug, my bad. I didn't see you had added those changes to your PR.

jfredett:
No, I'm aiming for some roll templates next, then digging into the skill section (that's a big one to chew on, and right now the custom CSS I'm putting into my test campaign doesn't seem to be loading, and I'm getting errors in the chrome console; I suspect something to do w/ the interface overhaul).

I think I'm going to combine my two open PRs into one (the ship sheet PR + the attack button fixing PR), that'll make it so @Rhallinan just has to manage the one PR. Once that's done, hopefully it'll be a bit less confusing. I suspect @Rhallinan is probably busy with other things atm (I assume he's the same intern from the blogpost a few months back); so probably best to let things settle before charging in with more changes.

PR #960, opened 2015-07-22
Ziechael: Adding HeroQuest and Risus character sheets

Ziechael:


Rhallinan:
The html file for Risus is labeled as a .txt instead of .html. Once you fix that just update it and let me know and  I will be able to merge it in

Ziechael:
Thanks Ryan, that'll teach me to learn how to use github while in a meeting!

PR #961, opened 2015-07-23
first pass diaspora character sheet

drspin:
thinking of starting a diaspora/fate game on roll20 so... here's a super simple diaspora character sheet. :)


Rhallinan:
Simple but good. I'll test it out and merge it in.

PR #982, opened 2015-08-06
Implement roll-template for all roll buttons.

### LINKED ISSUE #979: Stars Without Number: Initiative not using correct die
From a user:

"Rolling for initiative off of the character sheet uses 1d10+dex, but the rulebook (at least the free version from DriveThruRPG) says it's 1d8+dex."

jfredett:
What it says on the tin, the templates aren't the prettiest, IMO, but I used some NASA starfield image in the background, and it looks a hell of a lot nicer than the bare `/me`

Also adds defaults for attribute bonuses.

Also fixes a bug in the initiative button. The wrong die was being used (1d10 instead of 1d8). Fixes #979 

Bumps the version to 0.0.8

### DIFF
--- a/Stars_Without_Number/ReadMe.md
+++ b/Stars_Without_Number/ReadMe.md
@@ -6,6 +6,7 @@ This sheet is created for use in Stars Without Number on Roll20. Based on the ch
 
 #### 0.0.8
 
+* Fix errant initiative die.
 * Adds Roll Templates for all the roll buttons on the sheet.
 * Sets a default of '0' for all attribute bonus fields.
 
--- a/Stars_Without_Number/ReadMe.md
+++ b/Stars_Without_Number/ReadMe.md
@@ -4,6 +4,11 @@ This sheet is created for use in Stars Without Number on Roll20. Based on the ch
 
 ### Changelog
 
+#### 0.0.8
+
+* Adds Roll Templates for all the roll buttons on the sheet.
+* Sets a default of '0' for all attribute bonus fields.
+
 #### 0.0.7
 
 * Add support for multiple tabs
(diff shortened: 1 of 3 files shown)


jfredett:
As an aside -- if I need to make some breaking changes to the sheet (stuff that will, for instance, potentially change the names of some attributes) is there a way to specify an attribute migration?

In particular, I'm looking at separating out the 'culture' and similar skills into a repeating sheet, but that's going to break any existing values. Obviously I don't want to break players sheets, is there a process for this already in place?

SteveKoontz:
There is no way to migrate attributes. Are these breaking changes necessary? Couldn't you have the default ones, limited in number in the skill list, then add a section for adding a repeating section if you want more below? That way it wouldn't break existing sheets?

jfredett:
Clarification - There are no breaking changes in this pull.

As for the breaking changes, I think I can avoid them, but wanted to know if I had to work around them. The skill section is pretty messy, in terms of UX and the underlying code (I suspect it was created prior to repeating sections). I can probably work around that, but it just makes things harder.

What's the procedure if, say, I wanted to offer a 'version 2' that people could use w/ the improvements (maybe with some note that the existing sheet is deprecated?)

Mostly I think the limited-in-number version is a bit nasty, for my campaign, I don't mind a customized sheet, but if everyone can benefit from my UX improvements, that'd be awesome.

SteveKoontz:
We'd be open to a overhauled version two of the sheet if it was substantially improved.

jfredett:
Okay, I'll keep working on my new version, the point may end up being moot if I can keep it all compatible. For now, this pull was the last real set of changes I wanted to get in, my players are going to love it. Thanks!

PR #983, opened 2015-08-06
Remade sheet for Mutant Year Zero from scratch

pinmissile:
Last sheet was pretty damn broken mostly due to improper attribute names, it was also rather sloppily made. Made a new sheet entirely, spoke with Steve K. about it, he had a look at it and told me it was cool.


SteveKoontz:
After reviewing how much is going to change with attribute names, I want to try this in a different way. I want to set this up as it's own sheet, in it's own folder. And we're going to phase out the old sheet for this one, but not overwrite it. That means for existing games, they'll still have access to the old sheet, but moving forward this will be the option. I'm going to close this request, go ahead and re-submit it in a new folder with whatever naming conventions or changes you want.

pinmissile:
Alright, let me just remove the code for the old sheet, pull request should be up in 10-ish minutes.

PR #1009, opened 2015-08-19
Paladium Rifts improvement/refactor

RonaldZ:
Refactored the Paladium-Rifts sheet, added/removed/moved several menus, and changed several auto-calculations.

Previously the sheet was missing a location for bionics, some currency values, combat values, saving throws, and a few misc. boxes. Also a few equations were incorrect. For instance the calculation for Skill bonus from IQ was the IQ-14. This works only if the player's IQ is between 16-30, which is the majority of the time, but fails when the player's IQ exits those bounds.

This refactor fixes these issues, as well as making the html more readable than previously. The tabs are not displayed sequentially instead of in reverse order nor are they part of successive labels. Attribute names were also extended to help users find values within the roll20 attribute section.


SteveKoontz:
Were the attribute name changes necessary? This will break those attributes on existing sheets.

RonaldZ:
I'd argue yes, but it's debate-able. Most of the attributes were given severely shortened names which made it difficult to figure out what the attributes correspond with. My group had a lot of problems setting up macros because of this.

Also, I wasn't sure how updates to character templates worked. I didn't see a way to update a sheet to a new template, and it didn't appear to happen automatically when I was previewing it in test games.

If updating the template would break already existing characters, I can revert the attribute names to their previous names to prevent that. Still though I'd still like to urge us to convert to attribute names to something that have an immediately clear meaning.

SteveKoontz:
I appreciate the desire to make the names more meaningful but I don't think that would outweigh the cost of breaking all the existing users sheets.

The sheet "template" as you say, is the master repository code that you're requesting to merge into. I'm going to close this pull request and look for your updated one without the attribute name changes. 

The only possible way to change all the attribute names at this point would be to have your new version as a separate and independent sheet. While we're not opposed to better sheets replacing older ones, the new sheet would have to show substantial improvements or design direction.

PR #1021, opened 2015-08-25
Moldvay Contest

amandasearle:


amandasearle:
Added Sheet Image, updated json

PR #1027, opened 2015-08-28
Paladium Rifts Refactor & Touch up

RonaldZ:
Did a major refactor of the Palladium-Rifts character sheet.

General:
* The sheet will now scale properly, so if the user expands the sheet the sheet will expand as well.
* Formulas were updated/removed for being inaccurate/wrong/homebrewed. This includes perception, IQ skill bonuses, Skill total percantages, Base Saving throws, Character speed, etc.
* The character overview can now be seen from every tab.

Combat:
* Weapons were moved into the combat section.
* There is now a seperate menu for ancient and modern weapons.
* Weapons now have an additional place for special bonuses.
* A second armor slot was added to the combat section.

Powers:
* Powers now have a SAVE attribute.

Did a pull request for the same changes in the past, but I had changed the attribute names to more verbose names. Reverted them per request. While the verbose names would be helpful to future/prospective sheet editors, they would also break the already existing character sheets.


richard087:
Hi, this is much appreciated. I tried a much more basic touch-up, but you've done a way better job :+1: . I think the call on the attribute names was the right one (please don't break my game).

If there's anything I can do to help. please let me know.

RonaldZ:
Heh, don't thank me on the attribute name thing. I was for changing the attributes to something more sensible. Credit for that goes to SteveKoontz. I didn't know that attributes updated with the sheets. Thought the sheets stayed static with the same html and css.

richard087:
Is there a box for SDC, or am I missing something? Do you just expect people to add SDC on to hit points?

RonaldZ:
It's just Damage Capacity at the moment. Creatures are either Mega-damage creatures or are Structure-Damage, never both. So the same box can be used for either. However, I can add in a small drop-down box so you can switch between SDC and MDC.... 

and all that above is moot because after going back over the sheet, it looks like I forgot to add it back in. One moment.

RonaldZ:
I've seen a few rounds of pull requested accepted now, what's keeping this one from being accepted?

richard087:
I'm a user of these. I've no idea why they haven't been accepted. I'm not
too clear what the "right" forum for this is.

PR #1057, opened 2015-09-11
Add variant setting support to Runequest sheet

DigitalPagan:


shasjas:
Not sure if this is the right place for requests but is there any chance for the ability to add more inventory slots?

DigitalPagan:
Hi,
Sure thing, I'll work that into the next patch I do.  I would have made it
a repeating section except I need to keep a set number to do the
encumbrance calculations.  How many slots do you and your players need?

Thanks for the feedback.

shasjas:
Hi thanks for replying,
Uh I think we need 24 though a few more might be useful as well. Obviously only if it doesn't mess with the sheet design.
Great sheet otherwise, thanks.

PR #1088, opened 2015-10-08
update from sohaho

Meteox:
sohaho's update missed the html file


Meteox:
ignore this, as i have seen sohaho already commited the missing file

PR #1102, opened 2015-10-15
Torchbearer French Translation

SionoiS:
French Translation of the original Torchbearer character sheet.


Rhallinan:
We don't support .scss files so i wont be able to merge this to main.

SionoiS:
It's an exact copy of the sheet already in the system. The only difference is the .html
I don't understand how it can't work.

PR #1106, opened 2015-10-18
Exalted 3e ability revision

binaryquinn:
added the performance ability, made a martial arts style section that functions like crafts

and resized the divs to fit this change


Alanith:
Ahh, nice work. I'm planning on trying something to fix the blasted radio buttons at some point, once i have the time.

PR #1117, opened 2015-10-25
Cypher System v1.3

NathaTerrien:
* Corrected roll with difficulty and inability: now compute the total difficulty
* Replace "Action" parameter inputs by drop-down lists, to ensure that Roll20 updates correctly the attributes
* Negative edge allowed (handicaps)
* Removed the word "target" from the attributes and template roll code to avoid confusion with the Roll20 keyword "Target", allowing to create custom macros


Rhallinan:
Will do. I won't merge this :P

NathaTerrien:
Thanks.
Just to be clear, I meant that just #41bc340 should not be merged. All the others should be.

PR #1125, opened 2015-10-28
Fate core with roll and some enhancements

sioc-de-narf:
New fate core character sheet with dice roll with or without Modifier pop-up, additional stress categories, # of stress box per categories,
stunts and extras as repeating fields with name and description separated.


Rhallinan:
Would it be possible to just merge the dice rolls with the main Fate sheet? We try to keep just 1 sheet unless there is a good reason to have more than one.

sioc-de-narf:
It could be possible, but a lot of things are missing in the current one, like : 
* the clarity for Stunts and Extras (currently just a 2 plain text areas)
* there are no way to add additional stress bars. (It is asked in a lot of hacks)
* the additional level 2 Stress box for 5+ in Physical and Will 
* there is no reminder of the number of stress box a character has.
* the Skill pyramid is too small for real campaign.

The current sheet has a Social Stress that is not in the basic rules... 

The current sheet was a nice basic first sheet but it's a poor one.

The only trouble there is that by changing from the current to mine, is the cuts and pastes to do from 2 fields into repeating fields.

PS: I am also going to post a French translation of mine.

sioc-de-narf:
Dman i did commit instead of pull request, sorry

sioc-de-narf:
I can edit the new sheet with a radio button or a tab so you can have both the ancient stunt / extra and the new one so nothing complicated for upgrade.

PR #1127, opened 2015-10-31
Added missing ability

BigFatFae:
Was missing performance, added it in, alphebetically


Rhallinan:
I believe it was merged in https://github.com/Roll20/roll20-character-sheets/pull/1106 in this merge. I just merged it in today so hopeuflly that covers that.

PR #1131, opened 2015-11-04
Update Palladium-Rifts.html

richard087:
Changed name of "attr_IQ_max" to  "attr_IQmax". Applied similar changes for MA, and PB

Justification for changing the attribute name:
Some bug in the formula replacement component of Roll20 doesn't evaluate the version with an underscore. These values are trivial to lookup in Palladium (a standard, long standing, table is used); and getting it working saves a lot of effort maintaining a character as it levels up.


RonaldZ:
Huh, did not know this was an issue. Thanks for the patch.

Rhallinan:
Yeah I saw this bug earlier this week. I'll merge this in and keep an eye out for wherever else I see it.

PR #1132, opened 2015-11-06
Mage 20 character sheet

mantonyz:
m20


SteveKoontz:
In the instructions section of the sheet.json file, you use symbols that break the .json. You need to strip those out. I'd suggest making a support wiki page and linking to it from the instructions sections. Here is an example of one such wiki:
https://wiki.roll20.net/13th_Age_Official_Sheet

Resubmit the pull request with changes to the .json

mantonyz:
Ok, json modified, wiki created and made a few other small tweaks.  Thanks for time and support.  Resubmitting shortly.

PR #1134, opened 2015-11-06
preview

Guitou31:
adding preview


SteveKoontz:
There were a couple of issues. The name of the preview image didn't match the .json. Also the .json file was mislabeled. It has to be "sheet.json" spelled exactly that way. I changed it for you.

PR #1153, opened 2015-11-20
Significant update - second roll template and config options

Gerdofal:
Changes in this version:
- Added GM only option and a second template (Thanks Neovatar for the idea)


Rhallinan:
I'm getting merge conflicts on this one. I'm updating a lot of character sheets so that might be the cause but I would try pulling and resubmitting this change

Rhallinan:
https://github.com/Roll20/roll20-character-sheets/pull/1154 I think it was due to this being here also. I was able to merge the linked one

Gerdofal:
Ok, I'll pull and re-upload the most up to date

PR #1188, opened 2015-12-09
PF-tabbed and ad&d1e

vince-roll20:


vince-roll20:
- updated sheet to work with the new repeating section features
- drag/drop and use repeating button name within another macro
- you must update existing macros that use repeating attributes to
include "$" before the rows name.  ie @{repeating_foo_$X} where X is the
row number.
- added a background toggle at the bottom of the sheet
- added a footer with sheet info
- added some addition sections that will get more enhancements in the
future (just ran out of time)
- added movement and encumbrance fields/attributes
- more to come...

PR #1199, opened 2015-12-15
Basically Version 1.0. Made some more fixes. Should be good to use now.

noahowings:
For Symbaroum


noahowings:
This is for Symbaroum, if that wasn't clear.

PR #1220, opened 2015-12-17
Pathfinder - Neceros

vince-roll20:
- fixed buffs column for Ref save not being applied
- updated Attention text
- partial/temp solution for Buffs indicator(top of the sheet) showing
"NaN" for new characters.  Just uncheck buff 1 from array to auto-calc
indicator.


rileydutton:
This has been merged

PR #1232, opened 2015-12-17
Pathfinder

vince-roll20:
- changed value="" to value="0" for attributes outside of <select>
- small formatting change
- hide wound threshold(unchained ruleset)by default


vince-roll20:
Pulled the latest version down and added some small changes.

PR #1233, opened 2015-12-17
Pathfinder

vince-roll20:
- updated formatting on version info
- date change in footer


vince-roll20:
You're too fast Riley.  ;-) 
But Don't stop.  Thanks

vince-roll20:
This is like a game.  Who can update faster...

plutosdad:
I merged these updates in the pathfinder-neceros.html file with my PR #1234 
So you can just use my pathfinder-necros.html file from that PR, and don't try to merge them yourselves, since we already did it.
Other files in 1233 here i did not touch.

PR #1256, opened 2015-12-22
Pathfinder - Simple

vince-roll20:
corrected the names of repeating attributes.  I mistakenly renamed them
when doing the update for repeating sections.


vince-roll20:
Request a quick merge on this since the mistake makes all repeating spells inaccessible.
Thanks

PR #1259, opened 2015-12-23
Pathfinder - Simple

vince-roll20:
- rolled back my changes to repeating attribute names (this was not
necessary)
- only titles of the repeating attributes have been changed to include
"$"
- kept previous change to damage mod selection and attention text


vince-roll20:
I resign...  ;-(
The repeating spells were still having issues because I renamed the repeating attribute names, but I only needed to remove the "$" from them.
This PR fixes that.
Sorry for the train wreck

PR #1260, opened 2015-12-24
This time it should be correct!

PapaRabe:
Thank you very much and Happy Holidays from Bavaria! :)


SteveKoontz:
Still had merge conflicts. You'll need to pull down from the repository before applying any changes. If you're having issues you can email me at steve@roll20.net. Happy holidays.

PR #1261, opened 2015-12-24
DD3.5-fr wis fix

Ammo283:
fix of the wis mod calcul


Ammo283:
The pull doesn't unclude Torchbearer update, it is just an error in the commits.

PR #1264, opened 2015-12-26
fix-up for unicode breacking the html code

sioc-de-narf:
fix-up for unicode breacking the html code


sioc-de-narf:
Hello,

Yes I fixed the same trouble on my 2 French sheets : Fate Core with roll VF and Tigres Volants.

Thanks to merge

PR #1265, opened 2015-12-27
Pathfinder 0.16

plutosdad:
0.15 and 0.16: skill page, acp and maxdex fix, buff total dmg fix, css divider cleanup,touch, cmd, check, class  buffs, vision


plutosdad:
this is ready

PR #1287, opened 2016-01-06
Ghostbusters expanded spanish

Noctropolitan:
Corrected .Json.
Expanded and translated into Spanish Ghostbusters Character ID


SteveKoontz:
Hey, the sheet.json file is still missing a comma. You can validate that the .json is valid from a site such as: http://jsonlint.com/

Also, are you intending to have different versions of the sheet, one a .html one a .htm. They appear to have different content. Could you confirm for me the sheet.json is referencing the correct file?

Noctropolitan:
My bad, the html is outdated, the htm is the correct one, the jsonlint.com
gives me an error

2016-01-06 19:00 GMT+01:00 SteveKoontz <notifications@github.com>:

Noctropolitan:
I added the comma and corrected the json to call the .htm, not the .html file. That should do it

PR #1296, opened 2016-01-08
Pathfinder - v19

vince-roll20:
- fix for class skill bug
- global whispers moved to inside macro-text so that emotes in
macro-text work
- fix for HP-formula recalc syntax error
- fix for total skill points not updating with a class level change


vince-roll20:
- merged another branch for the PF old/legacy sheet.
- new PR made

PR #1299, opened 2016-01-09
Pathfinder (old/legacy)  Corrected HP formula value to be included within "( )"

vince-roll20:
- Corrected HP formula value to be included within "( )"
"((@{CON-mod} * @{level}) + @{total-hp})"


vince-roll20:
Sheet users using the default value for HP formula were experiencing strange Max HP values.  Noticed that the HP formula value was not contained within closing parenthesis.  My last update only added a closing ")", but both should have been included.  ;-(
correct formula:  "**((@{CON-mod} * @{level}) + @{total-hp})**"

PR #1304, opened 2016-01-11
Pathfinder v19 and Pathifnder sheet Corrected

vince-roll20:
- accidentally saved over the pathfinder sheet with the pathfinder-neceros on my last merge.
- this is the correct neceros version and rolls back the pathfinder sheet.


vince-roll20:
thank you

PR #1314, opened 2016-01-14
New try. Fixed some minor issues.

PapaRabe:


pjbehr87:
I think the new image you uploaded was corrupted somehow, I'm not able to view the image. I'm also seeing a merge conflict. If you can re-save your preview image and handle the conflict then pull again, that would be great.

PR #1315, opened 2016-01-14
Master

PapaRabe:


pjbehr87:
Hey, this still has merge conflicts and the image is still giving me issues. You need to do a pull and fix the merge errors before you make the pull request. Also, I think the image that you currently have is corrupted, so if you could re-create it and re-add it to git that would be awesome. You can email me at phil@roll20.net if you have any questions.

PR #1317, opened 2016-01-16
Pathfinder v19

vince-roll20:
- minor formatting updates
- added "none" option to Init ability selector
- changed "Non-lethal DMG" to "Nonlethal DMG" to help a little with the
forced line break
- changed the size and default value of the class selector for repeating
class abilities and adjusted some column sizes to compensate
- changed default of Attack type to melee and Damage ability to STR
(these were set to none previously to force users to change them due to
repeating section limitations) None is still an option.
- added additional labels to NPC skills and enlarged some of the
"field-only" skills like craft, perform, profession, misc...


vince-roll20:
Cheers

PR #1319, opened 2016-01-16
Tenra charsheet fixes

mpureka:
Updated to add the "Extra" Vitality field to the max vitality total, and added a corresponding "Extra" soul field for the same purpose.

Also, removed spinner buttons from text fields in Mozilla, since that requires a different method than doing it in other browsers.


pjbehr87:
What is the MotW/MotW.zip that was committed?

mpureka:
Whoops; That's error and shouldn't be in there at all. Apologies, please do
NOT merge that. That's an entirely different character sheet and not ready
for primetime.

Feel free to reject the pull request and I'll sort this out.

PR #1324, opened 2016-01-19
Reducing broken decimals for 5e_Shaped inventory weights

azswengineer:
My players have complained quite a bit about the oddball repeating decimals that pop up in the inventory weights quite regularly.
The problem occurs because binary cannot represent most common decimals (e.g. 0.3) exactly, so a great many calculations produce oddball results. Simply doing all of the math as integer math in hundredths fixes this issue.
This patch changes the calculation by multiplying by 100 at the source of the inventory values, keeping coins an integer count until the final calculation, then dividing by 100 at the end.
I hope this significantly reduces the binary fraction errors that currently plague the inventory weight calculation. I haven't thoroughly tested this, but it seems to work in many cases (assuming I transcribed things correctly from the mingled sheet to the source components).
Ideally there would be a more sweeping change to just not use decimals at all, since this still ends up with some weird calculation errors, but it would require some visible changes to the sheet to do the more comprehensive change, and visible changes might cause issues for campaigns using the sheet.

### DIFF
--- a/D&D_5e_Shaped/precompiled/components/inventory/inventory.html
+++ b/D&D_5e_Shaped/precompiled/components/inventory/inventory.html
@@ -1,7 +1,7 @@
 <div class="sheet-inventory-row-{{num}}">
   <div class="sheet-shaped-row">
     <div class="sheet-col-05 sheet-vert-middle sheet-inventory-row-number">{{num}}</div>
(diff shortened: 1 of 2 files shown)


Alzam01:
I think it's better if you push this to mlenser's repo instead of directly to roll20 (so he can see your changes)

azswengineer:
Will do.  Wasn't sure the protocol.

PR #1327, opened 2016-01-20
Call of Cthulhu 7th Ed Chinese Traditional translation

ZongLong:
Call of Cthulhu 7th Ed Chinese Traditional translation.


pjbehr87:
It looks like the preview image you committed is still the english version. If you can go ahead and update the preview image, this will be good to go. Thanks!

PR #1328, opened 2016-01-20
Master

finderski:
Adding images to github directory for better image delivery


SteveKoontz:
Two of your images have a .html tacked onto the end. Is that what you meant?

PR #1337, opened 2016-01-22
Monsterhearts-Simple Sheet

lemadpierrot:
Initial commits for a simple Monsterhearts character sheet.


pjbehr87:
Thanks for the new character sheet!

PR #1342, opened 2016-01-24
Character sheet Oltréé! V1.0 - French

DiChim:
French character sheet for Oltréé! an rpg by Le Grümph


pjbehr87:
Hey, just FYI the file name case has to match between the json file and the actual file names. For example you have "Oltree" in the json, but the files are all named "oltree". I will fix it this time.

DiChim:
ok, I will carefull the next time. 
Thanks.

PR #1353, opened 2016-01-28
Dice Rule Fix + Extra Bruised Health Level

pensman:
This fixes the dice rules used in the buttons on the Character sheet.
Werewolf 20 does not use exploding dice (either for normal rolls or
specialties)

The buttons also neglected to use the 1s subtract from successes rule
standard to classic world of darkness (with the exception of damage
rolls). Buttons have been modified to correct this.

An extra Bruised health box has been added to accommodate users with the
Huge Size merit.


pensman:
This is my first time using Github and Roll20. I apologize for any mistakes in trying to submit this fix.

PR #1359, opened 2016-01-30
DungeonWorld character sheet modifier calculation tweak

aldergames:
Modified the Dungeon World sheet automatic modifiers to better match the table in the book. This fixes discrepancies for stats lower than 5, correctly setting the modifier to -2/-3.

For full calculations/demonstrations, see this gdoc:
https://docs.google.com/spreadsheets/d/1EnI0PmM7sgqAPYxGdS94O2jvUQhYQrTruRB1iXh0Cxo/edit#gid=0

### DIFF
--- a/DungeonWorld/DungeonWorld.html
+++ b/DungeonWorld/DungeonWorld.html
@@ -53,7 +53,7 @@
 		</div>
 
 		<div class="sheet-row-3">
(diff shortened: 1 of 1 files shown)


aldergames:
As a note for why this would be needed (as using the default set stats, it wouldn't apply, which is why I assume the simplified formula is used), consider that things during play might permanently lower your stats. There is also rolling for stats (3d6), which can result in lower stats.

PR #1363, opened 2016-01-30
AD&D1e - minor update

vince-roll20:
- made text clickable for section/sub-section collapsible titles
- fixed spells/day for spell levels 1-9 not being unique attribute names


vince-roll20:
Thanks

PR #1365, opened 2016-01-31
D&D 5e OGL NPC HP Hiding

m1sf17:
The NPC formula for rolling a NPC's health does not respect `wtype` and causes the result to always be printed global.  Added `wtype` to allow results to be whispered to GM.


pjbehr87:
Thanks for your help. Steve has already made this fix in dev and it will be in the next update.

PR #1368, opened 2016-02-02
Skill macro fix

jfrondeau:
Add missing ] on secondary roll on certain skills


pjbehr87:
Hey, thanks again for your work, but for now if you find an issue with the new 5e official sheet can you check the current forum post and post there if it's not already mentioned. Right now, while the sheet is still new, Steve is basically working on it full time, and it changing 1000s of lines of code throughout the week. While this is happening it will save everyone time if you just mention it in the thread and he'll fix it, if it's not already fixed. After a week or two, though, everything is fair game and you can go ahead pull request till your hearts content. (we still may not accept them, but that's how it always is) We really do appreciate all the help we get managing all of the sheets in the Roll20 catalogue, we just don't want you wasting your time while Steve is still in the primary development stage on this new sheet.

PR #1374, opened 2016-02-03
Added in a json file for the Leverage character sheet

tapiochre:
Sorry I missed that step previously.


tapiochre:
Ready to have this added back into main

PR #1380, opened 2016-02-06
Creating Urban Shadows sheets

tuxtradamus:
I have taken the currently existing sheets for Apocalypse World and adapted them to another game that is based on it, Urban Shadows. The goal for the sheet is to allow the user to only need to reference the PDFs at creation and advancement, then everything else they should need is on the sheet. The game has some moves per class that never change, and I have included them in the sheet to refer to, but custom/class moves are inputted in a repeating section.

Most of the changes are for fitting the content of the new game, the same overall "style" has been kept.  I changed some of the images to fit more of an urban theme as opposed to an apocalyptic feel. 

I changed the roll template to use the helper functions as opposed to having a condition for each possible case from a 2d6.


pjbehr87:
The new sheet looks good, thanks for all of your work!

tuxtradamus:
Of course! Thanks for the approval, can't wait to start using it.

PR #1382, opened 2016-02-07
D&D 3.5 Str check fix, add options for attack and save.

orjanSvandal:
Added size modifiers to strenght checks, because I think you add your size bonus to most (if not all) strenght checks.

Added customization on what ability to use on the different saving throws. These can be changed by using feats and classes. Added all abilities just to have the options available.

Added additional customization on what abilities to use when calculating attack and damage. You can switch to these abilities fairly easy when using spells and different classes.
Added extra option for power attack multiplier, set standard to 2. Now it only asks you once on how much power attack you want, the damage is then calculated correctly based on the power attack multiplier. And you do not have to write it in with a minus sign. If you do power attack 5 you now substract 5 to hit, and add (power attack multiplier) x 5 to damage.
Removed to question flank. If you have flank just type it into additional to hit bonuses. You do not ask for higher ground, or any other thing that makes it easier for you to hit, so why ask for flanking. I think it is better the less things you have to type in before each attack.

Note that I just started learning HTML a couple of days ago, and I am am very bad at making things look good on screen :) I will try this in my own games a while and do additional changes if I find anything else to modify.


orjanSvandal:
First time I have done this. Hope this works. 
Changes are 
strenght check macro
Ability for saves
Attack section. Same changes in all 10 weapon attacks. See my other comments.

orjanSvandal:
Closed since author disagrees with changes

PR #1383, opened 2016-02-07
Oltree-FR V1.1

DiChim:
Adding some informations to the sheet.


pjbehr87:
You've got some merge conflicts. If you could take care of these conflicts and open/reopen a new PR we would appreciate it.

PR #1385, opened 2016-02-08
IKRPG Update

CoalPoweredPuppet:
General cleaning; sheet is a direct copy of what I use in my games, and it works.  If something is broken in this one, I will do an line by line search for the issue.


CoalPoweredPuppet:
Thanks!

PR #1391, opened 2016-02-11
created shadowrun2e sheet

SileoNeco:


pjbehr87:
The styling for your new sheet is broken, the sheet is still usable and new so it won't mess up any games it progress, so I'm still going to add it to the approved list. You'll want to get your fixes done soon though so it will make it on the list in time for next week's pull.

screen shot 2016-02-18 at 2 20 17 pm

PR #1397, opened 2016-02-15
[DnD_5e] Split saves, skills, and attacks buttons into separate single and double d20 buttons.

DejaAugustine:
Also contains minor bug fixes for initiative button and attack buttons

### DIFF
--- a/DnD_5e/DnD_5e.css
+++ b/DnD_5e/DnD_5e.css
@@ -1,25 +1,25 @@
 
 .sheet-rolltemplate-5eDefault .sheet-rt-card {
-	width:95%;
-	background-color:#FFFFFF;
-	font-size: 11px;
-	
-	-webkit-border-radius: 5px;
-	-moz-border-radius: 5px;
-	border-radius: 5px;
-	
-	border: black 1px solid;
-	padding: 5px 0px;
-	
-	-webkit-box-sizing: border-box; 
-	-moz-box-sizing: border-box;   
-	box-sizing: border-box;         
-	
+    width:95%;
+  background-color:#FFFFFF;
+  font-size: 11px;
+  
+  -webkit-border-radius: 5px;
+  -moz-border-radius: 5px;
+  border-radius: 5px;
+  
+  border: black 1px solid;
+  padding: 5px 0px;
+  
+  -webkit-box-sizing: border-box; 
+  -moz-box-sizing: border-box;   
+  box-sizing: border-box;         
+  
 }
 
 /* Generic header */
 .sheet-rolltemplate-5eDefault .sheet-rt-header {
-	text-align: center;
+  text-align: center;
     position: relative;
     color: #fff;    
     margin: 0 -10px 10px -10px;
@@ -53,16 +53,16 @@
 }
 
 .sheet-rolltemplate-5eDefault .sheet-rt-title {
-	font-size: 1.5em;
+  font-size: 1.5em;
 }
 
 .sheet-rolltemplate-5eDefault .sheet-rt-subheader {
-	
+  
 }
 
 /* Melee header*/
 .sheet-rolltemplate-5eDefault .sheet-rt-header.sheet-weapon {
-	background-color: #2C5D28;
+  background-color: #2C5D28;
 }
 
 .sheet-rolltemplate-5eDefault .sheet-rt-header.sheet-weapon:before {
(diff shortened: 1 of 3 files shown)


Actoba:
Please dont merge this.  

There are updates coming that will allow users to choose whether there is a 2nd roll (for adv/disadv) without adding a 2nd button for all of them.  I understand that many users would like an option to disable the 2nd roll and until a few months ago this wasnt possible without the use of API scripting.  Now that it is possible I will be adding in functionality as part of a major sheet overhaul to incorporate newer features such as sheet workers and the SRD/compendium integration.  

I'm at work at the moment and unable to review the other changes so if you could let me know what the other bugfixes are i'll take a look....i'm not aware of any bugs within the sheet.

DejaAugustine:
@Actoba When is your overhaul expected to be released?

What is the harm in letting this oft-requested feature go out in the meantime?

Since the changes in this PR are largely cosmetic and don't impact any of the data/attributes, it won't break existing characters and shouldn't affect your overhaul.

Actoba:
Oft-requested is really dependent on your PoV.  I think oft-debated would be more accurate going by the feedback I have received but I completely agree that it's a hot-topic for sure...though i'm certain that 3d dice users (of which I am not) lie more on the side of wanting more control or to disable it altogether. I just disagree that the solution is to add more buttons thats all.  

I think I have a better solution in the works that will provide significantly more control for all users and tentatively plan to have update 1 of the revamp out in 7-10 days.  I'm trying to break down the work into smaller pieces and release more often whilst still ensuring that what does get released is fully formed/tested and has a minimal impact for transition...though some updates upcoming will undoubtedly have a bigger transitional impact than the sheet has had since it's inception due to the excellent updates from the roll20 guys recently and the now outdated workarounds/fudges the sheet currently uses to get around prior limitations.

DejaAugustine:
@actoba, I'm more than a little surprised that you are investing so much effort in blocking this.

This PR does not remove or change any of that existing functionality. Those who want to roll 2d20 for every check can still do so with the exact same behavior as before. If anything, this PR nullifies the controversy by offering a compromise.

Most importantly, however, is that unlike every other request for this feature out there, I'm not asking you to do any work on it. The work is done. Since it's all tracked in git, my changes can easily be reverted away so as not to interfere with your revamp.

For those who liked the sheet the way it was, this is literally a zero-impact feature.

After 20 years in the web and software industries, I know how easy it is for plans (especially tentative ones) to change, how quickly 1-2 weeks can become 3-4 months, or how the first update can end up not having all the features that were hoped for.

I agree that adding more buttons is not ideal. It's a workaround. But, even as a workaround, it offers an incremental improvement while we await your revamp.

I'm not asking you to do any work here, and I'm not trying to invalidate any work that you're doing. If anything, I'm buying you more time while providing a better overall experience for both sides of the 1d20/2d20 issue.

Actoba:
I'm not sure what you mean by investing so much effort.  I'm not sure how adding a comment here qualifies as such :(

Regardless, this isnt a purely cosmetic change.  The base function of the rolls right now is to roll 2d20, your change changes the base function of those rolls to only roll 1d20 instead, and, provide an alternative button to roll 2d20.  As the base function of the sheet has been well known for a long time the change forces people that are used to it operating it a certain way (and may have either used it or adapted their usage of it to match the bas behaviour) to change any macro bar/action bar's that they may have setup to accommodate a change that as I mentioned is debated anyway (anecdotally i would suggest it's a 50/50 split based of the feedback i have received).

As such the change is aimed at 50% of users, some of whom may already have made adjustments to use the base behaviour, and it will impact the other 50% negatively by removing a feature they desire and force them to make adjustments.

So in summary, i dont think the change is a good one either in the short or long term as it will likely have a much wider impact on user behaviour than may have initially been thought.

SteveKoontz:
Since Actoba is actively working on a different and more final solution to the advantage/normal roll feature we won't be merging in this request. That said feel free to offer it as a gist on the Character Sheet forum for users who want to use it until Actoba's update is ready. Also, I'd recommend you two having a conversation about the direction of the sheet and what ways you can both help without running into each other.

Actoba:
@DejaAugustine  - FYI - https://app.roll20.net/forum/post/2997169/5e-community-sheet-v1-dot-1-update-roll-options/?pageforid=2997169

PR #1400, opened 2016-02-16
AD&D1e - minor update

vince-roll20:
- set defaults for qty and carried to 0 and unchecked to help sheet
workers trigger  for auto-calcing weights
- set default carrying capacity limits based on STR 8-11.


vince-roll20:
Thanks Phil.

PR #1404, opened 2016-02-19
Initial draft of Dogs in the Vineyard Character Sheet.

smalone11:
Dogs in the Vineyard Character Sheet


smalone11:
Forgot to commit one other change.

PR #1408, opened 2016-02-20
Spelling Corrections in the approved.yaml

shdwjk:


pjbehr87:
The spelling error with NWOD Official is in the folder name, I will go ahead and rename the folder to be correct as well.

PR #1424, opened 2016-02-24
Add a 6th Ed Call of Cthulhu character sheet (French)

NathalieMinster:
Hi,

I made a character sheet for the 6th French edition of Call of Cthulhu. Unfortunately I do not know how to make a preview of this sheet to include in the PR. 
Also, I used in the sheet a logo from the French Editor website (to make the title) and I'm not 100% sure that it's ok for the copyright. Let me know what I should do about that. 

Thanks,


pjbehr87:
To create a preview, just use a cropped screenshot of the sheet. As for using the official logo, that's definitely something we can't do so I'm going to close this and you can open a new Pull Request with a preview and without the official images. Also, as far as images go, you can upload them to github as well and then you can reference their raw link for image in the sheet. This makes sure that a 3rd party site doesn't go and remove the image eventually.

NathalieMinster:
Thanks for your answer. I remove the official images.
About the preview I really don't see how to do it. I found something in the
wiki (https://wiki.roll20.net/Building_Character_Sheets) that suggests a
Preview Pane exists, but this option seem to be only accessible for
Pro-suscribers - which I'm not. With a little trick I managed to see what
the sheet should look like but I can't view it as it will eventually be (it
lacks the buttons "add/modify" that will be added by roll20). For this time
I will include the best first preview that I can, and update it as soon as
the sheet is integrated in roll20, unless you suggest otherwise.


2016-02-29 23:01 GMT+01:00 Phil <notifications@github.com>:

pjbehr87:
Since you do not have a Pro account, you can just ignore the preview image part for now. I will create one for you.

NathalieMinster:
Ok thanks, I create another pull request for these files as this one is closed now.

PR #1426, opened 2016-02-24
fate core with roll bug fix

sioc-de-narf:
Fix the sharing of repeating field between stunt and the dynamic skills.

Delete one won't delete anymore the other.


sioc-de-narf:
please merge asap

PR #1430, opened 2016-02-26
added a German translation of the The One Ring charactersheet 

uhu79:
based on the excellent work from Michael Heilmann
pls publish, thx :-)

br, uhu79


pjbehr87:
The image you are linking to for header image is uploaded to Roll20, this won't work for anyone other than you since those images are limited so only the uploader has access. You need to upload the image to github and reference the raw URL for the image. Since it is a new sheet, I will go ahead and leave it anyways, but you will want to fix this asap.

PR #1436, opened 2016-02-27
Fragged Empire Character Sheet

crash-cy:
New character sheet for Fragged Empire.  Contains all of the data from the Character and Combat sheets of the core rules.  Spacecraft and NPC sheets to follow.


pjbehr87:
You've committed these files to the root folder of the charsheets repo, please create a folder for them.

PR #1441, opened 2016-02-29
Pathfinder_simple

vince-roll20:
- In order to properly link HP for token bars and macros, @{HP|max} has
been changed to @{HPCurrent_max}. Link token bars to @{HPCurrent} to
display current vs max HP values. For existing characters, please update
your Total HP. Use the Attributes and Abilities tab of the character
journal to learn the value of @{HP|max}. Sorry for any inconvenience.

- Added carried coins section and included them in Total Weight and
encumbrance.

- The Critical damage roll has been changed from simply multiplying
damage based off of the crit multiplier to rolling damage once(w/mods),
and prompting for additional damage(w/mods), x2, x3, or x4 damage.

### DIFF
--- a/pathfinder_simple/pathfinder.html
+++ b/pathfinder_simple/pathfinder.html
@@ -1,12 +1,17 @@
 <div>
 <b style="font-size:14px; font-weight:bold; color:red;">ATTENTION:</b>
-    <input type="checkbox" class="sheet-section-show" title="section-show" name="attr_section-show" value="1" style="opacity:1;width: 16px;height: 16px;position: relative;top: 0px;left: 0px;margin: 0px;cursor: pointer;z-index: 1;" checked/><span></span>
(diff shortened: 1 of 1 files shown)


pjbehr87:
Sorry, this isn't going to work. We can't be losing everyone's HP for every character. If you want to make this change, you will need to create a sheet-worker to track the sheet version and automatically update the character's HP from the old values.

vince-roll20:
I'm going to have to do some js work?  but that's not my forte... ;-P
I get it.
Thanks Phil

pjbehr87:
Sorry :( This sheet system is just too popular to be losing one of the stats that is used on probably close to 100% of the sheets.

vince-roll20:
Not sure how this HP setup ever snuck by.  Currently players cannot link HP properly.  I'll see if I can figure out the js.

PR #1451, opened 2016-03-02
Master

Dragonfox257:
Added new content from expansion


Dragonfox257:
Everything except the latest commit should already be in the master branch. If not, I've been lagging behind in my commits.

pjbehr87:
There is definitely something up, you have merge conflicts. If you could fix those and open the PR again that would be great.

PR #1453, opened 2016-03-03
Updated to FFG Star Wars RPG API-Compatible to 4.0.3.12

KnowledgeRhino:
Added Roll Templates to Dice Rolls
Also changed display of species/career/specializations.

### DIFF
--- a/Star Wars FFG API-Compatible/FFG-SWRPG-CharacterSheet.css
+++ b/Star Wars FFG API-Compatible/FFG-SWRPG-CharacterSheet.css
@@ -1,6 +1,6 @@
 /*
 Version Info:
-Verion 4.0.2.18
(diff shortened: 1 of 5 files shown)


pjbehr87:
Since this is going to be replacing the current API Compatible sheet, we want to make sure it's possible to, as much as possible, seamlessly transition from the old one to the new one. From what I've seen the Character portion of the sheet does just fine, all of the data seems to be there. The issue I'm seeing is that the Vehicle section doesn't copy over at all. On the new sheet all of the data is totally blank when switching over from the old sheet. If you could fix this, by either refactoring the attributes to match the old sheet or to create a sheetworker in the new sheet that would copy any old data into the new attributes, we could finally enable this new sheet and depreciate the old one.

PR #1454, opened 2016-03-03
Fragged Empire Character Sheet

crash-cy:
This is a character sheet for Fragged Empire.  It includes the Character and Combat sheets as well as a Spacecraft sheet.


pjbehr87:
Just FYI, your "sheet.json" is named incorrectly, it needs to be actually named "sheet.json" (no quotes) and the file names inside of the json don't actually match the files you are uploading. I will fix them this time, because it's an easy fix, but please double-check this in the future if you create any more new sheets.

crash-cy:
Thanks for that, first time using Github so I'm still figuring this out.
Do you know how long it takes after merging for the character sheet to show
up in the character sheet list when creating a game?

PR #1455, opened 2016-03-04
Add STAR Marx

sioc-de-narf:
Fiche de personnage pour STAR MARX : le soviet des étoiles.

Character Sheet for the French game STAR MARX : le soviet des étoiles.


pjbehr87:
Your images aren't showing up, you'll need to set the src to reference the raw on github rather than referencing them locally.

sioc-de-narf:
https://github.com/Roll20/roll20-character-sheets/pull/1467

PR #1456, opened 2016-03-04
Fate

sioc-de-narf:
Update of the french sheet to match fate core with roll


sioc-de-narf:
Ok, looks like i succeed :)

PR #1457, opened 2016-03-04
PF v36

vince-roll20:
- fix for Domain/Sub-domains/School/Patron/ect not retaining values.
Added missing " from spell domains input
- removed min=0 from Misc field for Initiative to allow for negative
adjustments


vince-roll20:
Newer version up that includes these changes as well as a js fix Spell ranges not sticking,

PR #1460, opened 2016-03-05
Pathfinder 0.37 fix spell range

plutosdad:
fix spell range on recalc


plutosdad:
added in the changes from Vince's change here 
https://github.com/Roll20/roll20-character-sheets/pull/1457

so you can just merge this one and i'll tell Vince to close the other

rileydutton:
It's live.

PR #1461, opened 2016-03-05
PF_simple - HP|max fix, carried coin weight, crit damage roll query

vince-roll20:
- Total HP, @{HP|max} has been changed to @{HPCurrent_max} so that HP
can be properly linked with token bars. This conversion is handled
automatically using sheet workers to verify the version of the sheet and
updating @{HPCurrent_max} with the value of @{HP|max}. Once the sheet
has been updated, any changes made to @{HPCurrent_max} will also be
reflected on @{HP|max} for backward compatibility.

- Added a Carried coins area and they will be included in Total Weight and encumbrance.

- The Critical damage roll has been changed from simply multiplying
damage based off of the crit multiplier to rolling damage once(w/mods),
and prompting for additional damage(w/mods), x2, x3, or x4 damage.


vince-roll20:
thanks Phil

PR #1465, opened 2016-03-08
Updated to FFG Star Wars RPG API-Compatible to 4.0.4.0

KnowledgeRhino:
Added back in old Vehicle tabs to allow for a better transition to the
new character sheet format.


KnowledgeRhino:
Testing showed a problem, turned out it was a problem with my testing.

PR #1477, opened 2016-03-11
Fix Atomic Highway background

Lithl:


Lithl:
Not sure why GitHub is notifying of a conflict, since I updated from Roll20/master immediately before making changes, and only changed 1 line. However, #1476 is apparently changing the same line to link  http://www.gsa.europa.eu/sites/default/files/files/content/news/images/home/cracked-asphalt-closeup.jpg instead of the old image.

CoalPoweredPuppet:
Honestly, if it was up to me, I would say use Brian's update.

Lithl:
Well the question isn't "Use Brian's vs. use CPP's", since #1476 changes a lot more than just the background image. The question is more like "Use CPP's _and_ Brian's vs. use CPP's only".

pjbehr87:
You've got merge conflicts, it looks like you are changing the image that was just updated so, I think this is probably not needed. If you have a good reason to change it,, let me know once you've fixed the conflicts and re-opened the pull request.

CoalPoweredPuppet:
Uh...can you drop my changes and just use the other guys?

On Mon, Mar 14, 2016 at 7:43 PM, Phil <notifications@github.com> wrote:

> You've got merge conflicts, it looks like you are changing the image that
> was just updated so, I think this is probably not needed. If you have a
> good reason to change it,, let me know once you've fixed the conflicts and
> re-opened the pull request.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Roll20/roll20-character-sheets/pull/1477#issuecomment-196572828>
> .
>



-- 
“Just because you are paranoid, it doesn't mean they aren't out to get you.”

Never underestimate the power of Stupid People in groups.

PR #1483, opened 2016-03-12
Update M20.html

mantonyz:
fixed typo in extra ability 4

### DIFF
--- a/CWOD-M20/M20.html
+++ b/CWOD-M20/M20.html
@@ -424,12 +424,12 @@
                     <input type="radio" name="attr_Abi3" value="5" /><span></span>
                 </div>
                 <div class="sheet-row">
-                    <input type="radio" name="attr_bi4" value="0" class="zero" checked/><span></span>
-                    <input type="radio" name="attr_bi4" value="1" /><span></span>
-                    <input type="radio" name="attr_bi4" value="2" /><span></span>
-                    <input type="radio" name="attr_bi4" value="3" /><span></span>
-                    <input type="radio" name="attr_bi4" value="4" /><span></span>
-                    <input type="radio" name="attr_bi4" value="5" /><span></span>
+                    <input type="radio" name="attr_Abi4" value="0" class="zero" checked/><span></span>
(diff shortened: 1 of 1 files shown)


pjbehr87:
Is this typo causing issues, other than just having a misspelled attribute? Changing this will cause everyone that uses this sheet to lose that data since the the inputs will be referencing something that doesn't exist.

mantonyz:
This error causes the buttons in the Dice Pools section to not work if any
reference made to Abi4 in the pulldowns.  Alternately, I could change lines
1053,1119, 1185, 1257, and 1317 from

<option value="@{Abi4}">Extra Ability 4</option>

to

<option value="@{bi4}">Extra Ability 4</option>

I don't think that would cause the issue you mentioned.  Although, that one
being bi4 and the other Abi1, Abi2, and so on, will bother me, but I'll get
over it.

pjbehr87:
Since this is an existing sheet for a popular system, we can't change the attributes out from under players without also fixing it for them. Since there is a bug with it, we can't just leave it as is, so there are 2 options:
1. Change the dice pool rolls to reference the "typo'd" attrs (easy, messy)
2. Change the attr names and add a sheetworker to automatically copy the any values from the typo fields to the correct fields (less easy, not messy)

I'll leave it up to you. If you would like to do opt 2, but don't know how to do sheetworkers let me know and I'll give you a quick rundown of what you would need to do.

mantonyz:
I definitely prefer doing the sheetworker.  I have never done one, nor do I
know how to so any help you could give is appreciated.

pjbehr87:
If you copy and paste this at the end of the HTML for the sheet, it should take care of it for you: https://gist.github.com/pjbehr87/00faccfc6734079ee384

I'm going to close the PR, you can re-open it once you've made the changes. If you could also test it and make sure everything transfers correctly too I would appreciate it. Add a value in the old version, make your changes to the new, fixed, version and the value should still be there. As well as there not being issues with loading the sheet several times and changing the value on the fixed version.

mantonyz:
Will do thanks

PR #1488, opened 2016-03-14
Shaped 2.2.11

mlenser:


pjbehr87:
You've got some merge conflicts, if you could fix them and pull again I would appreciate it.

mlenser:
Any ideas why this isn't able to automatically merge? I can do so with Actoba's sheet (which is the parent of my fork) and have had no issues in the past going to the main repository in the past.

I spent several hours today trying to rebase my fork based on main, but in the end couldn't push it. I'm not sure if that would even solve the problem though as it looks like nothing has changed on Shaped or Reshaped.

mlenser:
I'm told that the merger can see the files that are preventing the merge. If so it would be of great help if you could inform me of them.

PR #1489, opened 2016-03-15
Tenra Bansho Zero sheet.

salmelo:
So I see now that there's already a TBZ sheet available on Roll20, but I swear it wasn't up when I started working on this a couple days ago. >.>
Anyway, after fiddling around with that sheet a little bit I think I do some things that it doesn't, and come closer to the look of the paper sheet, so I thought it might be worth submitting anyway, especially since the work is already done.

Apologies if I'm wasting anyone's time, and thank you for the great platform you provide.


pjbehr87:
Sorry, the other Tenra Bansho sheet has actually been around since last November. You say that your sheet does some things/some things better than the current sheet. If that is the case you can just go ahead and make those improvements to the current sheet. But, these minor differences do not warrant having 2 sheets for the same system. I'm sorry if you spent too much time working on this.

PR #1495, opened 2016-03-16
Roleplaying is Magic 4E sheet

Cazra:


pjbehr87:
Sheet looks good, should make it out on with the next character sheet push either today or tomorrow.

PR #1501, opened 2016-03-17
Pathfinder .38 debug off

plutosdad:
turn debug off
fix default value for 6 fields in npc page


plutosdad:
just going to upgrade to .40 instead

PR #1503, opened 2016-03-18
Added Apocalypse World 2e sheet version

brocktopus:
There are enough different classes with new moves/info fields, changed info for existing classes, etc. that I felt there was merit for a new/distinct sheet type (especially as some groups may continue playing 1e instead of 2e) from the original. I've kept credit ordered with the 1e sheet designers first, on whose work I've built.


pjbehr87:
Pretty much every system that has multiple editions has a separate sheet per edition, so players have the option to play whichever edition they want. I will go ahead and mark the current Apocalypse World as 1st Edition.

PR #1508, opened 2016-03-19
Additions to drain calculations

jflo:
I've added a prompt for a drain modifier, for PCs with centering foci or metamagics.
I've also added an input for initiate/submergence grade and included that in the drain/fade calculations.


pjbehr87:
If you haven't already, you will want to add the companion API scripts to our API Repo: https://github.com/Roll20/roll20-api-scripts
You might also want to mention in the game description that there are API scripts that go with this sheet.

tiggertazz:
Hello, I am looking to add some jpg's of Matrix Action/Program cards for Deckers. I would like to see if I can add an option to pull a jpg from a link  when I roll a specific Matrix action or program and throw it in the Chat with dice roll. I have attempted to go through the Attributes and Abilities page on the character sheet but I an having issues distinguishing what are the Matrix Actions and Program fields after they are created.

PR #1512, opened 2016-03-21
pathfinder .40

plutosdad:
upgrade to .40. new split columns, css for skills, fixes for encumbrance, html fixes, debug off


plutosdad:
upgrade to .40. new split columns, css for skills, fixes for encumbrance, html fixes, debug off

PR #1517, opened 2016-03-22
Pathfinder .40 css file fix to spells so they don't overwrite

plutosdad:
small Css fix can we get this in ASAP


rileydutton:
Should be live.

PR #1520, opened 2016-03-23
Pathfinder .41 fix buffs and shields

plutosdad:
fixed buffs missing, fixed shields applying to armor


plutosdad:
added css fix for pages to be be display:none instead of visibility:hidden when not checked, to avoid compendium copying over other pages.

PR #1523, opened 2016-03-23
Pathfinder .42 more tiny fixes to recent release

plutosdad:
fixed armor (again)
update weight to allow floats and only round at end
fix all spells tab


rileydutton:
You guys gotta stop submitting so many of these little ones :P

plutosdad:
That should be all until next week :) every new major set of changes people keep finding problems. darn people using the sheet

PR #1531, opened 2016-03-26
Added RWBY RoC sheet

Kasenai3:
Added RWBY RoC unofficial tabletop RPG ver2.14.16 sheet


pjbehr87:
In the sheet.json for the roll20userid property, it needs to be a list of user numbers rather than your name. I have updated the value myself to what I assume is your account. If this is incorrect please update it to the correct number. You can find this number by going to your public account page and taking the number from the URL.

pjbehr87:
Your files were using an encoding that wasn't compatible with our system, utf-16le, so I converted them into something that does, utf-8. So, if you are going to do more work, please be sure to pull down the new copy and make sure it is the correct encoding before you make any further changes.

Kasenai3:
Hello, yes it is indeed my account, thank you for updating the values and encoding, I'll be sure to use the correct ones when I work on other sheets or this sheet again.

PR #1542, opened 2016-03-29
W20 background fix

CoalPoweredPuppet:
W20 background fix


pjbehr87:
I notice you've been making a lot of these fixes. The preferred way to make sure the image is safe is to commit it to git with the rest of that sheet's files and then reference the raw link for that image on git. That way we maintain control of the image and we don't have to worry about a 3rd party removing it for whatever reason.

PR #1550, opened 2016-03-31
more fixes

mantonyz:
Attr 'corresondence' to 'correspondence' with sheetworker
lower DP1 roll missing closing bracket
science and politics swapped


mantonyz:
Thanks Phil

PR #1558, opened 2016-04-04
Add default roll options panel

DejaAugustine:
Permits the default roll style to be set sheet-wide, while allowing individual rolls to be configured.


Actoba:
Looks like a good addition.  Nice work

PR #1561, opened 2016-04-07
5e roll template updates

chibigoku04:
Unfortunately, I don’t have a method for testing. So, beware….

### DIFF
--- a/DnD_5e/DnD_5e.html
+++ b/DnD_5e/DnD_5e.html
@@ -33,11 +33,12 @@
           <div class="sheet-col-1-3 sheet-ro-optionlabel">Roll type</div>
           <div class="sheet-col-2-3 sheet-padl sheet-ro-option">
              <select name="attr_ro_default_rolltype">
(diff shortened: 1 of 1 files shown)


SteveKoontz:
What do you mean, you don't have a method for testing these changes?

chibigoku04:
I don't have a Pro account to test these out in the full Roll20 environment. If there is an available demo environment where I can test these changes at, please let me know. I'll be the first to admit I probably didn't dig deep enough into Roll20's developer space.

mlenser:
I believe Actoba already fixed these - I suggest picking them up from there.

chibigoku04:
I actually see very little change in @Actoba's changes from where it was. Again, I didn't test the changes, but I know the way it was caused "roll2" to incorrectly calculate the additional attack metrics.

pjbehr87:
You've got merge conflicts, and github is saying that you are changing every single line of the html which probably means your line-endings are different. Please fix these 2 things, and then I'll take a look at it.

PR #1570, opened 2016-04-12
Shaped 2.4.8

mlenser:


mlenser:
I'd also like to rename my folders to get rid of the "Reshaped" moniker to align with the companion script.
Perhaps "Shaped_old" and "Shaped"? Let me know if that works for you.

pjbehr87:
Yeah, you can go ahead and do that. We will just need to update the approved file to match the new folders.

PR #1573, opened 2016-04-15
Adding new Kromore character sheet with tabs and enhanced features. I…

tekas:
Adding new Kromore character sheet with tabs and enhanced features. I did not wish to remove the existing sheet because some groups may still be using it since the transfer is not 100% seamless. Equipment data does not transition over since the new sheet uses fieldsets and the old sheet did not.


pjbehr87:
We try not to have multiples of the same sheet, that do the same things. That being said, your sheet definitely appears to be an improvement over the original, so we will want to replace the original with your new sheet. But, in order to do that, we want to make the transition as easy as possible for anyone wanting to switch to the new sheet. Like you said there is at least one section that will not transition. You seem to be proficient with the sheetworkers, so this last upgrade should be pretty easy. The way I would do it is:
In a sheet:open function getAttr for all of the attributes that would appear using the old method. If any of these are defined, copy them into your new attributes and set the old values to an empty string (so it only happens the first time).

Once you do this we can depreciate the old sheet and make yours the new default. People using the old sheet will continue to use that sheet, unless they switch it manually, and any new people will only be able to choose your sheet.

PR #1591, opened 2016-04-22
Update Ex3

Lithl:
*Significant* changes to previous version. I have tried to keep the old attribute names where possible, but in some places that simply couldn't be done due to the new design (eg, moving to or away from repeating items).

New design closely mirrors the sheet in the official core book, instead of mirroring the Second Edition sheets. New version also fixes and improves all calculated values, plus adding more.

This sheet is best viewed with Cimiez Roman font installed. This can be achieved by either telling the browser to permit the page to load unsafe content (being served from my web domain, lithl.info), or [download the font](http://lithl.info/cimiez.ttf) and install it locally. The sheet will fallback to Arial if Cimiez is not available.


pjbehr87:
We need to add functionality to transfer data from the old system to the new system, especially since this is an update to an existing sheet. Adding new attributes is fine, but we try to avoid losing data as much as we possibly can.

The easiest way to do this is, in a sheet worker, on sheet open use a getAttrs to get all of the attrs that use the old system. If they exist: add their values to the new system, and blank the value from the old system (so it only triggers the first time). This way all the players using the sheet don't lose any data between sessions.

If you have any questions, you can respond here and I'll answer them as best I can.

Lithl:
Okay, just went through both files and started planning for this. There will still be 28 attributes which do not carry over, either because they're not a thing (the current sheet was created before the edition was finalized), they convey no actual information (@{exalt-type} means nothing in the face of @{caste}, for example), it's something that doesn't need to be referenced on a character sheet (essence/ability requirements for charms and prerequisite charms only take up space, since you aren't supposed to put the charm on your sheet in the first place unless you meet those requirements), or it was an editable field that's now autocalc.

That still leaves me 80 fields to copy over with sheetworkers. I should have a new PR with those changes soon.

pjbehr87:
Excellent, thanks for your work! I don't usually do PRs mid week, but in this case, if you finish before next weekend, you can drop a message here when you're done and I'll try and pull it right away.

Lithl:
Just created #1598 with a sheet worker script to migrate old values.

It also fixes a couple bugs I noticed, including a significant one related to z-ordering of some of the abilities. (The Crafting and Martial Arts sub-abilities in their popup div were fighting with the abilities located below them.)

PR #1598, opened 2016-04-26
Ex3 update

Lithl:
*Significant* changes to previous version. I have tried to keep the old attribute names where possible, but in some places that simply couldn't be done due to the new design (eg, moving to or away from repeating items). **A sheet worker script will migrate all old values to new values, where appropriate.** However, some values on the old version simply do not appear on this version, for reasons such as changes to the system since the creation of the previous version of the sheet, or redundancy.

New design closely mirrors the sheet in the official core book, instead of mirroring the Second Edition sheets. New version also fixes and improves all calculated values, plus adding more.

This sheet is best viewed with Cimiez Roman font installed. This can be achieved by either telling the browser to permit the page to load unsafe content (being served from my web domain, lithl.info), or [download the font](http://lithl.info/cimiez.ttf) and install it locally. The sheet will fallback to Arial if Cimiez is not available.


pjbehr87:
Wow, you work fast! I've merged the new version, but it probably won't go live till tomorrow. Thanks for doing the extra work to ensure the players have a smooth transition.

PR #1618, opened 2016-05-12
DH2ed Duplicate conditions removed

Mistwolf:
The sheet had two fields sharing the same value and that created duplicity. I renamed one of them and also did a small edit to the WS and BS roll for detecting weapon hits.


pjbehr87:
You have a merge conflict. Please fix the conflict and pull again. If you can do it soon I'll try to make sure new code still makes it out this week.

PR #1621, opened 2016-05-12
Fate_Accelerated_No_Approaches Character Sheet, css, preview image

doycet:
This is a character sheet for Fate Accelerated that uses no approaches, replacing them with rollable Aspects (which also function in the normal Fate way).

Directory includes html, css, and preview image.


pjbehr87:
Your new sheet is missing the sheet.json. I'm going to add one this time, but if you make any new sheets in the future please add this yourself.

PR #1625, opened 2016-05-15
Fate Core sheet edits

Maetco:
1.2
- Add Notes text box at the end.

1.3
- Put stress and consequences sections under skill section.

1.4
- Make the sheet as wide as reasonable so two can be shown next to each
other on most monitors. Set to 850 px.
- Add as many skill boxes as reasonable per skill level so they fit to
above with good layout. Set to 8 per skill level.
- Make the skill boxes as wide as reasonable with sheet width 850 px and
8 skill boxes per level. Set to 83 px.

1.5
- Add modifier to Mediocre skill roll too (a former mistake).
- Update pic and json.


pjbehr87:
Just FYI, case matters in the sheet.json. It's a super easy fix, so I'll make it myself, but in the future please make sure the file name case matches in all the references.

Maetco:
Thanks for fixin my error. Which reference was written incorrectly?

pjbehr87:
When you updated the preview image and the file name changed from fate.jpg -> fate.JPG. In the sheet.json it's looking for fate.jpg.

PR #1626, opened 2016-05-16
nights black agents character sheet

laresgod:
simple new sheet, no macro, no tab


pjbehr87:
I had to revert the changes you made to Trail of Cthulhu and the French version. For the eng version, you don't need to link the CSS files, if you list them correctly in the sheet.json the system will automatically pick them up. In the French version, your changes broke a bunch of the special characters on the sheet.

I had to make changes to the new sheet you submitted, but in the end it still went out. The css needs to be included in a separate file and all of the class names in the css need sheet- at the beginning, as this is added to all of the class names when the sheet is loaded in our system. I also updated the image references, they can't be referenced locally. You have to reference the raw version on github. You can see all of the changes I made in the most recent commits.

laresgod:
thank you for your kind efforts!
I changed on my local files what you advised me to do (renamed styles to
sheet-..., styles in separate css file, removed pictures/will reference
them properly later).

I tried to delete my files of original trails of chtulhu and french
version, but I still don't know how to redownload the original files,not to
commit bad files, so pardon me if you receive a "delete" pull request on
those.

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Garanti
sans virus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DDB4FAA8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

2016-05-17 2:17 GMT+02:00 Phil <notifications@github.com>:

pjbehr87:
Rather than make all the changes yourself, you should do a pull request to make sure you get everything. To undo any changes you have on a file, rather than deleting it, you can use "git checkout file/path/name.html" to return it to the HEAD version.

PR #1628, opened 2016-05-16
FAE_No_Approaches.json

doycet:
Fate Accelerated Character Sheet with Aspects replacing Approaches in addition to standard Approach usage. Adding missing json file.

### DIFF
--- a/Fate_Accelerated_No_Approaches/FAE_No_Approaches.json
+++ b/Fate_Accelerated_No_Approaches/FAE_No_Approaches.json
@@ -1,0 +1,9 @@
+
+{
+	"html": "FAE_No_Approaches.html",
+	"css": "FAE_No_Approaches.css",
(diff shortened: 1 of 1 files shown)


pjbehr87:
You beat me to it, thanks. Just one thing, that I will fix in 5 seconds, but this is the one file that needs a specific file name "sheet.json".

doycet:
Thanks for the info! This has been my first git-related project, so I appreciate the advice.

doycet:
Out of curiosity, how long does it take for these sheets to show up on Roll20?

Thanks again!

pjbehr87:
I _usually_ do pull requests on Monday and they _usually_ go out Tuesday morning. Neither of those are set in stone though.

doycet:
Great. Thanks!

PR #1630, opened 2016-05-17
removed my bad changes in trailes of chtulhu english + french + adapted my nights black agents folder

laresgod:
removed my pictures in the folder + html
-created sheet.css
-renamed styles to sheet-...
-transfered css styles from sheet.html to sheet.css

tried to remove my bad changed files in original trails of chthulhu and french version


pjbehr87:
There are a few things wrong with this:
You're deleting the images that the sheet uses, this will break the images on the sheet.
The sheets don't need the html, head, and body tags. Only the html inside of what wold be the body tag should be included.
You've moved the Trail of Cthulhu and the French version files all to the root directory, out of the folders they were in.

laresgod:
sorry for the moving out sheets that are not mine. Working on 2 different computers and different softwares do that + still trying to get used to github.
I will look at MY sheets (nights balck agents) and fix that. 
+
The taking out images was intended, but don't bother. I will have another look.

I want to reduce the size of the sheet, so I will post my latest version again and fix according to your advice.

PR #1634, opened 2016-05-18
Sheet and CSS changes

Project462:
Updated sheet and css to support optional 1E classes, cleaned up
formatting and changed displayed text.


pjbehr87:
Instead of replacing the old file with a new one, could you make these changes to the original file so that I can see the changes that are being made. Even if you copy the code from your new file over the old one, I will still be able to see the changes that are happening.

PR #1645, opened 2016-05-21
Master

suntzu777:
Harn Character Sheet


pjbehr87:
You are deleting the our README from the charsheets root. You are also adding a duplicate sheet.json to the charsheets root as well. Please fix these and create a new pull request.

PR #1658, opened 2016-05-25
Pathfinder .51 bugfixes

plutosdad:
fixes spells not working in macros
other fixes found yesterday
needs updating anyone who updates .50 will have broken spell templates.


plutosdad:
closed we need more testing

PR #1679, opened 2016-06-01
Convert to UTF-8 (second try)

Thalar:


pjbehr87:
The css and html file are both already in utf-8. Make sure to pull the most recent version. You have a merge conflict, pulling before you commit will also help avoid these in the future as well. Please fix the conflict and pull again.

Thalar:
Ok thx

Date: Mon, 6 Jun 2016 14:37:13 -0700

PR #1682, opened 2016-06-03
Shaped Translated

mlenser:
Hey Phil,

I've setup translations. So if you can add those to your system that'd be great.

I'll need to work on the class features part next, but this should cover the main sheet.


pjbehr87:
Looks great! Some stats on the specific translations:
(keep in mind this is indivitual words, not strings. You have a bunch of strings that are really long and that make these 2 very different statistics)
**Russian**: 1% translated
**Italian**: 100% (the system says 99%, but I think it's not recognizing words that are the same or don't have an It translation)
**French**: 100% (98%, same as above)
**German**: 6% (it looks like about 50% of the full strings are translated, but it's all the short ones, leaving all the things like the full-sentence class feature descriptions untranslated)

You can continue to have your personal translators make changes (this will definitely be the fastest option, unless you already have friends on the translation team :P), but if you can have them make changes to the files in the translations folder that will make it easier on me. Otherwise our volunteers will hopefully get to your sheet when they have a chance. I'm only just now seeing updates to the French version of our sheet. It's a pretty slow process, since like the sheet authors, they are also volunteers.

pjbehr87:
I will talk to the other devs tomorrow about the static value translation thing for selects and whatnot tomorrow. I will respond in the thread when we figure something out.

mlenser:
Ya, the French and Italian guys are on it. We'll see if the others pick up parts (though the parts that the German sheet is missing is mainly the class features bit).

I now realize that I renamed the "translations" folder to "i18n" to match what is common in web development and your verbiage for data-i18n. But that perhaps causes a problem. Should I revert that back?

And thanks for looking into the translations on a select, a span, and the roll template.

Thanks again Phil

pjbehr87:
I know i18n is usually the name of the folder and everything (that's what it is for the rest of the Roll20 site), but I want to keep everything as accessible as possible for brand new authors at a first glance at the repo. Everyone knows what a "translation" is, much less likely that they will know what i18n is. The system I have set up to manage/upload/download/read the translations looks for the "translations" folder in the root of the character sheet's folder.

I start using i18n for the actual tool and everything because if they are actually going to use the tool, they have either looked it up or will already recognize "i18n" at this point.

mlenser:
so revert back to translations?

pjbehr87:
Yes please.

mlenser:
Done.

pjbehr87:
Awesome, thanks!

pjbehr87:
Also, I posted about the new tool in the thread, in case you hadn't seen it.

mlenser:
Saw it, looks great!

PR #1688, opened 2016-06-06
Master

suntzu777:
does this work ?


pjbehr87:
These changes look great. It just doesn't include the mutant chronicles sheet.

suntzu777:
Well it seams your instructions don't work with a Mac , the browser will not allow a folder containing any files to be uploaded to GitHub only actual files which is why those files ended up in the root directory , I have deleted those files now.

 This time I selected the folder from my repository but your system would not comply until I went to the actual files within the sheet directory within my fork of the repository and selected those items  which now turns out not to have worked as no files were sent.

I am guessing but there maybe security issues and the Mac OS is blocking this activity

Rob

PR #1699, opened 2016-06-12
Shaped 4.3.0

mlenser:


mlenser:
Phil,
From what was written on the translation thread I was under the impression that if a language doesn't provide a translation then it will default to English. I assumed this was done via the lookup, not done via populating English into all of the missing translations.

Populating English into all of the missing translations makes it rather difficult to make changes without running a replace on the whole folder. It also makes it harder to determine which translations are missing as several translations match the English translation so it's impossible to determine if it's the same word or if it's missing.

pjbehr87:
Crowdin will handle any changes you make to the original translation.json automatically, this will handle adding and removing keys. The only think it won't do automatically is if you change a key.

PR #1704, opened 2016-06-15
Fate Accelerated - Ver. Kor & Special Edition

derornos:
Fate Accelerated
1. Ver. Kor
2. Special Edition - Overman Alliance


pjbehr87:
These sheets both look identical, minus the title at the top. Though, I do not speak Korean. Do we really need both versions of this sheet? How are they different?

derornos:
These sheets look similar. but different. (Both were formally published.)

Special Edition deals with the life of the Over-man like super-man, spider-man, etc.
Thus Over-man are to be considered to have two identities in the Sheet.
It is because that is the name of the other Aspects.

derornos:
You can see that the 'Fate' system, translated into Korean in the next.

Site : http://cympub.kr/%ED%8E%98%EC%9D%B4%ED%8A%B8-%EC%8B%9C%EC%8A%A4%ED%85%9C

Fate accelerated : http://cympub.kr/wp/wp-content/uploads/2014/09/%EA%B8%B0%EB%8F%99%ED%98%95%ED%8E%98%EC%9D%B4%ED%8A%B8-204x300.jpg

Special Edition : http://cympub.kr/wp/wp-content/uploads/2014/12/%EC%B4%88%EC%9D%B8%EB%8F%99%EB%A7%B9_%EC%95%9E%ED%91%9C%EC%A7%80-204x300.jpg

If you read all those books, you will probably feel that each sheet are required. XD

pjbehr87:
Okay, that's fine. I just wanted to make sure, we like to keep it 1 sheet per system/language if possible. But, since you said these are 2 separately published games, that's what I needed to know. (I just couldn't tell because I don't speak Korean)

derornos:
Thanks, Phil.

PR #1705, opened 2016-06-15
Update conan 2d20.html

suntzu777:
copy / paste works wonders especially when you use the wrong file !!

taken this opportunity to update a lot of the files and fix a few spelling mistakes

### DIFF
--- a/Infinity-2d20/Infinity 2d20.html
+++ b/Infinity-2d20/Infinity 2d20.html
@@ -488,7 +488,7 @@
 
     
 <div class="sheet-header">Weapons</div>
-<b>you need to create a rollable table called "Combat-Dice" with name values of "Hit","Hit+Hit","-","-","-","Special Effect"</b>
+<b>you need to create a rollable table called "Combat-Dice" with name values of "1","2","-","-","-","Special Effect"</b>
                     
                 
                 <div class="sheet-repitem">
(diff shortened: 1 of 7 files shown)


pjbehr87:
You have merge conflicts. I'm 90% sure it's because of the sheet.json files. I had to make changes to them to get them to work, but you don't have those changes. If you copy the text from the version in github and paste it over yours, then make your changes again, I think this will fix the merge conflict.
https://github.com/Roll20/roll20-character-sheets/blob/master/mutant_cronicles_3/sheet.json
https://github.com/Roll20/roll20-character-sheets/blob/master/conan-2d20/sheet.json
https://github.com/Roll20/roll20-character-sheets/blob/master/Infinity-2d20/sheet.json

suntzu777:
Ok Phil , the changes to the Json files were just a bit of text explaining about the rollable table that the character sheet needs , will sort that out and submit another pull request
Rob

pjbehr87:
Yeah, the issue was that I had also made changes to those files. When 2 different people make changes to the same file, Git gets confused and that's where the conflicts happen. The changes I made were to fix the JSON formatting. When you have quotes ( " ) in a string that uses quotes "This string has a quote " in it" the json breaks because it sees "This string has a quote " a the whole string and then a bunch of extra text at the end ' in it" ' and breaks. The way to fix this is to "escape" the quotes, or put a backslash in front of it "This string has a quote \" in it", this way the JSON knows the " in the middle is part of the string, and not the end. This is what I did for your JSONs. If you take my changes and replace your files with them, then commit, then make your changes that should fix the conflict.

PR #1710, opened 2016-06-16
fallout formula correction in html

laresgod:


pjbehr87:
For the Savage Worlds Fallout version, is that the exact same as the Savage Worlds sheet, but with a different look to it? If it is, could we possibly merge it into the original sheet? We like to keep 1 sheet per system/language if possible. If you could merge it, you can put a button at the top that would switch between the classic skin and the Fallout skin. That way, each player has a choice and we don't have 2 sheets that work the same, but just look different. If you don't know how to do this, let me know and I can give you some pointers.

laresgod:
It is not only a look change.
Some skills have been taken out,  others have been added.
The same for derived stats.
I suppose I could try to tab it on the original version,  but I don't know
how to do it yet.

laresgod:
I do not mind if you don't validate the Savage Worlds - fallout sheet right now since I can make do with the big tabbed savage worlds sheet (although i have to tweak EACH character on creation, enabling/disabling skills).
Right now it is a very similar copy of the original simple savage worlds sheet, but I do plan to modify it and make it unique.

However i direly need the fallout pnp 2.0 sheet, that at first sight "looks" like the already sheet there, but the formulas are modified and make for a different closer to video game feeling.

PR #1738, opened 2016-06-29
correction text boxes attr name

laresgod:
-renamed text anc check boxes to "attr_..."
-corrected typos
-changed text objects to textareas


pjbehr87:
Please update your branch or change your text editor to use Unix (\n or LF) rather than Windows (\r\n or CRLF) line endings. You keep committing code with different line-endings after I change it to what we need.

laresgod:
Ok.  Will do.

laresgod:
Can I ask in what file you found the problem, since I am trying to spot it?

2016-07-06 6:46 GMT+02:00 G. Ti. <laresgod@gmail.com>:

> Ok.  Will do.
>

laresgod:
What can I do to make ticked checkboxes remain ticked?
I want the checkboxes I click on to remain checked (unless I uncheck them later) when I close the sheet and the session. 
Is that possible?

PR #1757, opened 2016-07-09
Warhammer Fantasy Roleplay 2

Almos80:
I don't really know how this works.


pjbehr87:
Is this an update to the Warhammer FR 2 that we already have? Is it a new sheet? If it's a new sheet what are your justifications for having a second sheet for a game system we already have?

Your files are being added to the root directory. If they are modifications to the sheet we already have they should be committed to that folder. If it's a new sheet, assuming having a 2nd sheet for a system we already support is warranted, it should be committed to a new folder as well as needing a preview image and the sheet.json file that every sheet has.

Almos80:
I completely messed up when adding this and had to quit before being finished so I know there are missing files. I never figured out how to add it to a new folder. I'll attenpt to remove this and try again.

Is it appropriate to post a discussion about the sheet, describe it and the justifications for adding it, before attempting to upload? This is my first visit to this resource.

pjbehr87:
If you just post on the forum, I would probably miss it in all honesty. Your best bet would be to message me here. I can also help with the whole Pull Request process, if you need it. Just let me know what you're having issues with and I'll try to point you in the right direction.

PR #1760, opened 2016-07-10
Godbound v1.1.0

jfredett:
This is an update to the Godbound character sheet, among the changes are a conversion of the CSS from pure CSS to a SCSS -> CSS pipeline (makefile included). I've committed the rendered CSS so there's no change on your side, but it makes it a lot easier to maintain (and some other projects have done similar).

the README contains a changelog, but the summary is:

* Add tabs for Character, Faction, and NPC sheets
* Godbound and Faction project tracking sections
- Minor
   * Tweak spacing on Attributes, Saves tables.
   * Add upper border line for consistency
   * Use small-caps instead of all-caps for area titles

Let me know if there's anything you need from me to get this merged. I updated the screenshot.png to reflect the changes as well.


pjbehr87:
First, I would just like to thank you for level of professionalism with your commits. It's pretty amazing and makes my job so easy when reviewing these PRs. You are already one of my favorite authors now.

Since you are using "compiled" css now, it might be a good idea to minify the output css file. Not only does this make loading ever-so-slightly faster it will discourage anyone else, if anyone else wants to help with this sheet, from making changes to the output file rather than the scss files. Though, this is absolutely not a requirement, just a suggestion.

jfredett:
Aww shucks, I just try to keep things tidy. Glad to know it's appreciated.

I'll add minification in my next round. Easy to add. I'm hunting around for a "compiled" html alternative that isn't awful and does partials out of the box so I can do similar w/ the HTML. It's a 1500+ line pile of junk at the moment and I *know* there's duplication all over it that's bit me more than once now.

Next round is some more bugfixes and probably roll templates, see you in another week or so. :)

pjbehr87:
Check out the 5eShaped sheet, it uses compiled HTML and minification. @mlenser would probably have some tips and tricks for you as well. He's the only other author I know of that takes advantage of these tools.

mlenser:
I use gulp inject primarily. All my HTML files are nice and compartmentalised. It's fairly simple if you check out my gulp file and 5eShaped.html which has all the includes (which all have more includes)

Sheetworkers is all in 1 big file. :(

jfredett:
I looked over it, @mlenser, it looks interesting, but I'm wildly unfamiliar with the js ecosystem. I've been trying out using middleman (a ruby tool I have more familiarity with), since I already have scss, the ruby dependency isn't so bad.

Aside -- the 5eShaped sheet is *really* great, I've dug around in it for ideas/techniques, great work!

PR #1768, opened 2016-07-14
New character sheet for The Sprawl

timdenee:
A new character sheet for The Sprawl, with more options and fields and a more detailed UI.


pjbehr87:
Hey, a few things:
We already have a sheet for The Spawl, but you say this one has more options/fields. What exactly are you adding to the other sheet? Is there anything that the other sheet does, that this one does not. If there isn't, and you do a little of work to assure a smooth transition, we could replace the old sheet with this new one since it seems to have a lot more going for it.

I see that you are submitting the minified html and css. While this is good practice, could you also submit the un-minified code as well. I need to be able to see what is being changed when updates are made, and it allows for other users to be able to use the code and possibly modify it for their own custom games. Definitely keep the minified code as the code the sheet.json points to, but add pre-compiled/unminified the other code in a subfolder with the rest of your code. 

Thanks!

timdenee:
Hi Phil, thanks for this feedback - it's my first time submitting my sheet, I'm still getting my head around things.

As regards the other Sprawl sheet, there is nothing it does that this one does not. In fact I used Chris Wolf's sheet as a starting point, so all the attribute names are the same. 

On the minified html/css, I have added a subfolder for the readable code now, thanks for the tip. 

Let me know if there's anything else! 

Tim

pjbehr87:
The stuff you added looks good, but you are deleting the old sheet. I want to leave the old sheet there, but I will depreciate it and add your new sheet in its place. If you can revert the delete I'll add this new sheet. Also, thank you for using the same attr names, that way anyone using the old sheet can switch to the new sheet with no missing data.

timdenee:
Ah OK, I see -- that delete has been reverted and the new sheet is in a different folder ("The Sprawl" rather than "The_Sprawl"). Cheers!

pjbehr87:
Just wanted to say: This is one sexy looking sheet, you did an amazing job!

timdenee:
Thanks Phil - much appreciated!

PR #1775, opened 2016-07-18
First upload of Coriolis char-sheet!

devSithun:


pjbehr87:
The sheet looks really nice, just a few questions:
What language is this sheet in?
Does Coriolis use a specific game-system, or is it custom for this game?

devSithun:
Coriolis is a Swedish rpg (And thus the sheet is written in Swedish) that uses a rulesystem based on another popular Swedish rpg called Mutant: Year Zero, the rulesystem that was awarded a Silver ENnie for Best Rules at Gencon 2015. The rules are similar, but the sheets cannot be used interchangeably between the two rpgs, if that's what you're asking.

Although Coriolis currently is only available in Swedish at the moment, [a kickstarter for an English version has been funded](https://www.kickstarter.com/projects/1192053011/coriolis-a-sci-fi-rpg-from-the-makers-of-mutant-ye), and i will update with an English translation of the sheet when i get to see the English sheet itself. :)

pjbehr87:
That's everything I needed, thanks! When you do go about making the english version, could you use the new [Translation Tools](https://wiki.roll20.net/index.php?title=Character_Sheet_i18n) we released a month ago. We are trying to move away from separate sheets for different languages. These tools will let you have both the Swedish version and English, as well as any other language someone wants to translate, all on the same sheet. If you have any questions, let me know.

devSithun:
Yea just one question, as a GM, there will be many instances of when you want one campaign to be in english, and another, specifically for swedish rpgs, to be in swedish. But that all follows my account language settings! Will i be able to select language of a campaign on a case to case basis?

pjbehr87:
As the GM you can definitely say "we will be playing in Swedish" but that shouldn't stop someone from using the sheet in English or whatever other language if that's what they are more comfortable with. If you're talking about the LFP listing in particular, you can set the language for your game. Swedish isn't in the list because it's not as popular on Roll20, yet, but you can select Other and list it in the post title/description.

PR #1776, opened 2016-07-18
Pathfinder .56 tab update

plutosdad:
attempt to fix tabs
added more tab logging


plutosdad:
1776! the update that will bring freedom to tabs and stop them from getting locked up!

pjbehr87:
Yay and good luck!!

PR #1787, opened 2016-07-28
Adding Ars Magica 5e sheet

Rats808:
Adding a sheet for Ars Magica 5e, based on the one in the back of the core book, with some minor adjustments.

I'm aware another was merged a couple of days ago, but it has some inconsistencies with the rules(mainly spell Range and Target being numbers instead of text/select fields), and I started this one the same day without realizing that one existed.

Edit: Also some small changes to the Deadlands Classic sheet I submitted a while ago, I guess, since I forgot to make a PR when I made those commits originally.


pjbehr87:
Hey, sorry you did all this work only to have another sheet pop up first. We like to keep it to 1 sheet per game system, except in a few small cases. If there are system inconsistencies with the other sheet, you are more than welcome to fix them there, nobody "owns" their sheet here.

As for the Deadlands changes you made, I'm not able to grab those since you have merge conflicts. If you can add those Deadlands changes by themselves, without the conflicts, I'll be happy to merge them.

PR #1802, opened 2016-08-01
Add World of Dungeons character sheet

jangler:
Should go under Apocalypse World Engine, since it's a minimal hack of Dungeon World that was a stretch goal for the Dungeon World Kickstarter.


pjbehr87:
Thanks for the game system info, it's not always easy to find that info and it's impossible for me to keep up with the thousands of RPGs that are out there. :P

PR #1805, opened 2016-08-02
Auto-calculated attribute modifiers

kevinsearle:
Roll template changes and auto-calculated attribute modifiers.


pjbehr87:
New sheet looks good. I've updated the preview image to be a little bit cleaner, so make sure to update before doing more work so you don't get any merge conflicts.

PR #1810, opened 2016-08-03
fate accelerated edition fallout fan conversion

laresgod:
*edited pictures
*edited approaches
*edited rolls


pjbehr87:
How different from the original sheet is this? Would it be possible to combine it with the original sheet and just allow the user to select a "Fallout Skin"?

laresgod:
It is not only a reskin. Rolls have been changed to appear and approaches
have been renamed/adapted.
It is true that I could use another sheet and modify it + use a button
check to choose the skin,  if I knew how.  I do not have that skill yet.

pjbehr87:
If you look at the official 5e sheet and see how the sheet is made to switch between PC and NPC sheets with a checkbox.

The general form is:
HTML:
```
<input type="checkbox" class="fallout-theme-switch" name="fallout-theme" value="1">
<div class="original">...</div>
<div class="fallout-theme-sheet">...</div>
```

CSS:
```
.sheet-fallout-theme-sheet,
.sheet-fallout-theme-switch[value=1] ~ .sheet-original
{
    display: none;
}
.sheet-fallout-theme-switch[value=1] ~ .sheet-fallout-theme-sheet {
    display: block;
}
```

The way this works is using CSS's "~" operator, which means "sibling of".
So:
1. we hide the fallout theme by default.
2. If .sheet-fallout-theme-switch[value=1] exists, which means the checkbox has been checked, select the fallout theme, which is a sibling, and un-hide it while at the same time hiding the original sheet


If this doesn't make sense, or you need more help, let me know. I really would rather combine this with the original sheet, since it's mostly a theme and not an "official" system.

laresgod:
Understood.  I will try to change the original and test.  Then I will
commit again.

laresgod:
as requested,
 i modified the original fate accelerated sheet to add a fallout sheet at
the bottom
and 1 checkbox to hide/unhide original/fallout version
with a simpler trick in css.

I have just uploaded and asked for a pull request.

I hope it fits your recommendations and rules

2016-08-11 5:03 GMT+01:00 G. Ti. <laresgod@gmail.com>:

> Understood.  I will try to change the original and test.  Then I will
> commit again.
>

pjbehr87:
Alright, since you've modified the other sheet, I'm going to go ahead and close this PR.

PR #1812, opened 2016-08-04
fix strength roll macro for D&D 3.5 french

Polaris-Dust:
fixing issue with strength check macro

### DIFF
--- a/D&D_3-5-fr/D&D3.5-Fr.html
+++ b/D&D_3-5-fr/D&D3.5-Fr.html
@@ -29,7 +29,7 @@
 								</tr>
 							</table>
 						</td>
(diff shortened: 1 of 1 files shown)


pjbehr87:
You're changing the name of the attribute from str-mod to for-mod, this probably shouldn't happen. The different translations of the sheet should continue to use english attribute names, this way if a user switches between the different translations of the sheet it will transfer all of the data like they are expecting. I think the fix you are looking for is "[[@{for-mod}]]" at the end of that same line should be "[[@{str-mod}]]".

Polaris-Dust:
I understood that, for internationalization purpose, an english name has to be kept. The changes (using the french name) have been done this way to remain consistent with the whole sheet that was never intended to support any translation ("-fr" directory).
Nevertheless, I will submit another pr with the "str-mod" as you suggested (thanks!), but I will not refactor the sheet to use english attributes. Maybe the author is planning to do it ;)

PR #1816, opened 2016-08-06
Fixed CMB2 macro

jspjut:
Fixes [this issue](https://github.com/Roll20/roll20-character-sheets/issues/1815).

PR #1822, opened 2016-08-08
Pathfinder v.62

plutosdad:
1. Vertical scroll bars disappeared from textareas
2. 3rd iterative attack critical confirmation roll is labelled incorrectly as "4th attack" in the normal template, it is correct in the simple template. The number is a good critical confirmation roll, however, only the label is wrong. 
3. Skill totals not updating for new skills
4. Clickable area does not extend to the right to cover entire word.
5. Add extra macro to SKILL roll does not work if "add query to roll" is "No"
6. Damage in main attack macro per row does not add in damage calculations IF you had a custom macro that the sheet updated. If you had a custom macro,you must put a + between the dice and @{damage_macro}.
7. Craft section disappears when minimized.
8. Spell normal template: does not wrap to new row after Duration before Rolls.
9. Spell defensive casting bonus does not appear in template until after hitting refresh.
10. CMB2 roll incorrectly references CMB(1)


plutosdad:
Merged in update from https://github.com/Roll20/roll20-character-sheets/pull/1816

PR #1841, opened 2016-08-17
Create Traditional Chinese translation of Dungeon World character sheet.

shadowevor:
Full translation to Traditional Chinese from original English version.


pjbehr87:
Just FYI, new alternate translation sheets will not be accepted starting September 1st. The current existing sheets, including this one, will still exist, we just won't accept new ones. If you want to add a translation to an existing sheet after that date you can use the new translation tools we released several months ago, documented here: https://wiki.roll20.net/Character_Sheet_i18n

PR #1842, opened 2016-08-17
AD&D 2E Revised

Project462:
HTML, CSS and .json for a new AD&D 2E sheet. It was originally based off of Dylan G.'s sheet, but the changes implemented exceed 90% of the original material.


Project462:
_Does your new sheet do at least everything the old sheet does?_
The short answer is yes :)
_Is your sheet more complicated than the old sheet?_
Yes, in a few ways, but I've tried to keep as simple as possible, while maximizing its versatility.
_In your personal opinion, is the old sheet worth keeping around?_
Most assuredly. I had to change a lot of attributes for the new sheet, so taking away the old one will result in a lot of players and GM's losing information. I don't want that, I would just like this sheet to be a viable alternative.

pjbehr87:
I had to make some changes to your files, to please make sure to update/create a new fork if you plan on doing any more work on the new sheet. Please understand, _I'm not being critical_, I know you're new to the whole character sheet thing and to GitHub, so please use this as a learning experience if you plan on making more sheets. 👍 
Your sheet wasn't in a folder, so I put it in AD&D 2E Revised
The file names were all wrong, they didn't have css/html/json extensions and they didn't match the names of the files in the sheet.json file and I had to rename the json to sheet.json. (every sheet needs the file to be named sheet.json, because that's what our system looks for for the metadata for your sheet)
The json was not was not formatted properly, it had ";" line endings instead of ",".

You can look at the commit I made to see all of the changes.

Project462:
Ahh, tsall good. I actually have an update ready, (spelling errors,
formatting, etc) but trying to update my fork, the files are not showing
correctly. How do I go about submitting my update?

pjbehr87:
Save your changes somewhere else. Delete your branch, you can do this in the options, there is a big red button. Create a new fork, find your files in the new folder, make the changes, then create the PR and you should just see your changes.

Project462:
So, I've got eveything together for an update, except the image. I can't
seem to find a way to get a scrolling screenshot of the whole sheet. How do
I get one?

PR #1846, opened 2016-08-21
Update

sepianra:
Just minor attribute name changes for consistency

### DIFF
--- a/PokemonTabletopAdventures/PTA.html
+++ b/PokemonTabletopAdventures/PTA.html
@@ -21,7 +21,7 @@
 
     <div class="sheet-col">
         <p style="text-align: center;">
-			<img src="http://pokemontabletop.com/images/logo.png" style="max-height: 100px;" />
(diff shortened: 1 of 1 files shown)


pjbehr87:
Unless there is a bug caused by the naming convention of the attrs, I would rather not make this change. If we make this change every character using this sheet sheet will lose its max HP values, because the sheet is now pointing at a blank attribute.

sepianra:
Oh yeah totally understandable. I just made the change so that tokens could be linked to the HP values properly. But I understand 

Sent from Yahoo Mail on Android 
 
  On Mon, Aug 22, 2016 at 7:09 PM, Phil<notifications@github.com> wrote:   
Unless there is a bug caused by the naming convention of the attrs, I would rather not make this change. If we make this change every character using this sheet sheet will lose its max HP values, because the sheet is now pointing at a blank attribute.

PR #1839, opened 2016-08-16
Pathfinder-Neceros v63 patch

vince-roll20:
- fix for a display issue of Background Skill Totals

PR #1847, opened 2016-08-22
Pathfinder v0.64

plutosdad:
iterative macro fixes
html css fixes


plutosdad:
THIS INCLUDES THE .63 pull request so you can ignore .63 and go right to .64
https://github.com/Roll20/roll20-character-sheets/pull/1839

PR #1848, opened 2016-08-22
D&D 3.5 (fr), fix str check

Polaris-Dust:
better fix thanks to comments by @pjbehr87 in the https://github.com/Roll20/roll20-character-sheets/pull/1812


pjbehr87:
Thank you for making these changes!

PR #1850, opened 2016-08-23
Unknown Armies Third Edition sheet

Rats808:
Adding a sheet for Unknown Armies 3e.


pjbehr87:
There is already a sheet for Unknown Armies, but I do not know the system enough to tell which edition it is for. Do you know which edition the current sheet is for, so that I can label it properly when I add this new sheet?

Rats808:
The current sheet is for 2nd Edition.
(Though I think it'd also work for 1e as well, I'm not entirely sure.)

pjbehr87:
If it at least works for 2e, then I'll label it as such. Thank you for the information!

pjbehr87:
The new sheet looks really good, I like the text styling. You have a few CSS issues, though. The sheet will probably go live tomorrow, and you can see what I'm talking about then. (since it's a new sheet, I'm not worried if there are some issues. You'll be able to fix it before most people know it's there)

screen shot 2016-08-29 at 7 07 08 pm

PR #1857, opened 2016-08-26
New Twilight 2000 V2.2 character sheet, Image replacements for Flashing Blades

Lockbox313:


pjbehr87:
Please change the options in the text editor you are using to use the line ending LF instead of CRLF. The system our servers are running on are unix and expect the standard unix line endings (LF).

Lockbox313:
Will do.

PR #1858, opened 2016-08-28
Remove ABS calls in 13th Age Official

Wolvereness:
I don't have the ability to directly test these changes and how they interact with other operations, but the calculations themselves work and have a rather dramatic performance gain. I encountered issues as a GM using the AC/MD/PD and serious delays in UI responsiveness after attempting to execute macros using those values.

Using the history, first use of ABS in 13th age sheets comes from ea32475ee90052d49bc375a8f7bef643675cfcff.

Commit message explanation copied here:

> This commit removes calls to ABS in internal attributes in favor
of other built-in functionality: dice calculations.

> Calls to ABS suffer from performance issues, with AC/MD/PD calls
taking upwards of 10 seconds. When changing these to use the
built-in dice calculations, there are no noticable performance
issues.

> The built-in dice functions are treated as clamp operations,
providing concise min and max operators, and tests, providing
level-range operations.

> Min and max operators are used rather directly for calculating
the middle-mod for stats. They are also used to clamp the
escalation die to [0, 6] range.

> The max operator is used to form baselines for health
multipliers, with higher health "overflowing" from 0 when the
level matches the next change in step-size.

> The level multiplier for stats in damage rolls is calculated
by performing 3 tests, adding together the successes to form
the [1, 3] range of multiplier. A level of at least 1 is assumed
during the calculation, and that test is kept for reading
clarity.

### DIFF
--- a/13th Age Official/13th_Age_Official.html
+++ b/13th Age Official/13th_Age_Official.html
@@ -705,23 +705,15 @@
     </div>
     
 <!-- HIDDEN FIELDS USED FOR CALCULATIONS -->
-<input type="hidden" disabled="true" name="attr_AC-max-con-dex" value="((((@{CON-mod}) + (@{DEX-mod})) + abs((@{CON-mod}) - (@{DEX-mod}))) / 2)" />
-<input type="hidden" disabled="true" name="attr_AC-max-con-dex-wis" value="(((@{AC-max-con-dex}) + (@{WIS-mod}) + abs((@{AC-max-con-dex}) - (@{WIS-mod}))) / 2)" />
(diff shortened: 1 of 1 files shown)


feildmaster:
I'd like to stress that the new calculations is computed in less than a second, where as the current calculation takes over 15 seconds to compute.

Testing using:
#AC-fast [[@{selected|WIS-mod}+@{selected|CON-mod}+@{selected|DEX-mod}-{@{selected|WIS-mod},@{selected|CON-mod},@{selected|DEX-mod}}dl2-{@{selected|WIS-mod},@{selected|CON-mod},@{selected|DEX-mod}}dh2+@{selected|AC-base}+@{selected|level}]]

As opposed to:
#AC [[@{selected|AC}]]

PR #1863, opened 2016-08-29
Adding character sheet for FASATrek

SETRusty:
Star Trek RPG by FASA from the 1980's - This character sheet is not 100% complete but I am looking for the ability to preview some of the formatting and calculations in Roll20.


pjbehr87:
You don't need to include the header, it will actually get stripped out. The css file that you link in the sheet.json is the one that will be included with the sheet.

SETRusty:
Ahhh ... Gotcha! I'll note that for the next rev of the sheet.

Thanks kindly!

SETRusty

pjbehr87:
Also, please change the options in the text editor you are using to use the line ending LF instead of CRLF. The system our servers are running on are unix and expect the standard unix line endings (LF).

SETRusty:
Thanks for the fix. I'll pull down once my GM and I have a look at what we've got. Thanks a bunch for the help here!

SETRusty

PR #1870, opened 2016-08-31
Revert "Remove ABS calls in 13th Age Official"

Wolvereness:
https://github.com/Roll20/roll20-character-sheets/pull/1858 actually ends up breaking the value display. The system doesn't seem to support those types of calculations except for final roll calculations.

> This reverts commit 8453945106fc396fde964763d10799e16cc478ba.

> The character sheet display system doesn't support "drop" logic or
"success" logic.


Wolvereness:
Closed by 0085c4b159dabbc2493a0bc5554c003834306dc9

PR #1872, opened 2016-08-31
Translation of strings to Portuguese

elvis-pereira:


pjbehr87:
Thank you for your translation work!

PR #1880, opened 2016-09-07
Master

darkxyro:
New Shadowrun 5th Edition Character Sheet


pjbehr87:
Hey, we already have a sheet for Shadowrun: 5e and we try to keep it to a single sheet per system except when there is a reason to have more than one sheet.
Are there reasons for a second sheet?
Does your sheet do things that the current sheet does not?
The old sheet has not been touched in 2 years, are there things that are wrong that couldn't be fixed by just modifying the original sheet?

darkxyro:
Hi,
Yes the new sheet is a full reboot with all the functionality of the old one and a large number of new features (been working on it for a few hunderd hours).
Improvements over the old one:

-organization in tabs with relevant information for differnet situations being found in one tab.
-more options to have stats organized and distinct.
-additional input possibilities for rolls with bonuses especially from new supplements
-more html/css wizzardry to achieve higher levels of customizability
-added missing sections which the old one was plainly lacking
-especially the vehicle section was basicly unuseable on the old sheet
-tooltips for every single input/button/etc

It was too much of an overhaul to try to cram into the old sheet.
Once all the bugs are out the old sheet should be redundant.
So far the sheet is being used in 3 campains i am part of or running and we have not found anything worse than a typo.
Sorry if this is not standard procedure, been looking through the forums and all i came up with is submiting it here on github

pjbehr87:
The new sheet looks really nice, you did a good job. I've added it to the list as Shadowrun: 5th Edition (Advanced), and am going to keep the old one as well since it has received updates recently and I don't want to remove other people's work. I had to add the "preview" line to your sheet.json so be sure to pull my updates before you commit any new work or you'll have merge conflicts.

PR #1881, opened 2016-09-08
Spelling corrections and better abbreviations

elvis-pereira:
Spelling correction on line 201. "ARMOR CLASS", "Hit Point Maximum" and "EXPERIENCE POINTS" shortened to better fit the layout.


pjbehr87:
Thank you for making these changes. If there are other changes or if these changes aren't able to properly convey what they need to, please let me know. I am able to modify the sheet design specifically for each language, so I can fix it for PT. I hadn't done it yet, because I've been getting ready for a vacation and am out all this week. I can make style changes for the sheet next week.

elvis-pereira:
The phrases I abbreviated were the only causing problems with the layout, I believe that there will be no more such problems, but it is good to know that you can change the layout to suit the language :)

PR #1895, opened 2016-09-15
New sheet for Hawkmoon

MrPadRpg:
New Character Sheet for the rpg Hawkmoon.
The default version is in french but the translation is available.


pjbehr87:
I had to rename the translations file to translations.json so it would work with the system. So, you'll want to make sure to update your branch before doing any work on it.

PR #1904, opened 2016-09-20
All Flesh Must Be Eaten- Damage Fix

CoalPoweredPuppet:
Damage is now number rolled times multiplier (and strength, if checked), instead multiplier worth of dice rolled


CoalPoweredPuppet:
Thank you!

PR #1908, opened 2016-09-24
Update Shadowrun (Advanced)

darkxyro:
big improvements to the advanced shadowrun sheet


pjbehr87:
This PR has a bunch of merge conflicts from other sheets, which means you need to update your branch and make your changes again. The best way to do this is to make a local copy of your SR5th files then completely delete your branch. After you've deleted it make a new fork and re-add your changes back to the new updated branch.

PR #1911, opened 2016-09-24
PF - Neceros updated sheet preview

vince-roll20:
- just an updated sheet preview image
- removed some text from description


vince-roll20:
Thanks

PR #1912, opened 2016-09-25
Pathfinder (Cazra)

Cazra:
A simple, but flexible alternative to the official Pathfinder sheet. My players were unanimously dissatisfied with the UI and UX design of the official sheet, so with this sheet I made it my goal to provide an easy-to-use but still powerful alternative. Features tabs, auto-calculated fields, rollable stats with notes, and more while providing an easy-to-use and versatile interface. So far, my players have been very happy with this sheet, and I hope that other Roll20 Pathfinder users enjoy a similar experience from this contribution.


pjbehr87:
There is already a Pathfinder (Simple) on our list, but it hasn't really been touched, other than a quick mispellings fix, in about 2 years. So, I'm going to deprecate that sheet and put this one in its place. Games already using the old simple sheet will not be affected, and will continue to use that old sheet, but new games will see this sheet instead of the old one.

PR #1916, opened 2016-09-26
New Sheet for polaris RPG

MrPadRpg:
I added a new character sheet for the RPG Polaris


pjbehr87:
Thank you for adding both the English and French translations for this sheet, that's super helpful.

pjbehr87:
I just wanted to let you know that I am going to change the base translation file to be in English, rather than French. I am doing this because all of our translators know at least English + their language (because the website is originally in English, and they volunteered to translate from English to their language), rather than French + their language. You will have a better chance of one of them translating the sheet into another language is the base file is in English.

If you would rather the base be kept in French, please feel free to change it back and I will respect that decision since you wrote this sheet.

MrPadRpg:
No problem to put the English as base file.

PR #1932, opened 2016-09-29
DFRPG Spanish translation

Zexar:


pjbehr87:
Thank you for the translation!

PR #1941, opened 2016-10-06
Fix for dice sometimes not showing up if the result is 0 or less.

LewisWakeford:
Will make old version of this sheet's roll templates unviewable. I don't think the sheet is widely used at the moment though.

If that is an issue I can go back and add some backwards compatability.


pjbehr87:
This sheet is still pretty new, and for a small game system, so I think it's safe to let it slide this time.

PR #1948, opened 2016-10-10
The Sprawl character sheet - Removed IDs from HTML so that multiple sheets can be open at the same time

timdenee:


pjbehr87:
You've got some merge conflicts with files from other character sheets, which means your branch is out of sync. The easiest way to fix this is to make a local copy of just your sheet changes, delete your personal branch, fork the master repo from scratch and re-add your changes back to your new branch.

PR #1958, opened 2016-10-14
Improves attribute tests for 3D&T ALPHA

jcemelanda:


jcemelanda:
I separated this pull request into 2 others becaus they were related to differente character sheets

PR #1959, opened 2016-10-14
Vampire the Requiem: health boxes

CoalPoweredPuppet:
replaced images used in open and lethal damage areas.


pjbehr87:
The best way to handle images is to add them to GitHub and link to the "raw" version. That way, we don't have to worry about a 3rd party randomly removing the image.

CoalPoweredPuppet:
Thanks for the tip, and the merge.

PR #1961, opened 2016-10-14
Implements Tormenta ALPHA 3d&t scenario cheat sheet

jcemelanda:


pjbehr87:
There is already a Tormenta sheet. Is Tormenta ALPHA different than Tormenta? If they are the same, how is this sheet different than original?

PR #1967, opened 2016-10-16
Fixed translations and added new translated sheets.

shadowevor:
Traditional Chinese is currently not supported by the new i18n sheets due to the lack of Traditional Chinese option in account settings. Therefore I added new sheets for Chinese translated version.

D&D5e:Fixed incorrect translations.
Stars Without Number:Add new translated sheet.
Lotfp:Add new translated sheet.


pjbehr87:
Rather than add a new translated sheet, I'm going to add Chinese Traditional to our list of supported languages. We have several sheets that already use it, so it's obvious there is a need for it. I'm sorry you did the work to translate the sheets already, but would you mind using the new system? By the end of the week Chinese Traditional should be a language option from the account menu.

shadowevor:
Sure, that would be nice. Just to be sure, how do I volunteer to translate with the new system? By simply edit the corresponding language json file manually?

pjbehr87:
You can edit the zh.json (the 2 letter symbol for Chinese Traditional) file in the translations folder. If you would like to use the official translation system, Crowdin, which would give you more tools, a better interface, and the ability to also help translate the rest of Roll20 you can email team@roll20.net and ask to be added to the translation team. If you are adding a translation for a sheet that has not been updated to the new system, you will have to modify the sheet to accept alternate translations. You can find out how to use the tools [here](https://wiki.roll20.net/Character_Sheet_i18n).

If you are going to be modifying a sheet and have additional questions, you can ask me here and I'll help in any way I can.

PR #1973, opened 2016-10-23
IK RPG Massive Update

CoalPoweredPuppet:
Update to IKRPG
Condensed Character, Warbeast, and 'jack to a single page
Added transformation area for tharn and warp born, etc.
Added boosted rolls
Upgraded roll templates 
Other minor changes

### DIFF
--- a/Iron Kingdoms RPG/IKPRG.html
+++ b/Iron Kingdoms RPG/IKPRG.html
@@ -1,3 +1,5 @@
+<div>
+
 <input type="radio" name="attr_tab" class="sheet-tab sheet-tab1" value="1" style="width:120px;" title="Basic" checked="checked" id="radio1" /><label class="sheet-bob"  for="radio1">Basic</label>
 <input type="radio" name="attr_tab" class="sheet-tab sheet-tab2" value="2" title="Skills" id="radio2" /><label for="radio2">Skills</label>
 <input type="radio" name="attr_tab" class="sheet-tab sheet-tab3" value="3" title="Abilities" id="radio3" /><label for="radio3">Abilities</label>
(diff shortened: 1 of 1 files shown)


CoalPoweredPuppet:
Thank you!

wizard04wsu:
Not quite backwards compatible. I had to fix some things from a hard copy of mine after this update:
- character stats were back to base values
- race was set to Bogrin
- weapon skills were all set to Great Weapon
- at first I thought all my skills were wiped out; maybe make it more obvious that they're just hidden?

Other issues:
- the armor modifier of the shield is included in ARM even if set to "No Shield", and another +1 is added when set to "Shield"
- "Character Name" and "Warbeast Name" use the same value (should be `attr_character_name` and I'm guessing `attr_Warbeast_name`)
- **Most of the fixes I made in pull requests #1782 and #1799 were not included**

CoalPoweredPuppet:
- Race is a drop down menu.
- Weapon skill is a drop down menu
- Click the hide/show button (looks like a gear) in skill tab, pick which skills you want shown.
- Will Look into the shield/no shield thing
- They both use character_name.  
- Actually did not know about those changes.  

Other than what was listed above, what did you fix?

wizard04wsu:
I really like your changes, by the way. :-) Sorry I didn't say that up front.

I see where the race/weapon skills are, but perhaps there's a way to script it so when the character sheet loads it will use the old values to update the drop-downs (and character stats?), then remove the old values. I haven't figured out how the value-updating works yet; you'd obviously know better than me if it's possible.

For the skills, how about splitting it into two gears, one right after the Military Skills heading and one after the Occupational Skills heading? I think it would make it more obvious that there's more to see in those sections.

My changes in IKPRG.html:
- changed lone ampersands to `&amp;`
- removed the `/` from the opening tags of textareas
- for Cygnaran, changed the input name from `attr_Caspian` to `attr_Cygnaran`
- for "3 advancement points", changed the name of the third input from `attr_ap_2` to `attr_ap_3`
- changed "3 advancement points" to "3 advancement points (+3 stats)"
- under "Other Defenses", changed the note box name from `attr_shield_note` to `attr_otherarmor_note`
- several other small changes in #1799 

My changes in IKRPG.css:
- indent the archetype benefits sub-lists
- make roll buttons adjust their heights automatically by removing `height:24px` (though I see you doubled their heights in this update)

CoalPoweredPuppet:
I have started testing improvements and your upgrades in this campaign: https://app.roll20dev.net/join/38791/-5HkAA

PR #1977, opened 2016-10-24
Fixing formula typos

KingMarth:
Not sure of etiquette here, but I hit this typo on the special-damage macro of Move 3 in my game recently; the attribute "@{snipe]]}" isn't closed properly and should be "@{snipern}]]}". @cwstra is the proper sheet maintainer and I'm not a Pro subscriber so I can't test this in-game, but this should work.

I'm also reasonably certain that the Move 6 macro shouldn't be referencing Move 3 attributes.


cwstra:
Thanks Marth! Nice to know the code is legible enough for someone else to find and fix bugs. :P

PR #1980, opened 2016-10-25
Pathfinder v.68 harder better faster stronger part 1

plutosdad:
1. added more collapsing sections, so you don't have to see all 6 rows of the class grid that you're not using, shrunk down spell domain/school area and made collapsible sections there. will do same for armor and hopefully skills. 
2. reorganized code to be modular for easier support. Reworked refresh to be faster
3. fixed: attack or damage sometimes does not appear when rolling attack (fixed in .68 beta)
4. fixed: cmb rolls with no damage still showing damage:0 or damage:(buff value) when rolling (Fixed in .68 beta)
5. fixed: negative ability buffs act as "drain" instead of as "penalty" (fixed in .68 beta)
6. fixed: ability penalties, conditions, and damage not adding correctly (fixed in .68 beta)
7. Added command buttons above top menu so players see them (many don't even know they are there)
8. Added abilty checks, and concentration/caster level to top of spellbook, base attack types to top of attacks


plutosdad:
I merged from the source over and over, but in the github windows tool it is still telling me it won't do it and there is a commit to merge back. Not sure why.

PR #1984, opened 2016-10-28
Master

SolCheetah:


pjbehr87:
Your sheet.json is missing the commas to make it a valid JSON. This is an easy fix, so I will do it myself, but you will have to be sure to pull down my changes or you'll run into merge conflicts if you try to make more changes.

PR #1990, opened 2016-10-30
Added English version character sheet for Coriolis RPG in anticipatio…

mguilli:
Added English version character sheet for Coriolis RPG in anticipation of full game release at the end of 2016.


pjbehr87:
We are trying to move away from having separate sheets for different languages. Instead of adding this new sheet, could you please modify the original sheet to use [our new translation tools](https://wiki.roll20.net/Character_Sheet_i18n).

PR #1991, opened 2016-10-30
White Lies : new sheet (english and french)

NathaTerrien:
New sheet (english and french) : White Lies.
Based on Swords & Wizardry WhiteBox, White Lies brings all the simple
elegance of WhiteBox out of the dungeon and into the realm of espionage
and paramilitary action and adventure!


pjbehr87:
I think you forgot to add the translation.json file. You said English and French, but only included the fr.json file.

NathaTerrien:
@pjbehr87 
Sorry, since the base html is in english, I thought I wouldn't have to provide it.
I've added the en.json file, if it's not too late.

pjbehr87:
It's no problem, and it's not too late. The way the translation system works is by overwriting all of the text with the language file tied to the user's chosen language. It does need another small change, but I will do it because it's easy. The base language should be in the root of the sheet folder as "translation.json". Any _additional_ translations go in the translations folder like you have for French. In Crowdin, all of the translations are saved as a "translation of" the source file, the translation.json file. It's all a little confusing, but I had to do it this way in order to be able to automate the process. (which is definitely needed as we continue to add more and more translated sheets)

NathaTerrien:
Ok, thanks.
Duly noted for my next multilingual sheet.

PR #1992, opened 2016-10-31
Add sheet of Knight

Zakarik:
This is my sheet of Knight.

I hope it's ok.


pjbehr87:
This sheet looks really nice! Great job!

PR #2000, opened 2016-11-02
1st commit of Naruto d20 Modern

RinVindor:
Hope I followed all the instructions properly, I've created a d20 Modern sheet from the original one on the repository that is designed for [Naruto d20](http://narutod20.com/). The reason being is that the system changes so much to d20 Modern that it requires it own sheet. I've had numerous requests from forum goers on the site and so this was my crack at getting something functional.


pjbehr87:
You did everything correct, the sheet looks great. Congrats on your first sheet!

RinVindor:
Huzzah! :D I appreciate that!

PR #2008, opened 2016-11-04
Added Files for future Enhancements

finderski:
Planning to try using Git Hub as a repository for character sheet
required files so I don’t need to rely upon external web servers to
contain the necessary files. Also wanting to experiment with getting
fonts to work on my character sheet without having to “Load Unsafe
Scripts”


pjbehr87:
I don't think this will actually help anything, but I need to talk to the other devs who know more specifics about how the fonts work with the character sheets.

finderski:
Ok. Just let me know. I know images can be accessed via github, so I figured I'd try this approach for fonts to see if I could get them to work.

Rich Finder

"Nice tuxedo! Nice tuxedo to die in!"

finderski:
Just wondering if there’s been an update to this? Because I noticed that my Pull Request is still hanging out there.

Thanks,

Rich

pjbehr87:
Sorry it's taking so long, the Dev that knows the answers to my questions has been out for a bit.

finderski:
No worries.

Just didn’t want this to get forgotten. 😀

Rich

pjbehr87:
Finally was able to talk to Riley, sorry for the wait. It looks like the way we currently handle things this won't work with our system. We don't currently support the ability to import fonts, but this _might_ be something we add in the future. (it is definitely not a priority, so don't hold your breath) If you still want to add these to the git, for whatever reason, that's fine. I just need to know that these fonts are under an open license, otherwise we are not allowed to add it.

finderski:
They should be, they are all google fonts…but I’ll go ahead and delete them. Too bad, because fonts would really make character sheets pop, and I hate that people have to “load unsafe scripts” in order to get fonts to display. I was hoping that by having them on git, we could get around that issue, since hosting images on git seems to work. Oh well. Thanks for checking.

Rich

finderski:
Ok, I finally got around to deleting the fonts and committed and sync’d. The pull request should now be updated, ending in an empty Pull Request. I have no idea if/how to just cancel a Pull Request, so if you could now approve/accept it, that would be awesome.

Thanks for your help,

Rich

pjbehr87:
The whole fonts thing comes down to how browsers force fonts to be loaded and the current proxy we use to load external items. We may be able to fix this in the future, but it would take a considerable amount of development work, so it probably it won't happen any time soon. Sorry for the inconvenience.

I went ahead and closed the PR since it was empty.

PR #2011, opened 2016-11-06
Sheet for Masks: a new generation in French

FkinGuy:
This is a straight french translation of the sheet made by Matt Morton.
Pictures containing text aren't translated yet, as Magpie Games has not given the pictures in french yet (should happen in a few months with the official french translation of the game).


pjbehr87:
We have stopped accepting "alternate translation" sheets to try and slim down the total number of sheets we have, see [here](https://app.roll20.net/forum/post/3762625/upcoming-policy-change-for-alternate-translation-character-sheets) for more info. If you want to add a new translation for a sheet, we ask that you use the [character sheet translation tools](https://wiki.roll20.net/Character_Sheet_i18n). This will allow the original sheet to not only support French, but any other language one of our users might translate it to. I'm sorry for the inco

PR #2014, opened 2016-11-07
New character sheet for Call Of Cthulhu 6th Edition

MrPadRpg:
"Translatable" character sheet for Call Of Cthulhu 6th Edition


pjbehr87:
Just for future reference, you don't need to generate the "translations" folder and contents. (unless you are including an alternate translation to start with) Our translation system will automatically create the folder and all of the language files based off the "translation.json" file in the root of the sheet directory. If you are going to include an alternate translation, you would only need to include that one file in the translations folder. Thank you for using the translation tools!

PR #2015, opened 2016-11-07
Create Street Fighter O Jogo de RPG

trueshinken:


pjbehr87:
You're missing some files that are required for a new sheet. Every new sheet needs at least 4 different files:

- [x] The html, which should end in ".html"
- [ ] The css, which should end in ".css" (even if it's empty)
- [ ] The preview image, which can be either a ".jpg" or ".png".
- [ ] The "sheet.json" file that points to all of these files. (you can see an example [here](https://github.com/Roll20/roll20-character-sheets/blob/master/13th%20Age%20Official/sheet.json))

Then, all of these files must be in their own folder, just like every other character sheet in the list.

PR #2020, opened 2016-11-09
Fix issues in Shadows of Esteren character sheet

radoslawg:
Major: 'Easy' difficulty step was set to low!
Minor: Logo image was missing


pjbehr87:
Just FYI, the best way to handle image hosting is to upload the image to the GitHub in one commit, then link to the Raw version once we've accepted the PR. That way you don't have to worry about a 3rd party randomly removing the image.

PR #2023, opened 2016-11-10
Dungeon World French v3.1

NathaTerrien:
Dungeon World French v3.1 : optional interactive bonus to rolls, new
Notes tab and better tooltip in roll templates.


NathaTerrien:
@pjbehr87 : the merged update is still not available on the production server. Is there a problem?

PR #2064, opened 2016-12-02
Fixing listen, find secret door, and find trap chances for Moldvay D&D sheet

DBendit:
For whatever reason, the chances for listening at doors, finding secret doors, and finding traps were all too high by 1. This resulted in incorrect indications of success when using the in-sheet rolls. Now, the chances all match those given in the Moldvay Basic rules.


DBendit:
I'd also like to add that I have no idea how to test these changes, as I'm brand new to this. Any assistance there would be appreciated. Right now I'm cribbing off of other sheets and the wiki.

Also, I figured out where the incorrect values for the checks came from: the spec sheet for the contest that started this sheet out. See https://goo.gl/tYGcsG. No idea where they got their info from, but it's directly contradicted by pages B21 and B22 of the Basic book.

pjbehr87:
If you have a pro account, you can go to the game settings and choose "Custom" and copy and paste all of the HTML/CSS into the given tables. This will let you use custom code for a game.

PR #2068, opened 2016-12-04
New character sheet for Blue Planet V2

Tomaxx:
New character sheet for Blue Planet V2 published by Fantasy Flight
Games. Optimized for Chrome.


pjbehr87:
The sheet looks great. Just FYI, you don't need all of the preview images, only the one image of the front page. One quick question, does this sheet have a specific system? I've tried to look it up and can't find anywhere that says what system it uses. If it's a one-off game that doesn't share it's mechanics with another game that's fine, I just want to know if I should be grouping this with some other sheets.

Tomaxx:
Hi Phil,

Thanks for your interst in my sheet :)

I understand that there only needs to be one sheet preview, but the extra
images are for the viewing pleasure and information of people looking at
the sheet on github ;)

Btw, I saw that Blue Planet V2 was on the sheet request list, so that will
need to be updated as well.

To answer your question, Blue Planet V2 usues a system that I believe is
unique to this game.
Blue Planet (v1) was published by a different company, and used a
completely different mechanic than the one used in the Fantasy Flight Games
version.

Best,
Tomaxx

PR #2069, opened 2016-12-04
No support available so rolling back to before the tab added

Lockbox313:
Cannot get any support from Roll20 to find out why 2nd Tab is missing so rolling changes back


pjbehr87:
Found the issue, there was a bunch of unclosed div and paragraph tags in the first tab. The changes I made will look pretty extensive, but I just ran an auto-indent program to make everything conform so I could find the issues.

pjbehr87:
Also, if you want to have a background image on the sheet, the best way is to upload the image to Git, then after we've added it to the repo you can link to the "raw" image directly on GitHub. That way you don't have to worry about a 3rd party randomly removing the image.

Lockbox313:
Thanks for taking a look at it.  That will be great if that resolves the issue, but the unclosed tags have been there for some time (i.e. before any updates I made), did the extra tab finally push it over the edge?

pjbehr87:
I guess so, yeah. I closed a few and it didn't fix it, but then I closed a few more and somewhere in there it started working. So, I don't know what specifically was breaking it...but in the process of cleaning up the tabbing and closing tags it started working.

Lockbox313:
If you don't mind me asking where did you do the testing as I could not replicated it using the custom sheet.

pjbehr87:
Being a dev I can run all of the code locally on my machine and swap out the sheet code like it is on live. I'm not exactly sure why it worked through "Custom" and not the other way, I've never seen it happen like this before. So it must have been some very specific use-case that was catching one of the very slight differences between how the sheet code is rendered between the 2 options.

Lockbox313:
Ok thanks again for doing that and sorry about the emails but it is frustrating not to be able to fix something you may have broken and having no answer for the people who were asking.

pjbehr87:
Yup, I understand. In the future you don't need to message us all over the place. If it's a sheet issue you can leave a PR message like this or PM me on Roll20, and I'll see it quicker. Also, I don't work on the weekends, and am usually pretty crazy on Fridays trying to have everything in order before I log out for the weekend, so try to get me during the day the rest of the week and I can be quicker to respond.

Lockbox313:
I do not want to add to your burden but there is another issue with
differences in  behaviour between custom character sheet and drop down
https://app.roll20.net/forum/post/4328475/call-of-cthulhu-6e-the-english-character-sheet-does-not-work

PR #2077, opened 2016-12-07
Pathfinder .72

plutosdad:
fix to skill macro and refresh


plutosdad:
closing, users found more bugs may as well fix them

PR #2081, opened 2016-12-09
New sheet : The Strange from monte cook games with 1 sheet for all re…

angelodiego:
New sheet : The Strange from monte cook games with 1 sheet for all recursions with calculated fields (took Natha's cypher system sheet and made it specific to The Strange)


pjbehr87:
There is already a The Strange sheet in the list for Roll20. We try not to have more than one sheet for a system, unless they offer different options. Please tell me how this sheet is different and/or better than the sheet we already have for The Strange.

angelodiego:
Hi, the strange is a peculiar game where you have some base stats and some
modifiers for each of the word you "visit", the way is handled with the
existing sheet is by creating multiple sheets for each world and updating
them each time the base stats changes (it's a pain for the players). My
sheet has a base stats tab and a tab for each world (you visit worlds again
and again) you update the base tab and everything else is updated.
Regards
Angelo

pjbehr87:
Alright, that's good enough for me. I have to add a "tag" to differentiate the sheet, I'm thinking (Tabbed). If there is a different tag you would like me to use, let me know.

angelodiego:
If it has to be a generic tag, tabbed is fine, otherwise multi-recursion could be a good one

pjbehr87:
Alright, I went with multiple recursion. Thanks

PR #2093, opened 2016-12-18
2 New Character Sheets

finderski:
Initial release of Prowlers and Paragons Character sheet, and the
initial release for the Quilk System Character Sheet.


pjbehr87:
I can't find anything on the Quilk system except for a single thread on Roll20 that you commented in. It looks like this is a homebrew system that is only used by one, single, person. We try to only add games that have full public rules set. I can add it here to the repo, but I don't think I'm going to add it to the dropdown on Roll20. If they want to use this sheet they can copy the html from the repo and use the Custom game option.

PR #2099, opened 2016-12-21
DnD5E Character Scrolls - a homebrew-friendly DnD 5E character sheet

cyanobacterium:
Hello Roll20 Team!

I understand that you are reluctant to add additional character sheets for D&D, so I won't take it personally if you decline this one. However, I created this character sheet because all of the existing D&D 5E character sheets on offer suffer from two or more of the following issues:
- Very few visual mnemonics (this is a problem because one of my players is dyslexic and prefers to have visual aids to accompany text and labels)
- No support for multi-class characters
- Cannot enter custom classes (there's interest in play-testing the Mystic class from the Unearthed Arcana articles)
- Irrelevant (to particular characters) computation tables that confuse my newer players

The root of the problem is that the existing character sheets have too much automation. They have too many drop-down menus and not enough plain-text fields to be filled in with custom content. Class choice became the most obvious problem because it is one of the first choices a player makes in character creation, but there were other aspects (such as spell slots) that were also too mathematically rigid on some of the existing D&D sheets. To be clear, I do greatly admire the existing DnD5E character sheets and the huge amount of work that has gone into making them take full advantage of the digital medium, but it's just too much for some groups.

Thus I humbly submit my own creation, the D&D 5E Character Scrolls sheet, for your consideration. I wasn't originally planning to submit this to you, but my players have told me that they like it more than the other Roll20 D&D sheets they've used, and if other people would like it as well, I don't want to deprive them of the opportunity.

dnd5e-character-scrolls_preview

Yours truly,
-Chris Hall (aka Cyanobacterium)

P.S. You might want to adjust the URLs in the .css sheet. I use a sprite-sheet image and two background images to populate the graphics on the character sheet. Presently, the URL points to the image resource on one of my GitHub repositories (which is fine by me if you want to keep it that way), but I don't know whether GitHub imposes a bandwidth limit on image access or whether your security filter caches such image resources (I did what I could to minimize the image file sizes).
 
P.P.S. I was a little confused as to whether I am supposed to submit pull requests to the *master* branch or the *add-new-sheets-fork* branch


pjbehr87:
I want to start with saying this sheet looks really nice, and the quality of the code and images is fantastic. That said, I will want to talk to the other devs to get their thoughts on having another D&D 5e sheet. This would bring us to 4 separate sheets, not counting the Chinese translation. But, of all game systems this is by far the most popular, so...maybe. The other devs are out on vacation, so it'll be another week or so before I can give a solid answer. In the mean time, I'm going to at least add it to the repo. If you don't mind updating the image links to point to our master version for me, after I've added them, that would be awesome.

pjbehr87:
I talked to the other devs and we have decided that your sheet has different enough functionality from the other 5e sheets we have and that 5e is popular enough to have room for it. So, I'll be adding it to the list with the rest of the updates this week.

PR #2105, opened 2016-12-22
[Pendragon] Quality of life changes.

RhOmicron:
Changes to dice rolls with the skill surplus beyond 20 added to the die roll.
Lines: All roll_ / roll_test lines apart from attacks.

-Additional fixes

Typo fixed. Ennemy changed to Enemy
Line:182

TPio changed to TSpi to reflect Spiritual instead of old Pious.[REVERTED]
Line:326-335 and 392

Distaff being Courtesy in roll text.
Line:441

Fashion being Falconry in roll text.
Line:461

OCD fixed, Chirurgery changed to chirurgery in roll text.
Line:426

Skills changed to look somewhat the same[REVERTED], as well as some 'roll_TestFOR' fixed to be the skill its supposed to be.
Lines: 413-551

@Cskillvalue changed to '@Cskillname (skill)' in the roll text at the end.
Line:557

Squire skills are called their name now instead of the skill values being put in. i.e: @{squire_skill_value} is now 'squire (skill) roll'
Lines:566 572 578 584

Squire horsemanship fixed to being horsemanship and not battle.
Line:584

'squire @{squire_skill_name} skill' instead of 'squire skill' in roll text.
Line: 590

'@{CCname} skill' instead of just '@{CCname}'
Line: 635

--For people that play with personalitiy traits 20--

Tried to make personality traits go above 21 on both sides as the rule states it can happen. And as it happens the surplus beyond 20 gets added to the die roll.

I was not able to make this happen.

The positive side of the personalities somewhat works as the surplus gets added. But the negative side becomes -1 and decreases as the positive one increases while it should stay at 0. (Fix for this if you need to make a opposed roll is to decrease the positive side to 20>)

The negitive traits can't go higher than 20 as the traits would need to be editable (becomes 20--1 if you go below 0 on the positive trait which you shouldn't be able to do. And breaks the surplus roll).

What I figured was, the easiest fix would be to make people set and keep track of these themselves.
If it was something people wanted.


RhOmicron:
Working on personality traits being able to be over 20 on either side and the opposite staying at 0

RhOmicron:
Tried to make personality traits go above 21 on both sides as the rule states it can happen. And as it happens the surplus beyond 20 gets added to the die roll. 

I was not able to make this happen.

The positive side of the personalities somewhat works as the surplus gets added. But the negative side becomes -1 and decreases as the positive one increases while it should stay at 0.

The negitive traits can't go higher than 20 as the traits would need to be editable (becomes 20--1 if you go below 0 on the positive trait which you shouldn't be able to do. And breaks the surplus roll).

What I figured was, the easiest fix would be to make people set and keep track of these themselves.
If it was something people wanted.

pjbehr87:
You've got some merge conflicts. There was another change to Pendragon that was submitted before yours, and I'm guessing that was the issue. You'll need to resolve the conflicts before I can accept the Pull Request.

RhOmicron:
Took some time to Ctrl C/V the whole thing I did just becouse the order of two things was changed. Still found some more errors though..

PR #2109, opened 2016-12-23
Master

rocketheadent:
Added character sheet for TSR's Top Secret.


pjbehr87:
If you want the forum tag "Sheet Author" to show up for your account, you need to add your account ID to the sheet.json. You can find your ID in the URL of your player profile. In my url: https://app.roll20.net/users/141325/phil-b my ID is 141325.

PR #2110, opened 2016-12-24
Degenesis update

Thalar:
- Translate rolls
- Fix missing or wrong  translations
- Delete useless "special test"


pjbehr87:
You've got some merge conflicts. Please resolve the conflicts so that I can pull these changes.

Thalar:
OK It's fixed

pjbehr87:
Just FYI, you don't have to add the new strings to the alternate translations, unless you are also adding the actual translations for those strings as well (and then you still only have to add them for the language you are translating). Our translation service will automatically propagate all of those stings to the other files if they've been added to the translation.json.

PR #2117, opened 2016-12-29
shadowrun (advanced) fix

darkxyro:


darkxyro:
hey,
i saw you did a "fixed line endings" update to this sheet.
looking at the diff either on github (shows everything has changed) or on a 3rd party site (shows nothing has changed) did not really show me what exactly i am doing wrong.
Could you please explain?
thanks

pjbehr87:
Our git is set up to always use the same line ending/newline characters in all files. This means changing them to the default if they are different. This is what is happening here. The server our code runs on, so we use the Unix new-line character which is just LF (line feed or "\n")  , where as your code was using the windows new-line characters CRLF (carriage return + line feed or "\r\n").

The reason some editors will show all the lines changing (GitHub) and some editors will show no lines changing is because only whitespace has changed (all line endings) and some editors ignore whitespace changes.

If you would like to avoid this in the future, most text editors give you the option to change which line-endings are used. If you change the line endings to Unix, then your code will match our settings.

darkxyro:
alright done.

pjbehr87:
Perfect, thanks!

PR #2121, opened 2017-01-01
Added Hero Kids Sheet

Project462:
If I'm not doing this right, please let me know how to better upload entirely new sheets :)


pjbehr87:
The sheet has a few minor display issues, but other than that everything looks good. You had all the files and the sheet.json filled out correctly, so there isn't any "better" way than "correct" :P

Project462:
Display issues? If I can get some details I can get right on fixing those :)

PR #2128, opened 2017-01-03
Street Fighter BR

trueshinken:
Added Brazilian Portuguese version of the Street Fighter Storytelling Game character sheet


pjbehr87:
I'm sorry, but we [no longer accept new sheets](https://app.roll20.net/forum/post/3762625/policy-change-for-alternate-translation-character-sheets) that are just an alternate translation to an existing sheet. If you would like to add a new translation for this sheet, please update the original sheet to use Roll20's [sheet translation tools](https://wiki.roll20.net/Character_Sheet_i18n).

PR #2132, opened 2017-01-03
Update Only War C

RandomnessUK:
Rolls to hit now provide dropdowns for difficulty modifiers, prompting for bonuses from aim or attack range as well as final modifier dependant on other conditions, simplifies the process.

The base of the sheet now includes common weapon damage formulas for special effects such as tearing, proven, primitive and 1d5 weapons. These can be placed in the current "Damage" fields without requiring additional code. 

Position of Size dropdown now more aesthetically pleasing.


RandomnessUK:
Please let me know if anything further is needed to complete the merge.

PR #2135, opened 2017-01-05
Translation tags for SW FFG API

Orta15a:
- Implement translation tags "data-i18n" for multi-language support.
- French translation


Orta15a:
Hello Phil, @pjbehr87 

It seems during integration of the PR, the accented characters in the "fr.json" file was lost.
I don't know how it happened (encoding issue ?), but do you want me to do another PR to fix that ?

pjbehr87:
Oh, yes. I checked to make sure the translation wasn't breaking anything, and it's not, but since I don't know French I missed the fact that it was dropping letters. It's definitely an encoding issue, use UTF-8 for the translation file and that should work correctly.

PR #2137, opened 2017-01-06
I modified coc 7th sheet using i18n. 

roric-123:
At first, I made korean version of original sheet. Then I was informed Roll20 haven't been accepting alternate language sheet for months. To translate the sheet, I modified original sheet using i18n tool.


pjbehr87:
Now that you've updated the sheet to use the i18n tools, you can translate the ko.json file. In the translations folder for the CoC 7th Edition sheet.

PR #2145, opened 2017-01-09
Symbaroum added translation

vince-roll20:
- updated the html with necessary i18n data for translations
- created translation.json and necessary folder/file data
- worked with [Glandulf](https://app.roll20.net/users/1645621) to add French translation
- minor changes to html/css to help with translation
- changed/added armor quality to a text input as per user request. Prev code
was  `<select class="attr_armor1-quality">`  so attribute has never
actually been created.
- added two new repeating sections Boons and Burdens as per user request.
- updated sheet.json with translation credit


pjbehr87:
Thank you for updating the sheet and adding the alternate translations! One thing, to save us both time in the future, you only need the default translation file, "translation.json", and any file this is actually translated, in this case "translations/fr.json", all of the rest of them our system will generate automatically so you don't have to.

vince-roll20:
I wasn't sure since this sheet hadn't had any localization done to it.  I may have a German translation soon as well...  Thanks for your help Phil.

PR #2148, opened 2017-01-09
5th Edition OGL - Prompt for Physical Die Rolls

DejaAugustine:
* Adds **Dice Type** option to **Settings** tab
* When **Dice Type** is set to _Physical Dice_:
  - Replaces the virtual die roll with a prompt indicating which dice should be rolled
  - Applies all modifiers and reports the result using the standard templates
  - Adds a note to the roll template indicating that it's a prompted result
  - Supports all PC rolls (checks, attacks, damage, crit, etc)

**Note**: Physical Dice Prompts are not supported for NPCs

See this post in the Pro forums for documentation [[5th Edition OGL] Roll prompts for players who use physical dice (code provided)](https://app.roll20.net/forum/post/4450250/5th-edition-ogl-roll-prompts-for-players-who-use-physical-dice-code-provided)


pjbehr87:
I like the ideas of this change, it is definitely fun to actually "roll" your own dice, and will probably accept this PR eventually. But, this is a lot of changes to the #1 sheet used on the site (by a pretty large margin) and is going to take a bit of extra testing on our end before I can push it out to the masses. So, depending on when I can get around to doing that, it may be a few days or a week or 2 before that happens. I just wanted to warn you so you don't think you've been forgotten. 🙃

DejaAugustine:
I would expect nothing less. I'll keep watching the thread here in case you find any bugs or need me to make any adjustments, and keep testing it in my own games in the meantime.

Thanks!

pjbehr87:
Also, just to set expectations, we may end up not accepting these changes. They offer tools that are very different from pretty much everything else on the site. We push really hard to make sure everyone knows that everything on Roll20 is actually random, and this breaks that. If we did accept it, the rolls would have to be distinctive in some way so that everyone knew that it's a manually input number and that they can't "trust" it to be random. It can still go either way, we are discussing it.

DejaAugustine:
I had similar concerns when I added it for my players, so I did add a "(Prompted Roll)" note to the roll template for each roll. Also, 3D dice aren't rolled at all, so it's visually and aurally distinctive.

And whether it's accepted or not doesn't concern me much. I made a thing for my players and am offering it up to the community for consideration. 

I appreciate the time you're investing in reviewing it. :)

pjbehr87:
We've talked about it and I think we're going to pass on this. It's a really cool idea, but one that requires explanation, which is something we want to avoid with the official 5e sheet. We are trying to keep it as simple as possible. You can definitely still share your repo with others, I'm sure there are other out there what would love to use these options. I did for my game last week, and my players loved it. Thank you for sharing your work!

PR #2154, opened 2017-01-12
Update Rolltemplate and French

Orta15a:


Orta15a:
@pjbehr87 I seemed to have uploaded a version missing some translation for the Rolltemplate when converting to UTF-8... Now showing in red, sorry about that.
This fixes it.

PR #2163, opened 2017-01-14
Update to Mongoose Traveller 2nd Edition

Berserker66:
Updated Skills and Weapons to 2e.
Just added a third dice for Bone/Bane, there is probably a better way to do it, but it's functional.

As far as I can tell 2e doesn't have robotics yet, at least I don't have the book for it if it exists, so I just kept that as 1e.


Berserker66:
Merge Conflicts for some reason. I'll do it again tomorrow.

PR #2167, opened 2017-01-14
Mongoose Traveller 2nd Edition Update

Berserker66:
Mostly affects Skills and Weapons

Added Boon and Bane Die to rolls potentially affected by it.


Berserker66:
Now also added Auto Calculation of Attribute Modifiers.

PR #2168, opened 2017-01-14
Fallout 3.5

SilverRahl:
[eup.docx](https://github.com/Roll20/roll20-character-sheets/files/706028/eup.docx)


pjbehr87:
We already have a Fallout 3.0 sheet. Do the changes for 3.5 warrant having an entirely new, separate, sheet rather than just updating the other sheet? (if they do, that's fine, I just need to know)

SilverRahl:
Hello, i added 12 more skills but doing so deleted some. Changed all skills' values and per level SP calculation. So its completly new :)

SilverRahl:
Should i re-upload sheet.json for able to change the name?

pjbehr87:
Thanks for the information. There is one other issue, but since I didn't notice it last time for you to be able to fix, I will go ahead and fix it for you; the files are all in the root of the repo, rather than in their own folder. I'm going to move them to a Fallout 3.5 folder.

PR #2177, opened 2017-01-16
adding a new character sheet called Aborea

Scaby79:
this is a first fully functional version of a new character sheet for the german roleplaygame Aborea


pjbehr87:
I forgot to mention this in the other message, but you will want to upload the autocalc API script to the API repo as well.

PR #2179, opened 2017-01-17
Mongoose Traveller 2e: Make Independent sheet and fixes

Berserker66:
Also adds the default Template to skill rolls.
All rolls are now in the template.


pjbehr87:
I merged this, but it did not add back the skills that were removed previously, so I reverted it and went ahead with using Puppet's PR instead.

PR #2184, opened 2017-01-18
Impulse drive shipdamage

salmelo:
Just a little cleanup and updating to make the Impulse Drive sheet work a bit better for the latest version of the rules.


madadric:
Thanks @salmelo! My css and HTML skills are very lacking so your help is greatly appreciated.

PR #2189, opened 2017-01-18
Update all the is Mongoose

CoalPoweredPuppet:
Grumble.  This used to be easy.


Berserker66:
As this pull request stands, it removes template output for attribute rolls and attacks. Removes the area to mark down skill training progress. Changes the name of weight back to mass instead of KG (Guideline: Keep it Familiar; on the original Sheet it is marked as KG).

It will also merge conflict with this earlier pull request:
https://github.com/Roll20/roll20-character-sheets/pull/2179

PR #2209, opened 2017-01-22
Fixes for The Laundry character sheet

Lockbox313:
Fixed the Characteristics Rolls
Added a sheet worker for the Damage Bonus
Fixed the Science skill header
Fixed the headers for the weapons table
Added an auto-calc for Max San
Created a Current Sanity field (well renamed one of the madness threshold boxes")
Fixed the Major Wounds and Maximum Hit Points (regarding this there should be a note in the wiki about using - in an attribute name as this breaks sheet worker)
Added Current HP field.


The Laundry/translation.json
```diff
@@ -75,9 +82,10 @@
   "heavy-machine" : "Heavy Machine",
   "height" : "Height",
   "hide" : "Hide",
+  "hp" : "HP",
   "idea" : "Idea",
-  "influence" : "Influence",
-  "initial-san" : "Mythos (00%)",
+  "influence" : "Charisma",
```
pjbehr87: I don't think this is correct. Either the key is wrong or the text is wrong. If this is the auto-generated json then the text in the HTML is incorrect.

The Laundry/The_Laundry.html
```diff
@@ -1349,15 +1357,15 @@ <h3 style="text-align: center; color: #FFF; background-color: #000;"><span data-
...
-               <th><span data-i18n="artillary">%</span></th>
-               <th><span data-i18n="artillary">Weapon</span></th>
-               <th><span data-i18n="artillary">Damage</span></th>
-               <th><span data-i18n="artillary">Attack</span></th>
-               <th><span data-i18n="artillary">Special</span></th>
-               <th><span data-i18n="artillary">Range</span></th>
-               <th><span data-i18n="artillary">HP</span></th>
-               <th><span data-i18n="artillary">Ammo</span></th>
-               <th><span data-i18n="artillary">Siz</span></th>
+               <th><span data-i18n="Skill%">%</span></th>
```
pjbehr87: If you just have the % symbol, it shouldn't need to be translated.

The Laundry/The_Laundry.html
```diff
@@ -127,7 +127,7 @@ <h3 style="text-align: center; color: #FFF; background-color: #000;"><span data-
                             </td>
                             <td style="text-align: right"><span data-i18n="idea">Idea</span></td>
                             <td>
-                                <input type="number" name="attr_idea" value="@{int} * 5" readonly="readonly" />%
+                                <input type="number" name="attr_idea" value="@{int} * 5" disabled" />%
```
pjbehr87: The "disabled" key word has an extra " after it. I don't think this will actually break anything, but it should be fixed just in case. This is also the case for several changes below this.

Lockbox313:
Changes made

pjbehr87:
Awesome, you got it all done before I had finished everything else too. Thanks!

Lockbox313:
NP

Lockbox313:
Did something weird happen with the commit/merge for translation.josn ?   Charisma is showing as CHR, that was one of the last minte changes I made.

pjbehr87:
It might have. I'll see if I can fix it tomorrow.

Lockbox313:
Thank you

pjbehr87:
Sorry, I wasn't able to push out the changes today. I've already made the actual changes, I just lost track of time and didn't push them out in time. I have some other things that they will go out with tomorrow though.

Lockbox313:
NP, Its not really an issue  it is just CHR being displayed instead of
Charisma

pjbehr87:
It didn't actually get pushed out today either. I keep having other things pop up, that are time sensitive and have to keep changing gears. If it's not too big of a deal, it might be easiest to say it will go out with the rest of the sheet changes early next week. If I am finally able to get my other work done and out tomorrow it will go out then, but it will probably just wait till next week.

Lockbox313:
Next week is fine it is a minor issue.  Anyway the request for fixes were
the usual drive by.  You know the fixes get done and you get no response
from the requester.

PR #2218, opened 2017-01-24
VtR: removing need to healthbox images

CoalPoweredPuppet:
Removed the need to link images for the health boxes.


CoalPoweredPuppet:
Thanks!

PR #2225, opened 2017-01-26
Ghostbusters Sheet:  Added Ghost Die, Special Abilities; Fixed Talent Bug

amandasearle:
*Added the Ghost Die back into the game
*Added repeating fields for Ghost Special Abilities along with a roll button for each ability
*Fixed a bug with the Talent rolls where it wasn't correctly getting the appropriate trait and then adding the die sets together before rolling them.
*Set Talent score default to value of 3


pjbehr87:
There is already a DCC sheet. We try to keep it to 1 sheet per system, except for the super popular systems. Is this new sheet an upgraded version of the old sheet? Meaning can we remove the old sheet in place of this sheet without losing any functionality. Can someone using the old sheet switch to this sheet without losing anything?

cmengel:
Having used both the currently published sheet and this one, I find this sheet is much more functional than the current sheet available on Roll20.  I would love to see this added as an option for others to enjoy as well. Great work @amandasearle !

amandasearle:
The tabbed sheet is a much upgraded version of the old sheet and I tried to keep as many of the same variables names as I could, but I completely reworked the spellbook, the armor and the weapons sections which now use different naming conventions.

Simlasa:
This DCC character sheet has more features and is much more robust than the previous one. It's very well done. Thank you!

amandasearle:
For right now, closing this pull request to try and fix my respository.  I will be re-submitting this though.

PR #2230, opened 2017-01-28
DCC Tabbed Sheet:  This is a new DCC sheet with Tabs, Monster and 0-level character sheet

amandasearle:
It has been in development and playtesting since 2015 as an alternative to the original DCC sheet.  Most everything has roll templates.  Includes classes from Crawl! and Crawling Under a Broken Moon.  There are plans to add in additional classes along with a custom class.


pjbehr87:
This is a duplicate PR of #2225 . closing

PR #2231, opened 2017-01-28
W20 Fix and upgrade

CoalPoweredPuppet:
Fixed the player/name issue 
Added roll templates to pretty much everything.
Roll button for Gnosis, Rage, and all that are more visilbe
One click transformation is the name (Homid, Lupus, etc).  
Replaced health boxes, added wound penalties
Wound calculations all dice pools, weapon attack and damage, and claw weapon and damage.  Willpower, Gnosis, Rage and all soak roll are unaffected.
Put the Brawling combat stuff under a claw marker.
Rebuilt the common dice pools.  PUTTING THIS IN THE SHEET WILL REMOVE ALL PREVIOUS DICE POOLS.  And the players will have to replace them.  You can pick and choose if you want the query or not.  
Re-colored and renamed the attack and damage rolls, so its easier to find them.
...and I am sure I missed something.

### DIFF
--- a/CWOD-W20/W20.html
+++ b/CWOD-W20/W20.html
@@ -27,8 +27,8 @@
         <h4 style="line-height:28px">Chronicle:</h4>
     </div>
     <div class="sheet-col" style="width:155px">
-        <input type="text" name="attr_player" />
         <input type="text" name="attr_character_name" />
+        <input type="text" name="attr_player" />
         <input type="text" name="attr_Chronicle" />
     </div>
     <div class="sheet-col" style="width:100px">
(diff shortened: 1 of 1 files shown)


pjbehr87:
"PUTTING THIS IN THE SHEET WILL REMOVE ALL PREVIOUS DICE POOLS"
Are you saying that current users will have to do this if I merge this or is this an option that the users will have? If it's something that's just going to happen, is there any way you could avoid losing this information? Maybe through sheet workers.

CoalPoweredPuppet:
I will look into it. I can't remember exactly why things were different.

CoalPoweredPuppet:
Okay.  Closing this one, and opening on that will not screw with the dice pool names.  That should fix any loss of dice pool attributes.

PR #2237, opened 2017-01-29
[L5R] Fixes and Improvements

RhOmicron:
Added fields for Macro-Skills

Double Affinities added

Insight bonuses added for Courtier and Etiquette under its box.


VivalaHell:
Thank you for the charactersheet and the new updates.

PR #2240, opened 2017-01-31
W20 fix & upgrade (no attribute name changes)

CoalPoweredPuppet:
Fixed the player/name issue
Added roll templates to pretty much everything.
Roll button for Gnosis, Rage, and all that are more visilbe
One click transformation is the name (Homid, Lupus, etc).
Replaced health boxes, added wound penalties
Wound calculations all dice pools, weapon attack and damage, and claw weapon and damage. Willpower, Gnosis, Rage and all soak roll are unaffected.
Put the Brawling combat stuff under a claw marker.
Rebuilt the common dice pools. Should not damage the dice pools. You can pick and choose if you want the query or not.
Re-colored and renamed the attack and damage rolls, so its easier to find them.
...and I am sure I missed something.

### DIFF
--- a/CWOD-W20/W20.html
+++ b/CWOD-W20/W20.html
@@ -27,8 +27,8 @@
         <h4 style="line-height:28px">Chronicle:</h4>
     </div>
     <div class="sheet-col" style="width:155px">
-        <input type="text" name="attr_player" />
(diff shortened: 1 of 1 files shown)


pjbehr87:
Thanks for making that change for me!

PR #2246, opened 2017-02-01
DCC Tabbed Sheet: This is a new DCC sheet with Tabs, Monster and 0-level character sheet

amandasearle:
It has been in development and playtesting since 2015 as an alternative to the original DCC sheet. Most everything has roll templates. Includes classes from Crawl! and Crawling Under a Broken Moon. There are plans to add in additional classes along with a custom class.

There was a long discussion here about it - https://github.com/Roll20/roll20-character-sheets/pull/2225


amandasearle:
Going to try and make a migration script for this - pulling for the moment

PR #2261, opened 2017-02-04
Hellywood character sheet v 1.0 with i18n

DiChim:
Hellywood character sheet in english and french.


pjbehr87:
The sheet looks great! Thanks for all the hard work and the translations.

PR #2264, opened 2017-02-05
CSS for proposed 2E rolltemplates

aideenfb:
Accompanying CSS code to support the proposed 2Edefault rolltemplate.

### DIFF
--- a/AD&D_2E/2EStyle.css
+++ b/AD&D_2E/2EStyle.css
@@ -244,21 +244,22 @@ input[type="checkbox"]:checked + span::before
     background-color: none;
 }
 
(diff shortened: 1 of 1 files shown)


aideenfb:
I have several other rolltemplates which I use on my custom character sheet which I'd be happy to make available to all. If you'd like I'd be happy to add them and write up the necessary documentation for how they should be used.

My attack template has some custom graphics however, which are simply loaded into my campaign, but that's not much of a hurdle. I'd also need to add support for additional damage for critical hits as I don't use bonus damage dice on crits in my campaign.

PR #2273, opened 2017-02-05
Mongoose Traveller 2e: Very large update

Berserker66:
Should break nothing *fingers crossed*
More info in https://app.roll20.net/forum/post/4514941/mongoose-traveller


Berserker66:
I'd also like to use this opportunity to point out that on roll20 I do not have the sheet-creator tag. I did E-Mail you guys a week ago with a "we'll look into it" answer.

pjbehr87:
Sorry for the late response, I was out of the office on vacation all of last week. It's because in your game's sheet.json you have the field name "roll20userID" instead of "roll20userid". Fix that and it should assign the tag next time the script runs (early the next morning, after the sheet has been pushed out)

PR #2275, opened 2017-02-06
Open Legend: Attribute Sub & Guard

BuildAndDie:
### 1.7.7 on 2017 January 27th
* Added Attribute Substitution for all Attributes
- Fixed drop downs to have ALL attributes except for the current
attribute being looked at
* Updated all function calls to account for 1.7.6
- Including calculations to display prorperly if attribute subsitution
was being used

### 1.7.6 on 2017 January 24th
* Added new universal function for determining attribute substitution
values
- Reusable code so less lines of code
* Updated drop down menus to reflect change of Evasion > Guard
* Renames lines "Increase/Decrease Dice Size" to be "Attribute Dice
Bonus/Penalty"


BuildAndDie:
I don't know what happened that caused all these 5eshaped things to show up on my commit. NOt sure if that is going to cause a problem??

pjbehr87:
There are merge conflicts. This means that the branch that you did all of your work on was out of date. You either need to make a local copy of your changes, then either pull from the main branch or delete and re-fork the main branch, then add all of your changes back that should fix it.

PR #2283, opened 2017-02-12
[GURPS] Major Update

sdJasper:
Resized several fields on the "General" tab to make major Attributes Stand out, give room for new fields
Added 2 "blank" stats for user pools like Energy Reserve or Blood Points
Added Lifting ST and Striking ST
Added step values for attribute costs, so users can just press up or down and get the right cost per level.
Updated the "Current" line in the Encumbrance box to show the current encumbrance level, load, move, and dodge.
Changed many of the auto-caculated fields to be updated by helper scripts.
Added first time script to run when sheets are opened and populate values for formerly caculated fields. This will also initalize any "blank" values from reapeating sections that are missing the default values now.

### DIFF
--- a/GURPS/gurps.css
+++ b/GURPS/gurps.css
@@ -211,6 +211,7 @@ input.sheet-tab:checked::before {
     margin-top: -2px;
     padding: 0 2px;
     line-height: 22px;
+    text-align: center;
 
     background: transparent;
     border: none;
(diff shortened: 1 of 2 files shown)


calibraptor:
I found a lot of bugs in the Roll20 GURPS sheet that I've submitted some issues for. If it's not too much of a hassle would you be able to address them?

It's mostly just calculation errors.

PR #2290, opened 2017-02-15
[Pendragon] Dice Modifier - Weapon damage

RhOmicron:
Added a selection box if the weapon gets an extra 1d6, or if its a Dagger/Javelin and use 1 or 2 dice less respectively.


RhOmicron:
Commit d19c385 commited by mistake to two pull requests. I hope this wont conflict. This one is not supposed to be here.

PR #2299, opened 2017-02-18
Traveller

CoalPoweredPuppet:
Fixing gear (this should add inventory that uses the old inventory names and attributes)
Math fix on extra skills
Allies and what not fixed.


CoalPoweredPuppet:
Thanks again

PR #2301, opened 2017-02-18
[L5R] Minor fixes and alignment adjustment

RhOmicron:
Etiquette and Courtier added back on insight.

One td data typo found that broke alignment fixed.

Double Affinity added back to all spells and added to translations as well.

High skill repeating skill alignment adjustment.


RhOmicron:
Commit d19c385 commited by mistake to two pull requests. I hope this wont conflict. This one is not supposed to be here.

PR #2302, opened 2017-02-18
[L5R] update translation

RhOmicron:
Added Double Affinity


pjbehr87:
Just so you know: you don't have to update the files in the translations folder, unless you are adding an actual translation. If you are just adding more strings to the list/updating old strings, you can just add them to the translation.json file and the other files will automatically be generated with the new strings.

RhOmicron:
Good to know.

PR #2344, opened 2017-02-27
Fix ARM calculation for Iron Kingdoms RPG

wizard04wsu:
- remove irrelevant shield modifier fields
- correct the ARM calculation
- incidental fixes


wizard04wsu:
@CoalPoweredPuppet Do you see this causing any problems?

CoalPoweredPuppet:
Not really.  I'm a little annoyed some one choose to do this, but since it might help their game, I'm not complaining.  

Thanks for asking

pjbehr87:
You are renaming a bunch of files, mostly change the casing. The servers we use are Unix, and file name casing matters, so this would end up breaking all of the sheets. Please remove these changes.

PR #2346, opened 2017-02-27
New sheet for Blades in the Dark

joesinghaus:
This sheet really needs an upgrade to a better-looking version. It does not have the same field names as the existing one, however, so I did not want to overwrite it. We are talking to the authors about having it become a legacy sheet, so the view does not get cluttered by two sheets for Blades.


pjbehr87:
This sheet looks like a pretty big improvement over the other sheet. So, I'm going to deprecate the other sheet. Those who are currently using the sheet will still see it in the list, and can change if they want, but new players will only see this new sheet.

joesinghaus:
No, I don't think there's a reason to keep the old sheet, I have all the functionality covered. We have also talked to the authors of the old sheet, and they are fine with deprecating it.

PR #2355, opened 2017-02-28
[DSA4.1] Clean Up, Style, Fixes & Better Adherence to Official Rules

kreuvf:
I gathered some changes here, because often I would just rewrite large parts of the HTML file and I did not want to do an own PR for everything. Some stuff got overwritten during rebasing and squashing anyway.

I try to give a short overview:
* Style:
  * Overhaul of all roll templates (result colour-coded)
  * Use more similar colours everywhere (white/gray/black/brownish -> light-brown/brown/black)
  * Unify styles
  * Remove references to images (trademarked, copyright status unclear)
  * Use standard font
  * Fix tab issues
* HTML
  * Be way more standards-compliant
  * Move inline styles to CSS
* Features
  * Speed calculation takes agility (Gewandtheit) into account
  * Fix calculation of TP/KK calculation (additional damage with high strength)
  * Skill checks honour negative modifiers (= easy) correctly now
  * Add missing spells from official sources
  * Fix numerous non-working skills
  * Add tool tips in many places
  * Automate combat with roll templates
  * Min/max values for number input elements
* Documentation
  * readme.md contains a non-exhaustive list of known issues

I plan to use sheet worker scripts in the future to introduce an internally accessible version number for conversions from one version to later ones. Also, I would like to get rid of automatically calculated values whenever possible. Since some of the formulas I am using are complicated (or may not even be easy to understand) I am also planning to add a Markdown file just for that as I do not want to pollute the HTML file with comments.


pjbehr87:
The preview image change makes this look like a completely different game. Is this still for the same game? Do all of the attributes keep the same name? If the names change all of the existing games that use this sheet will lose data, and we need to avoid that if at all possible.

kreuvf:
I removed the [trademarked (at least in Germany) image for the system](https://register.dpma.de/DPMAregister/marke/register/3020120524516/DE?lang=en) and also the background image as I could not find any copyright information for reuse of the image on Roll20.

The system is still DSA 4.1, I have access to all the relevant rulebooks and I am working to improve adherence to the official rules. The preview image looks different as I opened a different (more meaningful) tab.

To the best of my knowledge, I am aware of only two attribute changes, which can be found in 7228396:
* attr_Spez_ got changed into (the meaningful) attr_Spez_karthographie
* attr_Spez_Tätowieren got changed into attr_Spez_Tatowieren to unify the attribute names and use ASCII characters only

Why I deem these two changes non-breaking:
* In this system, one can specialize in a skill to get a higher chance to succeed under special circumstances. For example, you could specialize your cartography skill in "caves and tunnels" and you would get a bonus of 2 every time you are mapping a cave/tunnel/dungeon.
* The specialization field is a free-form text field and is only used in the roll template to show that you were using a specialization.
* The attribute name of the actual skills (e.g. attr_taw_karthographie and attr_taw_tatowieren) were not changed, they still work as before.
* One can have several specializations, but you can only profit from one at each roll. The free-form text field (likely) is thought to carry information on all specializations at the same time, so every time you roll it, the template prints all specializations. In other words: the current solution is not very clean.
* I bet that nobody ever invested any experience points for buying specializations for the very unimportant skills cartography or tattooing (the two affected attributes).

So, should I try adding a sheet worker script? I wanted to save that for future PRs as I have nearly no experience with JS and I would add some more functionality while I'm at it.

Apart from the above attribute changes, everything else stayed the same/got improved a lot.

pjbehr87:
Thank you for the explanation. In that case, everything looks good to go then.

PR #2373, opened 2017-03-02
Runequest 3 changed encoding to UTF-8 and removed graphics in console log

Lockbox313:


Lockbox313:
Also a first release for Call of Cthulhu Dark Ages

PR #2377, opened 2017-03-03
CRLF to LF in Open Legend document

complexmath:
A file made it into the master repo with CRLF line endings. It would be awesome if this were changed so I didn't have to fight with git over its auto-conversion. Alternately, an exception could be made in .gitattributes for *.md files.


pjbehr87:
Thanks for fixing this. I usually fix this for all of the actual code files, but I must have missed the ReadMe.

PR #2380, opened 2017-03-04
Added character sheet for Drakar och Demoner 6.0

Erik-Johansson:
Both the sheet and the game system are in Swedish.
The character sheet is based on the Stars Without Number community sheet.


pjbehr87:
Just a few things to note. I had to make some, mostly "meta", changes to your sheet. You'll want to make sure your code is fully up-to-date with my changes before you make any further changes/fixes to the sheet or you're going to have merge issues.
The changes I made:
- Fixed the formatting/encoding of the sheet.json. The roll20id and roll20name both expect a string, rather than an array. The file was a weird encoding and had some hidden special character that was breaking thing, so I changed it to UTF-8.
- Changed the line endings from windows (CRLF) to unix (LF). Please update your text editor to use this same setting so that we don't constantly have the line-endings change back and forth every time code is committed.
- Changed the file encoding for the html to UTF-8 so that the language special characters would show up correctly.

PR #2384, opened 2017-03-06
[5eOGL]Added Spell Attack Bonus Modifier.html

Bryce-Pollema:
added a Spell Attack Bonus Modifier since there was one for the Spell Save DC but not the Spell Attack Bonus.

I don't know why that wasn't there already it was literally 5 lines and someone had make the same changes for the Spell Save DC earlier

I haven't done anything on GitHub before so I hope I did things correctly?

### DIFF
--- a/5th Edition OGL by Roll20/5th Edition OGL by Roll20.html
+++ b/5th Edition OGL by Roll20/5th Edition OGL by Roll20.html
@@ -1644,7 +1644,7 @@
                 <span class="label" data-i18n="spell-save-dc-u">SPELL SAVE DC</span>
             </div>
             <div class="part">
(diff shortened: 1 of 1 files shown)


pjbehr87:
Thank you for your suggestion, but I'm going to reject these changes at this time. We have tried to avoid adding too many options, like this, to the sheet for simplicity sake, but we have had this suggestion many, many times now. We planning a sheet update in the near future, and will consider adding this when we do that. For now, you can just manually add the spell bonus to each spell attack, which usually won't be an issue because you won't have a whole lot of them. If you are looking for more finer control over everything right now, you should check out the 5e Shaped sheet.

As for the fix being only 5 lines, this is not correct. There are translations and sheetworker updates to be handled as well as these changes.

PR #2394, opened 2017-03-10
A-BF: Reworked Sheed, templates, optimization, extension

Lagnalok:
Reworked Anima: Beyond Fantasy character sheet.

Tried to keep the sheet mostly the same, expanding options while displaying important things compactly.
Some parts where merged together, a conversion sheet worker handels upgrading from the old sheet so there shouldn't be any problems with that.

Most important changes are:
More attributes for details, temporary values etc.
Reworked most repeating sections and some other parts to include expanding sections for detailed editing and compact display for fast accessibility.
Roll templates for everything including power/spell cards with macro buttons for follow up actions.
Full integration of the translation tool - will do a German fan-based translation later (game has no official German translation)

I'm available for any fixes needed or upcoming changes.


pjbehr87:
Thank you for including the upgrade sheetworker!

wbaldwin6:
This update is excellent, but there is one major issue with it - there seems to be a mismatch that makes the value for Dex be used for Agi's opposed rolls, movement values, and ki amounts.

Lagnalok:
yep, the dexterity final for the button had the wrong name, causing the agility final to be overwritten.
Just started a pullrequest that will fix this.

PR #2396, opened 2017-03-11
Fixing errors

pstubbs1020:
There was some missing coding which broke the sheet. That was fixed and some more UI was added.


pjbehr87:
Can you please update your text editor to use Unix (LF) line-endings instead of Windows (CRLF) line-endings. Our system is Unix and expects the Unix line-endings.

pstubbs1020:
I use Notepad++/ I have no idea what you mean by use Unix (LF) instead of Windows (CRLF) but I will try.

pjbehr87:
http://stackoverflow.com/questions/8195839/choose-newline-character-in-notepad tells you where the option is to change this setting in Notepad++.

Windows uses \r\n for new lines, or CRLF (Carriage Return + Line Feed)
Unix uses just \n for new lines, or LF (Line Feed)

The system that hosts Roll20 is Unix, and expects the Unix line endings.

PR #2404, opened 2017-03-13
CoC DA fixes Runequest 3 New Features

Lockbox313:
COC DA fixed double button icons added support for Cthulhu Invictus
Runequest 3 Added fieldsets for preception and stealth skills, automatic hit location roll.  Two new hit locations 4-legged and Walktapus. Fix for thrown wepon damage bonus in repeating sections when str or siz changed.


Lockbox313:
Phil,

I am getting this when I try to load a game with the Runequest 3 sheet

Failed to load resource: the server responded with a status of 500
(Internal Server Error)
loading.js:1 Uncaught ReferenceError: d20ext is not defined
    at loading.js:1
app.js?1488996874:26 Uncaught TypeError: Cannot read property 'siteURL' of
undefined
    at app.js?1488996874:26
    at app.js?1488996874:54
tutorial_tips.js:7 tuts loaded
https://app.roll20.net/editor/soundmanager2_debug.swf Failed to load
resource: the server responded with a status of 404 (Not Found)


It does not happen on any other game.

pjbehr87:
It looks like you deleted the Runequest3 html and css. If you could add those back with your fixes I'll pull it asap.

Lockbox313:
Phil,

What the... I could have sworn I uploaded those again. I added them  and
created a pull request.  Sorry for my stupidity .

Lockbox313:
Phil,

Anyway of finding out why css is being ignored on both the Runequest 3
sheet and The Cthulhu Dark Ages sheet

On the Runequest 3 the second  column of button with a class of
"sheet-new_roll" should show -? instead of the D20 icon. buttons on the
basic tab have a class "sheet-new_roll2"and "display correctly.

On the Cthulhu Dark Age sheet the blank skills on the sheet have a class of
"sheet-blank_skill" from what I see in the console it is in the styles

﻿.charsheet .sheet-blank_skill{
width:128px;
}

it is there but it is being ignored.

pjbehr87:
This took me 45 to figure out, but it's because of the file encoding. The current encoding is "UTF-8 with [BOM](https://en.wikipedia.org/wiki/Byte_order_mark)" which puts some, invisible, special characters at the beginning of the file. Which are then included in the first CSS rule, causing the rule not to match anything. In your case, the first rule is the one that is setting the content of the button. In my text editor, Sublime Text, I changed the encoding from "UTF-8 with BOM" to just "UTF-8" and everything started working. Your text editor might have "UTF-8" and "UTF-8 without BOM" as options, in this case you will want to select the 2nd one.

Lockbox313:
Yeah I figured it out when I pulled the html from the chrome console and
pasted it into notepad and got the #code at the start . I uploaded new
versions set to utf-8 without bom and made a pull request. Thank you for
taking a look at it.

pjbehr87:
When we run into weird "that doesn't make any sense, it should work" kind of issues, 90% of the time it's a file encoding thing. Your safest bet is to use UTF-8 without BOM for everything. It supports FAR more special characters than the standard windows encoding.

PR #2423, opened 2017-03-15
Attempt at roll template translations

aunderwo:


agjwilson:
No sorry. It did not work. :(

aunderwo:
@agjwilson Can you remember what the error was?

aunderwo:
@agjwilson and @antila I think I have found the source of the error and updated the pull request. Are you able to check please? Thanks

agjwilson:
I will test this today.

agjwilson:
That works !! Well done !

There are two small issues. The first being that Pilot (Skill) is not responding on my sheet. Is it not linked to an attr_ 
The second would be for a graphical change (see picture) it looks odd, so I can correct this if you can give me the line of code where the tables for these chat roll templates are.
correction01

aunderwo:
@agjwilson Thanks for testing and spotting the pilot error. Hopefully corrected now :)
The lines that define the roll template styles are on 4 - 32 of Coriolis.html with the css in lines 296 -360 of Coriolis.css

agjwilson:
Excellent. 

Do you think there is any room in the future to accommodate a DM sheet for NPCs and Dark Points?

aunderwo:
I think that would be useful but am new to this roll20 thing and not quite sure how it would work. @antila do you know if this is possible. I think the issue is that this sheet is per character.

agjwilson:
One thing I noticed on your correction of the pilot is that its not truly making the roll an automatic fail.

rolltemplates

agjwilson:
The sheet is per character. True. 

The one way I am thinking is that the Ship sheet could be put on the Character Sheet page. Then the Ship Sheet be made into a DM Sheet for the DM to have his stuff on. Its only an idea, and I have no time to start this currently. Perhaps a couple of weeks if we can agree to something.

agjwilson:
@aunderwo Let me know if you can help. I am having real problems trying to get these on the same lines as different sizes 
working
I think we could have the same for skills, where the skill name is a larger font than the abbreviated related attribute in parentheses.
I can get them to change font and stay on the same line. If I use an inline block tag then the table increases its width size.  

I actually went for a haircut just now so I wouldn't have to pull my hair out on this one.

aunderwo:
This is because I think those are wrapped in h5 tags. 
By default, both h tags are  "block elements". That means they make space above and below to ensure they are the only thing on a given line. You can override this within the css to make h5 display inline
```
h5 { display:inline }
```

or you can wrap the STRESSPÖANG and (SKP + KNS) in span tags (which are inline by default) and give the span a class that matches h5 style

```
<span class='attribute_large_header' data-i18n="mind-points">Stresspo&auml;ng (shp+kns)</span><span class='attribute_small_header' data-i18n="wits-empathy"><!---(shp+kns)---></span>
```
 in Coriolis.html

```
.attribute_large_header {
    text-transform: uppercase;
    margin: 0;
    padding: 2px 6px;
    font-family: 'Droid Sans', sans-serif;
    font-size: 11px;
    font-weight: bold;
    color: #000;
}
.attribute_small_header {
    padding: 2px 6px;
    font-family: 'Droid Sans', sans-serif;
    font-size: 9px;
    font-weight: bold;
    color: #000;
}
```
in Coriolis.css

pjbehr87:
To have a separate "DM" sheet, you would need to use the API, which is only available to Pro accounts. You can see an example of this in the FFG Star Wars (API-compatible) sheet. They have a "- Dice Pool -" character that gets created, that allows the DM to manage things, like what dice are rolled, across all other sheets. It also has a "Ship" portion of the sheet that is shared between all Players. Both of these things are handled via the API.

Coriolis/Coriolis.html
```diff
@@ -95,35 +95,35 @@ <h3 data-i18n="attributes">Grundegenskaper</h3>
                             <h4 data-i18n="strength">Styrka</h4>
                         </td>
                         <td class="sheet-value">
-                            <button type='roll' value='&{template:coriolis} {{name=Styrka}} {{roll=[[@{sty}d6>6]]}}' name='roll_str'></button>
-                            <input type="number" name="attr_sty" value="0">
+                            <button type='roll' value='&{template:coriolis} {{name=^{strength}}} {{roll=[[@{strength}d6>6]]}}' name='roll_strength'></button>
+                            <input type="number" name="attr_strength" value="0">
```
pjbehr87: You are changing the names of a bunch of attributes in the commit. This will end up erasing a bunch of values for anyone that is already using the sheet since the reference to the value they have is changing out from under them. We want to avoid this.

You have 2 options:
- Leave the attributes in Swedish.
- Add a trigger in the sheetworkers that will update all of the values from the old attr to the new one.

aunderwo:
I have changed the attr values back as suggested

agjwilson:
Okay now this has merged. I am going to but upload the graphics to GitHub and I'll be ready to do a pull request.

agjwilson:
Is there a way to hide dice results on auto-fails ?

aunderwo:
@agjwilson Just checked on roll20. This seems to have broken translation of rolls. Is it the case for you too?

agjwilson:
@aunderwo Yes you are correct. The roll templates are now back to Swedish.

Also, I did try inline blocks, unfortunately this method of keeping text to one line just stretches out the headers and tables across the sheet, which isn't ideal. The other method is to make the text smaller to fit on lines, but that's not really the solution I was looking at.

PR #2443, opened 2017-03-20
[Pathfinder] v1.19

plutosdad:
compatible with Vince's update (his is only for preview image, this one is the code)
1. Now hp updates when updating max hp again
1. Fix to attacks menus
1. Fix skill stealth-trait field
1. Fix to spell-notes
1. Fixes to NPC parsing:
    * CMD and CMB npc parsing
    * Fix fly and stealth npc import take size into account
1. Additional default settings
1. To re-upgrade from a backup of v.74:
    * Attack and skill macros will not reset if you re-upgrade backup of v.74
    * Fix to npc upgrade from v.74 will not overwrite any class hp or levels


plutosdad:
I combined the sheet.json updates from https://github.com/Roll20/roll20-character-sheets/pull/2435  into this, and did a merge of Vince's fork with mine, so it should say there are no conflicts even after you do his (assuming you do his PR first) .

PR #2480, opened 2017-03-28
shadowrun (advanced) update

darkxyro:


Shadowrun5th/Shadowrun5th.html
```diff
@@ -12455,23 +12533,14 @@
 on("change:repeating_knowledge_knowledgedicepool",function(){
     getAttrs(["repeating_knowledge_knowledgespec"], function(value){
         if(value.repeating_knowledge_knowledgespec==0){
+            console.log("dafuq")
```
pjbehr87: I don't think you meant to leave this line in here.

PR #2526, opened 2017-04-07
[L5R] Auto calculation repeatables

RhOmicron:
Added auto calculations for up to five skills in each tree.

Added width to repeatables so they don't extend with window.


MindlessCalm:
Hello, I was wondering if you had a functioning sheet for L5R, the one on roll20 has many issues.

RhOmicron:
It could be that you tried using the sheet as it was being merged with roll20. As it loads fine for me and things seems to be in order.

PR #2527, opened 2017-04-08
Basic Edits

Kittyzero:
Changed the sheet to replace drop boxes that wouldn't save with text
input.


pjbehr87:
If you want to keep the dropdown, but make it so that they actually save the data, you can add the name="attr_..." to the select fields.

PR #2536, opened 2017-04-12
Partial DFA.html revert

rentarded:
The change broke the roll buttons, but fixed the Fate Point calculations.

Rolling back the changes to the approaches.

Changed Mantle approach to fixed text because it never saved what was inside the text box. Also changed it to roll specifically for the mantle instead of Intellect as a fix from the first version.

I'm have no idea how to create flexible, saving text boxes for the approaches and still have them work, so this might be the next best option. Hopefully you (Ikidaore) or someone else can figure out a fix.


pjbehr87:
If you want a Git user to be notified about a message you have to put an @ before their name. Like: @Ikidaore

PR #2537, opened 2017-04-12
completely reworked character sheet

gigxli:
- resedigned html layout
- added more fields/properties
- added worker script to calculate many fields
- added translation
- added german language (main audience)
- added repeating sections for weapon, armor, spells, equipment and talents
- added roll buttons for typical checks and attacks


pjbehr87:
Your images won't work with the URLs as they are now, they need to use the following format to link to the image on GitHub:
https://raw.githubusercontent.com/Roll20/roll20-character-sheets/master/Dungeon-slayers-4/DS4-DEF.png

You can find the link for each image by navigating to the image on GitHub and clicking the "Download" button at the top right.

pjbehr87:
Also, since this is a complete rework of the sheet I have to ask, is the new sheet completely compatible with the old version. Will all of the games that were using the old version continue to function, without any issues, with the new sheet?

gigxli:
Thank you for pointing this out. I have fixed the URLs for the images. Is there anything else I should update before the next pull request?

I'm afraid, it is not compatible with the old version. The reason is for example that the previous sheet has had a fixed amount of entries for weapon, armor and items and I changed that to repeating sections. Additionally, I have added new fields that allow more flexibility but this changed how some fields are being calculated, so I guess some values (attributes) would get lost when migrating.

pjbehr87:
Alright, I'm going to have to revert this then. We try, as much as we can, to avoid changing things out from under sheets. But, since this seems to be a huge upgrade over the first sheet, I would like to replace the old one with this sheet. We just have to go about it a different way. We are going to create a brand new sheet, and just hide the option for the old sheet. That way the old sheet still exists, and all of the games that are currently using it will continue to use it and not lose any data. But, any new games will use the new sheet.

The way we will do this is: take all of your changes and put them in a brand new folder, and leave the old sheet as-is. That way it's a completely separate sheet. You will have to update the URLs again to point at this new folder on GitHub, but that should be an easy find/replace.

gigxli:
I understand, this is perfectly fine for me. Let me know when the new folder has been created or if I should create a new folder, I will then recreate all the changes and adjust the links accordingly. Thanks for your help by the way!

pjbehr87:
You can create the new folder, just add it like you would a new sheet rather than making changes to the old one.

gigxli:
Done. Let me know if there is anything else I can do.

PR #2538, opened 2017-04-12
Update de.json

drbleifuss:
Added a few more german words.


pjbehr87:
Thank you for your translation work!

PR #2544, opened 2017-04-17
Added Wises

ptasty:
Just a small addition to give a space for keeping track of Wises


MouseGuard/mouseguard.html
```diff
@@ -33,6 +33,9 @@
                         <td><label for="attr_cloak_color">Cloak Color:</label></td>
                         <td><input type="text" name="attr_cloak_color" /></td>
                     </tr>
+					<tr>
+						<td><label for"attr_specialty">Specialty:</label/</td>
+						<td><input type="text" name="attr_specialty" /></td>
```
pjbehr87: You're not closing your <tr> here. This should be fine, since it is the last row in the table, but it should be closed nonetheless.

PR #2546, opened 2017-04-17
Small JSON fix to Mage2e sheet

tmtabor:


pjbehr87:
Thanks for catching this so quick!

PR #2553, opened 2017-04-18
[Pendragon]Fixed NPC and Horse values. New Background.

RhOmicron:
NPC and Horse values same, breaking stats.

New Background.


Lockbox313:
There appears to be some problem with the newimage link.

RhOmicron:
There seems to be code missing from my testing and I can't seem to get it into a pull request because phone copying doesn't work in roll20 html layout(great...) I won't have access to a computer until Friday.

RhOmicron:
It is the dividers for the background missing amongst other things which I have no idea.

Lockbox313:
There is an image tag with its source looking for p-titre.jpg which is not
in the image folder

RhOmicron:
I see, the p-titre image part of the code has been removed in the code I have in testing. I somehow forgot to paste in this in my last pull request.

This will be fixed as soon as possible.

PR #2580, opened 2017-04-26
Add Savage Worlds Spanish character sheet

Seregras:
This is a Savage Worlds character sheet in spanish, simplified, and that includes auto buttons to calculate rises in fight and damage rolls (after comparing with parry and toughness).


pjbehr87:
I'm sorry, but we [no longer accept new sheets that are just an alternate translation to an existing sheet](https://app.roll20.net/forum/post/3762625/policy-change-for-alternate-translation-character-sheets). If you would like to add a new translation for this sheet, please update the original sheet to use [Roll20's sheet translation tools](https://wiki.roll20.net/Character_Sheet_i18n).

Seregras:
I didn't understand, this was not just a translation, it included new functionalities that the english savage worlds sheet doesn't include. If I prepare this in english and then translate it to spanish will it be valid? Thanks

pjbehr87:
Sorry, since this is brand new I can't see "changes" only that the whole file is new.

You are correct. If you want to make these updates, and add the Spanish translation, you'll need to make the updates to the original sheet and hook it up to the translation tools to add the new translation. We are trying to eliminate the need for users to use separate sheets for different translation, unifying everything into a single sheet that is the "same" for everyone except that it will use the user's selected language.

Seregras:
Hi,

Is a complete rework. Is so different that it calculates the rises on skill rolls and injuries on damage rolls (no other existing sheet does that). Also, it includes an attacks section, that uses token targets to calculate successful attacks. What I will do then is to include the translation tools to have it in English and Spanish, and allow further translations, and create another pull request. 

Thanks for your help

pjbehr87:
Perfect, thank you for your work in upgrading AND translating this sheet!

PR #2583, opened 2017-04-28
Paul 'Havoc' Stein alternate WFRP2E character sheet

timothystone:
Maintainers, I took a stab at the `approved.yaml`

This is an alternate API driven WFRP2e character sheet by Paul "Havoc" Stein (55273). Several forum members looked for this sheet to be added, and working with Paul, I obliged.


pjbehr87:
You'll need to add the API scripts to the repo, [here](https://github.com/Roll20/roll20-api-scripts), so that I can add it to the one-click-install list.

PR #2586, opened 2017-04-29
Thalya

Zakarik:
A New sheet for a RPG custom create by my friend.


pjbehr87:
The official character sheet list is for systems that have a publicly available rule set. I will add your sheet to the repo, so that you can have it in the repo, but I'm not going to add it to the list since it is just a personal, custom system.

Zakarik:
Thanks. But the rule set is publicly available. This game has a new forum :
http://thalya.forumactif.com/ and the book is here : https://www.dropbox.com/s/nia21t9kvm984o3/LdB%20Thalya%20Joueurs%20V1.pdf?dl=0

This game has a facebook too : https://www.facebook.com/jdr.Thalya/

PR #2587, opened 2017-05-01
Updated Chronicles of Darkness sheets

ElvesofZion:
Based on the "Official" Chronicles of Darkness sheets. At the request of forum users on Onyx Paths website I have added the missing second editions game lines.


pjbehr87:
I don't think we want to create a whole new sheet. Could you make these changes to the actual official sheet, so that we can see the changes you are making? If approved there we will just make changes to the existing sheet rather than making a 2nd, almost identical sheet.

ElvesofZion:
I agree that it should be the same sheet. One of my changes was to update the name to the actual official name of the game, so it is showing as a new sheet. Is there a way to change the name of the current one without duping it?

pjbehr87:
Just take your code can paste it over the existing sheet. I don't know about changing the name, I'll have to ask Steve (one of the other Dev and author of the other sheet). With official content, this stuff is important.

ElvesofZion:
Ok, it should overwrite the old version now rather than create a dupe. I'll check on the naming stuff after this update goes in.

Another thing is that the Copyright on the sheet is extremely out of date. Since this is an official sheet I would think that there should be an official copyright it should have. I will also ask on Onyx Path's forums for an appropriate copyright if roll20 doesn't have it.

pjbehr87:
Thanks for working with me on this. The changes look good.

PR #2590, opened 2017-05-02
A-BF: Version 1.1.0

Lagnalok:
Restructured and brighter layout.
Changed Templates to go with the style form the new layout.
Option page to set whipser behaivour as well as open roll and fumble ranges.
Option page holds Roll Buttons for simple openroll/fumble rolling.
Some minor Fixes.


pjbehr87:
This looks like a pretty massive update. Does this new version maintain the attributes of the old sheet? I want to make sure that not too many attributes have changed so that existing games will not lose data when the sheet is updated.

Lagnalok:
Alot of code had to moved and expanded for the new layout. From Version 1.0.0 to 1.1.0 all attributes have been kept and only new ones added. All new Attributes have default values set so modified autocalculation will stay as before the update. So there was no need to modify the update sheet-worker. 

In addition to fully testing the new sheet I've also done conversion Testing from v1.0.0 to minimize risks of errors.

pjbehr87:
That's better than I could have even hoped for! Thank you for doing all of this testing and confirming it for me. I wasn't able to make the push this morning, so you didn't miss anything. Your changes will go out tomorrow.

PR #2593, opened 2017-05-03
Updated instructions

dougansf:


pjbehr87:

PR #2595, opened 2017-05-03
Removed Targetting

Meduna:
Targetting has been removed from all attacks, abilities and spells. This
means you will no longer need to stat up every monster in roll20,
however, it also means you will have to look up/memorize the
armor/m.armor values of characters. This will also affect macros that
reference your target's eva/m.eva to determine your CoS. I'm considering
uploading an alternate version of the sheet with targetting intact for
those who prefer it, once I figure out how to resolve damage
automatically.


pjbehr87:
Rather than have 2 totally separate, nearly identical, sheets. Can you please add this an an option within the sheet, rather than make a 2nd sheet. You can add an "option" tab or, since it's just a single option, add it elsewhere on the sheet.

pjbehr87:
Closing this PR since no changes were made. If the changes are made please either reopen this PR or create a new one.

PR #2599, opened 2017-05-06
A-BF: Corrected Fumble Range, Chrome Fix

Lagnalok:
Changed Fumble Range default to 3 (Corrected value).
Fixed Roll buttons without icons not showing properly in Chrome.

Note: Do I need to update the files under /translations with the changes from the translation.json made with the last merge? They are missing for the current sheet on roll20.


pjbehr87:
You shouldn't have to, I have a script that should automatically propagate the changes to all of the individual languages. I might have missed your sheet last week, I'll be sure to double check and make sure all the changes make it this time.

PR #2610, opened 2017-05-09
[Pathfinder] v1.51

plutosdad:
Offhand damage multiplier and number of attacks fixed
Create attack from equipment fixed - does not create empty attack
Drag and drop equipment fixed
More macro fields exposed for customization
Show more fields checkboxes added locally not only one global field


plutosdad:
ok i don't understand these conflicts they are not even in anything I touch

PR #2619, opened 2017-05-14
Bw tweaks

aldergames:
Hello! I've noticed that some parts of the Burning Wheel sheet are not 100% compliant with the rules in the BWG core book. I'd be willing to make some updates to the more complex calculations involved, but thought I'd start with something simple, as I'm not familiar with the API yet.

This pull request contains a change of a auto-calculated attribute Hesitation to a Script Worker, which does some slightly more complex calculations on the value. I've referred to the wiki entries on making sheets for these changes, as well as the examples provided. I don't have a Pro account, but is there some other way to test/verify my code before bothering people here about it?


SteveKoontz:
There is not currently a way to test changes without access to custom character sheets. That's something we intend to change in the future.

I'm fine with the change, except that instead of triggering on open every time, the function should only be triggered by the change of will/will_shade and instead start with the default value of 10-@{will} by default. On character creation it'll be set to black and be correct. Then if any changes are made the sheet worker will pick up with the new logic. This keeps the function from running unnecessarily every time the sheet it opened.

aldergames:
To do that, would I set value="10-@{will}" or title="10-@{will}"? Since I don't know what the backend looks like, I don't know if the auto-calculations would interfere with the script worker somehow.

SteveKoontz:
That's exactly right, and not you don't need to remove the disabled. Once a new value gets set for the attribute it will overwrite the default value. So if a static number like 8 gets put as the value it will overwrite the calculation and the disabled math calculation would return 8 for the calculation of 8 and you'll get the result you want.

aquaalex:
Maybe Roll20 can allow people to create custom games without a PRO account but then you can not invite any players to the game. This will allow people without PRO account to make sheets and scripts, and also give them some incentive to become Pro members :)

PR #2624, opened 2017-05-15
Update Mass Effect.html

SilverRahl:


Mass Effect/Mass Effect.html
```diff
@@ -132,7 +132,7 @@
 <input type="radio" class="sheet-tab sheet-tab2" name="attr_core-tab" value="2" title="Combat" /> 
 <span class="sheet-tab sheet-tab2" style='line-height: 20px;'>Combat</span>
 <input type="radio" class="sheet-tab sheet-tab3" name="attr_core-tab" value="3" title="Feats & Specs" />
```
kreuvf: The ampersand should be replaced with &amp;amp;. The next line and possibly some more are affected as well.
See http://stackoverflow.com/a/3493425.

PR #2643, opened 2017-05-26
Dark Heresy 1ed improvments

Orta15a:
Translation french
Command basic skill
Roll template for skills
Style for roll template


Orta15a:
added roll template for Weapons and Psy rolls

PR #2644, opened 2017-05-26
Master

plexsoup:
Updated The Veil and The Black Hack.
Added two new playbooks to the Veil.
Added Class field and Notes to the Black Hack.


pjbehr87:
Your branch was out of date when you made your changes, so now there are merge conflicts with all of the translation updates from last week. If you could update your stuff and reopen the pull request I would appreciate it.

PR #2650, opened 2017-05-29
Star Wars Revised RPG Character Sheet submission

traec:
This is my submission for the Star Wars Revised RPG Character Sheet. It's works really well and I've done quite a bit of testing for bugs. It's very similar to the one in the core rulebook.


traec:
Finally figured out and added integrated weapon attack/combat rolls. Will repull with updated files

PR #2656, opened 2017-05-31
[Burning Wheel] Reflexes and Mortalwound

aldergames:
Adding script workers for Reflexes and Mortalwound attributes. Shade math is taken from the Monster Burner pg 367, the chapter on Gray and White Math. Also added a checkbox for the Tough trait, which affect Mortalwound calculations.

These changes disable the dropdown for shade for Reflexes and Mortalwound, since they are automatically calculated from the base stats.

### DIFF
--- a/Burning Wheel/Burning Wheel.html
+++ b/Burning Wheel/Burning Wheel.html
@@ -452,7 +452,7 @@
     <tr>
         <td class="statname" title="@{reflexes}">Reflexes</td>
         <td>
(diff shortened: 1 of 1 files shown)


pjbehr87:
I had to revert the changes because the javascript had errors and there were other errors caused by using the disabled auto-calc field on a select, which isn't valid. If you can fix these issues and create another pull request I can try and merge it tomorrow.

aldergames:
Mm, I just saw the errors and I've made a fix for it. Didn't know the disabled tag would mess with things (it's a regular html tag, that does work on select too, but I guess you are doing some special stuff with it for auto-calc).

Sorry if this stuff is a bit trial-and-error, but without a pro account I can't really test the sheet properly.

PR #2660, opened 2017-05-31
Created a new character sheet for Stars Without Number Revised Edition

Karlinator:
Stars Without Number is getting a revised edition, and most of the rules are now Draft Final, meaning they probably won't change before release, and almost all of them are in there. So I cloned the character sheet from Stars Without Number 1st Edition and rewrote it for Revised Edition.


kevinsearle:
Just want to verify for Roll20 that this is, indeed, for the new upcoming edition.

pjbehr87:
Thanks for chiming in @kevinsearle

PR #2667, opened 2017-06-03
Adding new sheet (Splittermond)

ap3h3ad:
I've created a new character sheet for the german RPG Splittermond and would be glad to see it in public use. :)


pjbehr87:
There are some characters that aren't showing up. The easiest way to fix this is to change the file encoding, of your CSS and HTML files, to "UTF-8 without BOM" (if your text editor has 2 UTF-8 options one with be with and one with be without the extra [BOM](https://en.wikipedia.org/wiki/Byte_order_mark) character, if there is only one option it will probably be the correct option by default).

ap3h3ad:
Hi!

Thanks for your feedback. I fixed the issue and made a new pull request.

PR #2675, opened 2017-06-06
Fixed Encoding problems

ap3h3ad:
I fixed the encoding problems and added some last minute-usability.


ap3h3ad:
Hi,

just out of curiosity: Is the reviewing process done with merging my branch into the master branch and can I expect that my sheet will be available for public use in the future or will there be another reviewing with separate feedback?

Greetings

PR #2676, opened 2017-06-06
Barbarians of Lemuria v1.4 + various readme files corrections + link additions.

NathaTerrien:
Updated with official german translation


pjbehr87:
We have a special setup for Patreon, that allows sheet creators to link to their patreon. It has a special disclaimer that we require. I will add functionality to allow you to link your Tipeee account officially. Until then, we ask that you don't link to it yet if that's alright.

NathaTerrien:
@pjbehr87 
I already checked with Stephanie (team@roll20.net) to be able to add links to my Tipeee page. Do you want me to forward you her e-mail?
As the sheet.json seems to only allow Patron URL, I assumed I could add the link in the description.

pjbehr87:
Oh, okay. She's off today, so I didn't know. I'll go ahead and merge this, and you can update once we make the changes to add your Tipeee link officially.

NathaTerrien:
Great! Thanks (again ;) Phil.

NathaTerrien:
PS : @pjbehr87  Is there a full documentation somewhere about what "field" are supported in the sheet.json file?

pjbehr87:
There isn't. But, it's mostly on purpose. We didn't just want everyone making a patreon and were keeping the list a little more exclusive and special. While we don't keep secret, it's not common knowledge. The way it will work is adding a "tipeee": "[tipeee link]" to your Json. You can add it now, it just won't show up until we get the update in on our end. I'm thinking it will probably happen some time before next week's pull.

NathaTerrien:
Ok, thanks.

PR #2680, opened 2017-06-10
[Pendragon] Battle size modifier fix

RhOmicron:
Battle size modifier change to the correct ones.

### DIFF
--- a/Pendragon/Pendragon.html
+++ b/Pendragon/Pendragon.html
@@ -1975,11 +1975,9 @@
 	   <br/>
 		<label style='display:inline-block; '>x Battle Size Mod.</label>
 		<select name="attr_BRBModifier" style='display:inline-block; vertical-align:top; width:160px; margin:0px; padding:0px;'>
-			<option value="2">Enemy 5:1 - x2</option>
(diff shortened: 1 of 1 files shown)


Solarswordsman:
You're reducing the number of options for this attribute as well as changing their value.  Existing games with the previous values are going to default to 1.  Is it worth the change for this sheet to lose that value?

RhOmicron:
This change is in accordance with book of battles 2nd ed, page 81. Where only 1x, 2x and 3x modifier exist. I assume you have recorded the battles glory in the glory tab as well?

Solarswordsman:
Sorry, I'm not familiar with the system.  I just want to make sure people with the old version of the sheet aren't losing functionality (e.g. might they be using the old modifiers, and still want to?) and that this attribute change won't be a destructive change to existing sheets.

RhOmicron:
It shouldn't change much unless they get uncertain how much they earned last battle and didn't record it in the appropriate tab. 

It would be easy enough for them to know which modifier they are supposed to have looking at how big army they were against.

Solarswordsman:
Ok, thanks for clarifying.  Looks good.

PR #2682, opened 2017-06-11
Translate Warhammer v2 sheet to french

Nhacsam:


pjbehr87:
You added a bunch of translation strings to the French version that will be lost because they were not added to the translation.json base file. It looks like the HTML already includes these keys, and they are missing from all other languages. If you could add these keys, in english, to the "translation.json" file, it would fix this issue.
The following keys will be removed:
special-rules
trained
poor
medium
good
best
shoot
throw
half
full
spell-name
encumbrance-choice1
encumbrance-choice2

Nhacsam:
Thanks, I will do that in the next PR,
There still are some text which is not translated

PR #2683, opened 2017-06-12
Blades in the Dark v1.4

joesinghaus:


Solarswordsman:
How will this affect existing players using the previous version?  There seems to be significant changes to some attributes, will this transition smoothly to the new version, and if it doesn't, how significant of an impact will it have?  You might consider writing a sheet worker to update older sheets.

joesinghaus:
It's all covered, there's an update sheet worker. I try to be very careful and never break anyone's old sheets.

joesinghaus:
Also, you're not Phil, hi there new Roll20 person :).

Solarswordsman:
Ah, my mistake!  I see you've got that handled.  Also, hi there :)!  I'm Jeff.

PR #2686, opened 2017-06-13
Werewolf 20 Updates

pensman:
- Moved Dicepool header so it's centered.
- Standard difficulty is 6, so attack difficulty has been defaulted to 6. 
- When rolling damage, wound penalties should not be applied. These have been removed from damage calculations in combat.
- Only extra successes above those required to hit should be applied as extra dice of damage. Changed wording of query to reflect this. Change default extra die to 0, since this is the minimum extra dice that can be rolled as extra damage. 
- Updated broken image for check boxes. Now slash image will display for both conditions, but will distinguish between checked and unchecked with red behind image. 
- Background was using a Chronicles of Darkness image. Updated to use W20 specific background.

### DIFF
--- a/CWOD-W20/W20.css
+++ b/CWOD-W20/W20.css
@@ -14,8 +14,9 @@ div.sheet-core-content {
 	display: none;
 }
 .charsheet {
-    background: url('http://i.imgur.com/Gqhoqv5.jpg');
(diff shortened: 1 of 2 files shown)


Solarswordsman:

pensman:
Excellent, I didn't think of that, thanks for the tip!

PR #2689, opened 2017-06-14
Blades in the Dark v1.4 Hotfix - please accept soon if possible

joesinghaus:
A bug crept into 1.4 which would disable the comfort feature of filling in crew data automatically. Not a severe bug, but it would be great if you could accept it before putting it live so nobody gets inconvenienced.


joesinghaus:
Thanks!

PR #2691, opened 2017-06-14
Master

MrPadRpg:
v1 of the Hawkmoon sheet.


Solarswordsman:
Your translation file is incorrectly named. It should be `translation.json` rather than `translate.json`.  If you had changes to the translation.json, please commit them to that file and remove `translate.json`.

MrPadRpg:
Ok, I made the changes.

PR #2694, opened 2017-06-16
Add files via upload

dougansf:
Earthdawn 4E attempt 2 to apply recent changes.


Solarswordsman:

dougansf:
I've gotten this feedback before, and would be happy to comply.

Unfortunately, I'm very new to this whole process, and I don't know which file you're referring to.  

The HTML and CSS were written in the Roll20 site, then copied to GitHub, pre requirements.  The other files were largely copied from other documents in the Roll20 archive, and adjusted to suit my needs.

Any more specifics you can give me would help me to resolve this.

Thank you for the help

Solarswordsman:
All four changed files have the wrong line endings.  You'll want to open all the files in a text editor that's capable of converting these line endings, save the files, then commit the changes on your repository.

For example, you could open the files in Sublime Text 3, go to View -> Line Endings and change to Unix.  You should also set that as a global configuration for Sublime Text 3 so you don't have to do this all the time; I would google the details of that.

PR #2699, opened 2017-06-18
First pass at Adventures in Middle Earth character sheet

RandallBurt:
Modified the 5e SRD sheet to support the changes needed
for Adventures in Middle Earth by Cubicle 7.


Solarswordsman:
I've gone ahead and corrected the naming of your translation file (it should simply be `translation.json`) for you, make sure future changes operate on that.  Everything else looks good.  Thanks!

PR #2702, opened 2017-06-18
Update d20 Modern.html

SilverRahl:


Solarswordsman:
Probably you didn't want to change your roll template calls to `&amp;`, but it should still work anyway, so I can merge this.  Just thought I'd let you know.

PR #2703, opened 2017-06-19
Adding Mythras character sheet

DigitalPagan:
Addition of a Mythras character sheet


b34s7m0d3:
Hey,

I try to make the sheet work but for some reason, can`t get it to look like the preview. What libs are you using? 

Thank you!

DigitalPagan:
Hi,
No special libs, just sheet workers and css.  I code mostly using the custom sheet editor in Campaign settings and play test the sheet in Chrome.  If you may, forward me a screenshot of the results your seeing and/or the browser and version you are using.
Thanks,

b34s7m0d3:
I tried to preview it on codepen, this is the result i see: https://codepen.io/b34s7_m0d3/pen/OgaOYW
For some reason, it doesn`t load the logo image either. Tried to put it on my domain, but reached the same problems. Browser is latest chrome.

DigitalPagan:
Ahh I see.  This won't work in codepin or your own domain.  Roll20 runs the sheets in a specialized sandbox.  Various security features prevent us sheet authors from hijacking their css on purpose or by accident.   You may notice classes in the css code start with "sheet-" but the html omits that.  That is because of the sandboxing.  If you manually add "sheet-" to the html or remove it from the css that may do the trick.  That said autocalculations, sheet workers and roll tmeplates will probably still not work.

The best way to test is via a pro subscription and using a custom sheet in the campaign settings of a game.  Simply copy and paste Mythras.html into html, Mythras.css into css and translation.json into Translation.

PR #2711, opened 2017-06-21
VtM 20th AE i18n formatting test

JackNexus:
I tried to create an i18n formatted character sheet to translate it in italian myself and permit other translations.
It should need a test to fix possible bugs due to my lacks of coding knowledge. 
I also created manually a .json translation file with the corrisponden keys used in this html.

### DIFF
--- a/CWOD-V20/V20test.html
+++ b/CWOD-V20/V20test.html
@@ -1,0 +1,501 @@
+<img src="http://theonyxpath.com/wp-content/uploads/2012/05/Vampire20Logo.png" style="max-height: 100px;" />
+</p>
+
+<div class="sheet-3colrow">
+    <div class="sheet-col">
+        <label data-i18n="name-u">Name:</label><input type="text" name="attr_name" />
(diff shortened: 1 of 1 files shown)


Solarswordsman:
Gave this a quick look.  It mostly looks good, with a few small potential issues.  Importantly, I'd look at your spelling of 'strength-u' in a few places, and more importantly, in the sections near the bottom (`<div class="DP2">`, `DP3`, and so on) it looks like you copied your changes from DP1, which is mostly fine, but the attribute being used is wrong (look at the name of the selection element for each block and the roll it is doing in the button element--all referencing `DP1`).

I'm going to close this PR.  Once you make these fixes, it should be fine to make them directly to the V20.html rather than a new file, so please do that and make a new Pull Request.  Thanks!

JackNexus:
Ok, perfect, thanks for the hints. Prepare for the fixed file ! ;)

2017-06-26 17:59 GMT+02:00 Jeffrey Lamb <notifications@github.com>:

PR #2715, opened 2017-06-21
Little Fears

CoalPoweredPuppet:
Adding names so stuff saves.


Pointandgeek:
Looks like there's a problem on line 29 so the fight roll is not working on roll20.

PR #2732, opened 2017-06-26
Add files via upload

GeneralPith:
[Burning Wheel.txt](https://github.com/Roll20/roll20-character-sheets/files/1100821/Burning.Wheel.txt)

I also really don't know how to code or use github, so I'm not even sure if I've done this right...

But assuming I have I made the following changes:
*Changed "Speed Multiplier" to "Stride"
*Added "Mounted Stride"
*Changed Bare-fist WS from "F" to "3"
*Corrected misspelling of "Divide" in the Character Burner Resources section
*Moved the Circles Attribute to directly above Reputations and Affiliations
*Included a "Loans" line under Resources
*Added Codex Stocks and Age Charts

### DIFF
--- a/Burning Wheel/Burning Wheel.html
+++ b/Burning Wheel/Burning Wheel.html
@@ -1,2158 +1,2241 @@
-<!-- BACKGROUND AND CONTAINER -->
-<div class="container">
-<div class="background">
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
+<HTML>
+<HEAD>
+<META HTTP-EQUIV="CONTENT-TYPE" CONTENT="text/html; charset=us-ascii">
+<TITLE></TITLE>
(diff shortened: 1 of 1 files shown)


joesinghaus:
This is not going to work. What you have uploaded here is a html page that shows the code for your character sheet, not the actual code for the character sheet.

Solarswordsman:
Indeed, the advice above is probably a good place to start.  I'd also point you to the root directory of this sheet itself in this very repository, as the code you provide will need to be a modification of that, rather than a modification of all the code you see in your browser.   I'll also point you here: [Building Character Sheets](https://wiki.roll20.net/Building_Character_Sheets).  

I'm going to close this PR for now.

GeneralPith:
I don't understand because that's not what I did. I have a Roll20 Pro account. I took the html from the root directory here and modified it in the html editor in Roll20's game settings page. Then I took that and copy pasted that in Notepad. I made a custom sheet and checked to make sure everything I changed works as intended and wasn't disruptive.

PR #2737, opened 2017-06-27
[Pathfinder] v1.58 damage by size update

plutosdad:
Fix to damage dice when changing size between small and medium
Fix to frequency fields showing on new abilities even when N/A


wing-it:
We can't view the diff, just wanted to make sure you meant to delete 14k lines from your html file?

plutosdad:
that doens't sound right,

plutosdad:
OMG hold on, ill fix it in a minute

wing-it:
Good, that did seem a little strange ;)

plutosdad:
OK i added the missing lines back. thanks. The problem was one of the source files is named "index.html" and the build file is named "index.html" And I pasted in the source file apparently. probably i should give it a different name. Maybe automate the upload too.

plutosdad:
thanks for letting me know. the users would be after me with pitchforks!

wing-it:
No problem!

PR #2739, opened 2017-06-27
Update BW html and css

GeneralPith:
*Changed "Speed Multiplier" to "Stride" (the correct term for the current edition BWG pg 93)
*Added "Mounted Stride" (relevant for those riding horses or great wolves)
*Changed Bare-fist WS from "F" to "3" (the correct term for the current edition BWG pg 466)
*Corrected misspelling of "Divide" in the Character Burner Resources section
*Moved the Circles Attribute to directly above Reputations and Affiliations (better formatting imo because Circles can be affected by Reputations and Affiliations)
*Included a "Loans" line under Resources (related to Resources and missing from the list BWG pg 375)
*Added Stocks and Age Charts from the Burning Wheel Codex


wing-it:
Looks like you've changed some attribute names, will this break anything for people already using this sheet? If so, you might consider adding a sheet worker to update the values for existing characters.

GeneralPith:
Well the only name change is the "Speed Multiplier" to "Stride" change, and it only appears once in the html and once in the css (it's also not used in any calculations in the sheet). I don't exactly see how it could break things for people, but I could be wrong. How exactly would a sheet worker improve the situation?

wing-it:
I'm specifically talking about attribute name changes. At first glance, I thought you were changing all of the `attr_custom2` attributes to `attr_circles`, but when I took a second look I realized you were just swapping the position of the two, so I see now that it is only the one attribute.

The idea is to use a sheet worker to update the old values and convert them into the new attributes, so that existing users aren't faced with empty fields where they once had stored values.

I'm not familiar with this system, will losing the data in the `attr_speed_multiplier` disrupt anyone's game, or be a significant inconvenience? If not, I will go ahead and merge this.

However, since there is just the one change, writing a sheet worker to automatically update to the new version should be pretty straightforward and would be an easy way to 'get your feet wet' with sheet workers, even if it's not completely necessary. If you are unfamiliar with sheet workers, I'd be happy to help get you started, if that's what you decide to do.

GeneralPith:
Oh okay I see what you're saying now. I have played the system, and it shouldn't be disruptive at all. It's so rarely used and so easily looked up that I'm sure a lot of people don't even fill in the field. I wouldn't be surprised if most people didn't even notice losing it.

It's probably unnecessary, but if you'd prefer going the sheet worker route, we can. I just definitely have no idea how to do that lol.

PR #2742, opened 2017-06-28
Image/Merit&Flaw/Button

pensman:
- Update images to draw from github
- Removes transparent background from buttons. Transparency was distracting and not as intuitively clear that the control was a button.
- Add more fields for Merits and Flaws for easier reading and detail. Select for type, text for name and number for point value. Make Merits and Flaws fields dynamic, to add as many or as few as needed.

### DIFF
--- a/CWOD-W20/W20.css
+++ b/CWOD-W20/W20.css
@@ -14,7 +14,7 @@ div.sheet-core-content {
 	display: none;
 }
 .charsheet {
-    background: url('http://i.imgur.com/CWLfKIZ.jpg');
+    background: url('https://raw.githubusercontent.com/Roll20/roll20-character-sheets/master/CWOD-W20/character_sheet_background.jpg');
     background-size:100%;
     background-repeat: repeat-y;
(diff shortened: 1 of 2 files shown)


wing-it:
So, it looks like you're removing a couple of attributes from this sheet (merits and flaws). Do you think you could come up with a way to preserve the old data so that it's not lost for legacy users?

pensman:
Sure, I can do that! I've got an idea in mind that I think will work. I'll
play with it and resubmit when I've got it working.

 Does updating the repository update anyone using the character sheet? I
was assuming that only new games would get the changes, which is why I
wasn't worrying about the old attributes initially.

wing-it:
Yes, when character sheets are updated it will affect existing games. Thanks!

pensman:
I've made an update to preserve the referenced attributes as well as give the user the option to switch between the current format for merits/flaws or the new way that I'm proposing. It will currently default to the old way, so current users have minimal interruption to their expected workflow when using this sheet.

PR #2743, opened 2017-06-28
Fixing V20 translation.json file after testing

JackNexus:
I noticed some errors in both V20.html after live testing, im proceeding to fix them. It's going to be necessary to recreate the translation files (en.json, de.json, etc) except for it.json which i'm fixing myself in another pull thread.


heathen-pwn:
Was this ever fixed? Facing the same problem in 2020.

PR #2748, opened 2017-06-29
MV Patch

Jhar226:
Fixes some issues, added "Skill Focus" options as well as "Miscellaneous Bonuses" options for all skills. Added a base value of 0 to all character stats (and base of 6 for move speed).


wing-it:
Looks like you're changing some attribute names that may cause users of the old version of the sheet to lose data. You might consider adding a sheet worker to update these so that users don't lose anything.

Jhar226:
Which attribute names am I changing?

wing-it:
For example, where you are changing Artisans to be a repeating section, that will not preserve old data because even thought he attribute names are the same, they are now in a repeating section so they won't automatically update.

Jhar226:
Artisan was changed to a repeating section last patch to begin with, and the code for Artisan prior to the last update was flawed to begin with. In order to change it to a repeating section, the code for the old Artisan wouldn't have been able to be carried over.

I'm sorry if this inconvenienced you or anyone else.

wing-it:
Ok, that sounds fine, but could you please resolve the conflicts on this branch so that I can merge this request?

Jhar226:
Which conflicts need to be resolved, aside from the Artisan field changes as mentioned above?

wing-it:
Github is detecting a merge conflict in `Multiverse/Multiverse Sheet.html`. You should see the message **This branch has conflicts that must be resolved** on this page below our comments.

You can resolve the conflicts using the web editor by clicking on **Resolve conflicts**, or it may be simpler to discard your changes, bring your fork completely up to date, and then re-apply your changes.

I'm happy to help walk you through any of this if this isn't clear.

Jhar226:
I've gone through and I'm pretty sure I've fixed all of the issues present.

PR #2751, opened 2017-06-30
Updated due to core rules released

suntzu777:


wing-it:
Looks like you are removing/changing quite a few attributes, how will this affect current users of the sheet?
You might consider adding a sheet worker to handle migrating existing characters to the new version.

suntzu777:
the old sheet was based on the beta release , the attributes have changed so i changed them to suit , i dont know how to use the sheetworker stuff so up to you if you want to merge this at all

wing-it:
I'm not familiar with this system, how drastic will the changes be for existing users? Will they lose a lot of data or will the changes be relatively minor?

suntzu777:
one attribute was changed from siegecraft to warfare , and have removed a couple of sections which are no longer relevant but those items would still appear in the old labels in the attributes and abilities section  those items will not display on the new sheet but would still be there if needed

wing-it:
Ok, that sounds reasonable to me.

PR #2756, opened 2017-07-01
Sword & Scoundrel

GrandHeresy:
# Sword & Scoundrel

This is the feature complete Sword & Scoundrel character sheet. 

I will be updating as often as I can with feedback from the community.

Thanks, Barbarossa


GrandHeresy:
Missing information

PR #2761, opened 2017-07-03
[OL:M] Open Legend (Minimalist) First Public Draft

DejaAugustine:
Initial public draft of the **Open Legend (Minimalist)** character sheet


BuildAndDie:
Nice Design! Played around with it a bit.

I did notice that when I had physical dice checked, and click to roll the action, it pops up an error and ends up asking twice for dice. Once for the attribute dice, and then again for just a d20. It shows the second input (asking for d20 result) to chat, but seems to error on the first one.

DejaAugustine:
@BuildAndDie Thanks!  And great catch.  I fixed that issue while performing a larger refactor (you may have to delete and re-create any test characters for the fix to show).

DejaAugustine:
@wing-it any feedback on why this was reverted?

wing-it:
Yes, sorry! I realized that I needed to review this further last thing on Monday, and haven't had time to get back to it until now because of the holiday.

As far as I can tell, this is just a different layout of the existing Open Legend sheet. We typically don't allow duplicate sheets, as referenced in our guidelines:

>**Don't Duplicate Work**
>We only need one sheet for each system. For some major systems like Pathfinder or D&D, we may allow more than one sheet if they are different enough, but contact us in advance before you spend a lot of time working on a new sheet for a game we already have a sheet for. We should focus on improving existing sheets, not creating totally new ones. Along those lines, this is a community effort, so if you want to help improve an existing sheet, just jump in and do so.

You can read the rest of the guidelines [here](https://github.com/Roll20/roll20-character-sheets/blob/master/README.md).

I apologize for the inconvenience and for not getting back to you sooner!

DejaAugustine:
@wing-it Ah yes, I expected that to come up. The reason I felt this warranted a new sheet was three-fold:

1. It is an explicitly minimalist approach to the character sheet, intended to provide a different experience from the existing sheet, and closely mirror the paper sheet experience. There are explicitly no hidden options or deep configuration available, so what you see is what you get. This makes the minimalist sheet fundamentally incompatible with the existing sheet.

2. There are some players that would certainly want the deeper configuration and automation available in the current sheet, so it did not seem appropriate for me to suggest replacing the existing sheet wholesale.

3. Some features, such as the physical dice prompts, would both require a very substantial reworking of the sheet and would be a substantial change in behavior for current users of that sheet.

I felt it would be worth offering the rapidly growing community a choice between a near-paper experience and one that leverages the digital medium more thoroughly, especially considering the number of systems that Roll20 supports that also feature multiple character sheet options.

wing-it:
Sorry for taking so long going back and forth on this, I just wanted to get a second opinion.

This seems like sufficient reason for a new sheet. If you open a new pull request, I'll go ahead and merge this in.

DejaAugustine:
I appreciate the due diligence, and I'm glad you've decided to accept the sheet 😄 

I've rebased to trigger a new commit SHA (for a clean merge around the original revert) and created a new PR #2787

Thanks!

PR #2762, opened 2017-07-03
Fixed some fields not saving and made significant milestones possible

Erjulk:
HTML: (numbers refer to lines of old code)
21 - fixed description long name (added "attr_")
37 to 46 - replaced 3 static Aspects with a fieldset for more variability
55 to 148 - replaced static values and dropdown for methods with static methods and number input for values (needed for significant milestones)
55 to 148 - rewrote the roll macro to work with new input fields and changed the default for "roll with mod" to +2 instead of +0 which already had a button.
174, 183 and 192 - fixed names ("attr_" again), gave them unique names and made the size consistent

sheet.json:
added myself as author and included my roll20 id

fate.png:
replaced with preview


wing-it:
Hi Erjulk, these all appear to be positive changes. However, it does look like you'll be removing some attributes. How will this effect existing users of the sheets?
You may consider adding a sheet worker to update the sheet, so that existing users don't lose any data.

Erjulk:
Hi!
Thanks for the fast reply.

i will add a sheet worker - can i use the getAttrs function to access the
values of the removed attributes?
the wiki wasn't clear on this.

thanks in advance

2017-07-03 21:41 GMT+02:00 wing-it <notifications@github.com>:

joesinghaus:
For values which have been changed at some point, hence saved to the character sheet, you can get them via getAttrs(). If you get nothing from getAttrs(), they have not been changed from their default in the old sheet.

PR #2764, opened 2017-07-05
Fix issue with resource tracking and clean up HTML

RandallBurt:


RandallBurt:
Any impact will be minor at best. The only possible issue would be that a
user would have to re-add a resource and even then, their sheet would
already be broken because the exhaustion resource would be missing.

PR #2773, opened 2017-07-08
Add Rokea sheet for Werewolf

alanman:


alanman:
I think I messed up the pull request process, but the data is all good.  I meant to make 2 different pull requests, but I think I somehow merged them?

Part 1 is adding a sheet for Werewolf: Rokea
Part 2 is cleaning up all the CWOD sheets so that the Mental Attributes are in the proper order: Perception, Intelligence, Wits.

PR #2779, opened 2017-07-10
Update Mass Effect.html

SilverRahl:
Fixing weapon roll table.


wing-it:
It looks like you're removing or changing some attributes. I'm not familiar with this system, how will that effect current users?
You might consider adding a sheet worker to update legacy users, so that no one loses any data.

SilverRahl:
Some attributes for skills were wrong so I fixed it. It will not affect badly for people who are playing right now as they still can change those attributes in game by manually. I made those changes to set as default. Beside that weapon roll system was broken. I figured it out and fixed it. Now everyone can roll weapon dice without error I suppose. We will see that after patch is uploaded.

PR #2784, opened 2017-07-10
DCC Tabbed Sheet: This is a new DCC sheet with Tabs, Monster and 0-level character sheet

amandasearle:
It has been in development and playtesting since 2015 as an alternative to the original DCC sheet. Most everything has roll templates. Includes classes from Crawl! and Crawling Under a Broken Moon. There are plans to add in additional classes along with a custom class.

Referenced on the forums by Phil here - https://app.roll20.net/forum/permalink/4841952/


wing-it:
Did you mean to include the css file from the MCC Tabbed Sheet?
The rest of this seems fine, I just wanted to check before I merge because it seems like that might not have been intentional.

amandasearle:
I did NOT mean to include that file and I can't seem to find how it got included in the pull request. :/

amandasearle:
Removed the MCC css file - this pull request should be good to go

wing-it:
So it looks like github read your last commit as moving the MCC file to the DCC folder, which removed it from the MCC folder completely. We've added a previous version of that file back in, so it's working now.

However I did want to let you know that we rolled that file back 2 commits, because the most recent file there was broken (it looks like it had extra lines from git diff that were causing it to break). So, you may want to go back and check the MCC css sheet and check to see if everything is how you want it to be.

Sorry about this, I should have noticed this on your pull request in the first place.

PR #2787, opened 2017-07-12
[OL:M] First Public Release

DejaAugustine:
First public release of the **Open Legend: Minimalist** character sheet.

Original discussion thread: #2761


wing-it:
You may have noticed that your sheet does not have the Default Sheet Settings section. You didn't do anything wrong, there was a bug on our end that was causing the sheet description to load incorrectly on certain sheets when the Sheet Settings section was present, so we removed that section so we could go ahead and get your sheet out there.
There is a fix in the works that should become active next week, as soon as it is fixed we'll add that section back to your sheet.json file.
Sorry for the inconvenience, but thanks for helping us find this bug!

DejaAugustine:
@wing-it It was on my list to ask about in the forums, but like you I was far more interested in seeing the sheet go live. Glad to have helped root out a bug, though!  Thanks for letting me know :)

PR #2790, opened 2017-07-13
Add missing .json

alanman:
Sorry, forgot to add one of these.  Also, I don't have access to the preview feature, so I cannot make one for character select.  Does the preview tool generate one?


alanman:
Added the Rokea preview to the commit too.

PR #2792, opened 2017-07-13
v1.2.0 of Mythras Sheet

DigitalPagan:
Should the edit to approved.yaml but part of our pull requests?  Seems to be taking much longer to get the Mythras sheet into production then others I made in in the past.


wing-it:
No, you've done everything correctly, it was our mistake that it didn't get added. Will correct it this week, sorry about that!

PR #2794, opened 2017-07-14
Blades in the Dark v1.8

joesinghaus:
Question: do I need to update all translations like this for the future, or only the translation.json?


wing-it:
You just need to make changes to translation.json, the other language files will be updated automatically. Thanks!

PR #2800, opened 2017-07-17
Fixed bug associated with Job Ability Range field.

Meduna:
A bug was causing this field to always change back to 0 no matter what you inputted. Thanks to Diego A for reporting this!


wing-it:

PR #2810, opened 2017-07-19
Upgrade the old Black Crusade sheet

teen120:
I made a new Black Crusade character sheet for one of my players. It take elements from the old Black Crusade sheet and Dark Heresy Advanced 2nd Edition, since both systems is pretty similar.


wing-it:
Although these changes do seem to be generally positive, they are pretty extensive. I'm not familiar with this system myself, could you tell me how these changes might affect users that are currently using this sheet? My main concern would be concerning attributes that have been removed or had their names changed, as this data would be lost in existing games.

teen120:
No need to worry, I have taken that into account.
I kept the old attributes names, only add new one, nothing will be lost.

PR #2819, opened 2017-07-21
Mutants and Masterminds 2nd Edition Character Sheet

UltraGuyver:
I would like to submit a character sheet for Mutants and Masterminds 2nd Edition


wing-it:
There are two pull requests for this sheet, it is unclear which one is the correct version. Please delete the incorrect version so that there is only one pull request for this sheet.

PR #2822, opened 2017-07-22
Update C&C.html

romelwell:
Modified Attributes by adding check box for Secondary and Ternary attributes.  Moved text of Attribute Category to the title attribute for each Attribute label.

### DIFF
--- a/Castles&Crusades/C&C.html
+++ b/Castles&Crusades/C&C.html
@@ -98,58 +98,75 @@
          </tr>
          <tr>
             <td class="bluehilite alleft attrchk">Check</td>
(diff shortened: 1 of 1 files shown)


wing-it:
It looks like you are changing some attribute names, which will cause existing games with this sheet to lose data. I'm not familiar with this system, how will these changes affect existing users?
It may be a good idea to add a sheet worker to update the existing users to the new version of the sheet, so that their data isn't lost.
Also, as a side note, it would be helpful if you would include the changes to all the files in this sheet in one pull request. Just commit the changes for all files on on the same branch and they will all be included in your pull request. Thanks!

PR #2825, opened 2017-07-24
Bugfix Sante/Orfevrerie

Zakarik:


wing-it:
It looks like you're changing some of the attribute names here, which will affect games currently using this sheet. I'm not familiar with this system, how will these changes affect current users?
You might consider adding a sheet worker to update legacy users to the current version of the sheet.

Zakarik:
I've add a little sheetworker for update attribute with the new name.

I've rework a little how the character sheet work for that too.

I hope it's good now.

PR #2832, opened 2017-07-24
Update Call_of_Cthulhu.html spanish

Gussancho:
1. The sheet has broken image links, hopefully solved with the spanish version of it.
2. Also had problems with the dice buttons, retranslated the english version with those problems solved.
3. With the book in hand, added skills and changed others
4. Arranged the skills alphabetically while waiting for list ordering and adding i18n to the original one

### DIFF
--- a/Call_of_Cthulhu_Spanish/Call_of_Cthulhu.html
+++ b/Call_of_Cthulhu_Spanish/Call_of_Cthulhu.html
@@ -1,7 +1,7 @@
 <p>
     <div class='sheet-2colrow'>
         <div class='sheet-col'>
-            <img src="http://kingstonnexus.ca/wp-content/uploads/2014/03/call-of-cthulhu-logo_3.png" style="width: 70%;" />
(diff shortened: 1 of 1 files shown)


aquaalex:
we should really start looking at using one standard character sheet with the i18n translation options for various languages instead of having separate sheets for languages.

Gussancho:
I'm going to end helping in that, but right now, I feel better working in a sheet where if I do something wrong, it will inconvenience a really little group of people while I solve it.

1. First I will solve the biggest problems the sheet has. Formatting, dices not working...
2. Second, I will work in separating the css. I thing it's also needed for the i18n.
3. Third, apply the i18n and pass all the work to the english one. I will wait till list ordering is open to all the sheets.

Also, in the Call of Cthulhu case, it has regional sheets. I have the original books in spanish and when I see the french sheet here is like I'm looking to another game with new rules.
[Here](https://app.roll20.net/forum/post/4727242/issue-with-sheet-translations) David has seen the same problem.

PR #2836, opened 2017-07-26
Resubmitting Scum and Villainy character sheets with line endings removed.

cmcdonald22:


joesinghaus:
Great that this sheet made it onto Roll20! Looks like your pull request has some kind of conflict? Also, your html ends up in the Scum And Villainy folder, not the Scum - Villainy folder.

cmcdonald22:
@joesinghaus Yeah, I'm still trying to figure out how github submmissions and everything work, and I'm not a professional developer by any means so to be honest I don't even understand what the line endings thing means (I just used a whitespace compressor and hope that fixes it?)  And I also didnt know if Roll20 uses the folder name as the template for the display name for Roll20 and I could't get the folder name to display the & properly the first time, so I was trying to submit a folder name change in the process and apparently just kinda messed it up.

PR #2838, opened 2017-07-26
Blades in the Dark v1.9

joesinghaus:


wing-it:
While removing the content from inside translated tags does work, best practice is to leave it there for the sake of readability. Was there a reason for removing it?

joesinghaus:
There's no pressing reason to remove it, and it doesn't improve readability
all that much (for me personally) to have lots of non-functional code. I
can revert the changes if you really want the content there, otherwise I'd
prefer them fine.

joesinghaus:
Anyway, I've added it back now, was only a quick regexp replace. It doesn't matter all that much.

wing-it:
I wasn't going to try to force you to change anything, I just wanted to let you know what we consider best practice. Thanks!

PR #2843, opened 2017-07-27
Delta Green Character sheet.  CSS based on Call of Cthulhu 7ed CSS.

dhatcher:
Simple but it works.  Anyone can feel free to make it look nicer.  For some reason the roll20 site doesn't respect the css I used to make the section headers with vertical transformations but otherwise it's serviceable.

### DIFF
--- a/Delta Green/delta_green.css
+++ b/Delta Green/delta_green.css
@@ -24,8 +24,7 @@ div.dg-header {
 
 img.dg-img {
     display: block;
-    margin-left: auto;
-    margin-right: auto;
+    margin: 0 auto;
     height: 40px;
 }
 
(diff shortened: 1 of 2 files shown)


wing-it:
Just a heads up, your request didn't include a sheet.json or a preview image, which are required for inclusion in the drop down menu. I went ahead and added these for you, however, I don't have your Roll20 user id, and I wasn't sure how you would prefer your name to be credited. If you want, you can go back and edit the sheet.json file and submit another pull request so that you are properly credited for this sheet. Thanks!

dhatcher:
Thanks for doing that.  The way you put me in is fine.

PR #2860, opened 2017-07-31
Update2 Zweihander.html

robertmyers:
Renamed duplicate text entry on line 595 from attr_trait-4 to attr_trait-whoops2.  The duplicate text line is at line 585


wing-it:
This appears to be a duplicate of #2869.

PR #2863, opened 2017-07-31
Add starship traits per fillable sheet from Modipheus

### LINKED ISSUE #2835: Star Trek Adventures 2d20 - roll macro, applicable focus + 1
When the applicable focus is selected yes on the roll macro it highlights all rolls that are focus# +1. It should only highlight numbers that are equal to or less than the focus.

RandallBurt:
Closes #2835


RandallBurt:
@GrumpyOldDM fyi. please review. Thanks!

GrumpyOldDM:
@RandallBurt we need to remove the 1+ from applicable focus roll macros - do you want to make that change real quick or do you want me to do a different push with that change so we can have it one single commit? - so all the 

`1+?{Applicable Focus`
should read
`?{Applicable Focus`

This is only on the character roll macros not ship.

RandallBurt:
@GrumpyOldDM yeah, I can make that change real quick.

PR #2876, opened 2017-08-04
Exalted 3e Sheet

pinmissile:
Updated the sheet to 2.1. See commits for more details.

Kindly merge this the same time as you merge the pull request I made with the API repository, as the script I updated had to be modified to remove hard-coded restrictions enforced by this sheet.


wing-it:
It looks like your branch was out of date when you submitted this pull request, which has resulted in a lot of conflicts which are preventing me from merging.
These conflicts will need to be resolved, the simplest way to do this may be to delete your branch and then create a new fork. This will make sure everything is completely up to date. Then, re-apply your changes and submit a new pull request. If this doesn't make sense, let me know and I'd be happy to help.
Also, how will these changes affect current users of the sheet? Mostly we're concerned about changing or removing attribute names. Normally I would glance through the commit to check for this but your changes are difficult to look through quickly.

pinmissile:
The changes will not affect  current players. Previously, the attribute 'peripheral-essence_max' and 'personal-essece_max' were hard coded to '@{essence} * 7 + 26' and '@{essence} * 3 + 10' respectively. This update changes these attributes to @{personal-equation} and @{peripheral-equation}, which are editable and by default set to the previously hard coded values.

There is also a worker script that changes the equation attributes, however it is only set to run when the attribute 'caste' is changed.

I think I get what you mean. I'll get around to submitting a new pull request with no conflicts. Thanks for the help!

PR #2877, opened 2017-08-05
Updates to SWRRPG character sheet

traec:
Added additional force powers and feats from "Power of the Jedi" Sourcebook. 

Added input values, repeating sections, and text areas for starships and vehicles


traec:
SWRRPG = Star Wars Revised RPG

PR #2882, opened 2017-08-06
AD&D 2E ver 2.8.51 and Coriolis fix

Project462:


wing-it:
Would you mind separating this into two pull requests, one for the sheet update and one for the new sheet?

Project462:
I tried to, but I wasn't sure how. Should I just push them some-now,
some-later?

Project462:
Ok, I removed the new sheet. I guess I'll wait for the other stuff to go through and then try to upload the new sheet. If it has to wait for next week, well, that kind of blows, but I get it :)

wing-it:
The way to do this is to have the different sheets on different branches. So, create a new branch for the new sheet and the pull requests for it will be separate.

PR #2891, opened 2017-08-08
[Pathfinder] v1.67 small size fix

plutosdad:
very small fix, added a function back to a module that is only used in one place, but has an impact on GMs but not players. 

@wing-it  can we put this in before next week?
it is just the html /js file.


wing-it:
Unfortunately, we've already performed all of our planned updates to the website this week. If we end up pushing anything else, I will see if I can get this included, otherwise, it will have to wait until next week.
Thanks!

PR #2893, opened 2017-08-09
Fixed error and small additions

traec:
-I messed up and committed window HTML rather than Unix so the line endings were wrong on my last commit. Adjusted to unix line endings .
-Resized a few inputs to show full numbers
-Added a few line breaks to even things up
-Added drop down for Vitality Dice
-Changed a few inputs from number to text as it works better to calculate .01-1 values.


traec:
For the Star Wars Revised RPG character sheet.

traec:
Also, noticed that html files for the older version and newer version are both still in the same folder on github. Not sure if that will cause any issues.

wing-it:
I apologize, I didn't notice the second file at first.
When submitting changes to a sheet we ask that you commit your changes directly to the original html file. I'm going to close this pull request, if you could delete your extra files and submit a new request with the changes applied directly to the original file, we'd appreciate it!

PR #2895, opened 2017-08-09
Cowtankerous patch 3

cowtankerous:
okay hopefully these changes will work, fixed the logo not working (was removed from the original linked site) undid an attempt to make a repeating entry within a repeating entry (that didn't work out so well but was nice attempt (it kept putting them in all the other repeated entries so that was no good.


cowtankerous:
aberrantlogo

cowtankerous:
how do i upload that image as a file so it can be linked internally?

PR #2908, opened 2017-08-13
Add files via upload

cowtankerous:
original site took down the linked image used on the character sheet. will link to wiki currently, but next update i can change it to link to this instead


wing-it:
No need to wait till next update, I've gone ahead and updated the link for you!

PR #2912, opened 2017-08-14
ACKS Sheet Encumberence

DimensionalCode:
More Encumbrance Adjustments.
GP for Treasure now calculates stone weights
Item Counts operate as multipliers.


DimensionalCode:
I've just been corrected on the gp weight mechanic (there's some conflicting rules on that) so I'll be removing it and resubmitting later when I get back to my computer.

PR #2919, opened 2017-08-15
Delta Green RPG

uriele:
New Character sheet for DG based on the rules provided in Need to Know and the Agent's Handbook by ArcDream Publishing.
Integrated roll template for the SAN check and all the attacks, skill and stat tests.
Support of the projection on bonds using the blue textbox next to the bond score.
Auto computation of the new Breaking Point with update every time the power of the character changes
Support of the automatic roll for special action like healing, resucitating, falling, catching fire, pinning, and disarm not supported in this version of the sheet.

These special rolls might be obtained rolling the relative skill and interpreting the output result.

Special thanks to AquaAlex in helping me creating the initial version


wing-it:
Would you consider submitting this as an update to the existing Delta Green sheet? It would require a little extra work to make a sheet worker to update the old sheet attributes to the new ones, but I think the users of that sheet would really appreciate this upgrade.

uriele:
Sure! How should I do it exactly?

wing-it:
You would start by closing this pull request, and then apply these changes directly to the existing Delta Green files.
It's likely that you can handle a lot of the update without sheet workers, simply by changing your attribute names to match the old ones. For the rest, you can access the old attribute values with a sheet worker and update them to something the new sheet can use. This shouldn't be difficult, but may be a bit tedious.
After you've done that you can just submit those changes as a new pull request.
If you have any further questions, please let me know, I'm more than happy to help!

uriele:
Mostly how to contact the other author to update him about the possible changes, if possible and how to change the sheet.json author file to the new sheet, changing names it something I can do right now :)

uriele:
Also should I change some of the fields in the other sheet? For example:
attr_v1 is attr_violence_1 and attr_h1 is attr_helpness_1 in my sheet
also I used attr_gender instead of attr_sex trying to follow the instructions in the wiki

The rest of the attributes seems to be almost the same (they only used a different format putting chk before the skill instead of after and I'm changing it)

PR #2925, opened 2017-08-16
Renamed Tabs

dougansf:
Reformatted Readme, Carrying capacity, minor cosmetic fixes


wing-it:

dougansf:
Sorry, I did try to fix it because I've gotten this feedback before.

wing-it:
Which text editor are you using? I may be able to help you get the settings correct.

dougansf:
I've been writing in the Roll20 custom character sheet editor.

I tried to use Sublime to correct it, but apparently it's not working.

PR #2926, opened 2017-08-16
New, styled character sheet for Ghost Lines

timdenee:
Completely new sheet for Ghost Lines, with a much more graphic approach
that better mirrors the print character sheets. Due to the completely different approach, the old Ghost Lines sheet would have to be deprecated (some key attributes have been handled differently).  

EDIT: Thanks to some sheetworker magic from Jakob, this new sheet now cleanly converts from the old sheet, so it can replace it entirely.


joesinghaus:
Awesome!
How extensive are the attribute changes? If it's not too complicated, I could write a quick migration sheet worker to just transfer over stuff.

timdenee:
Looking at it again, it actually transfers over very smoothly! The only things that would need to change:
- attr_Badge on the old sheet becomes attr_name on the new sheet
- attr_Gear on the old sheet goes into attr_notes
- attr_Rumors, attr_Notes,and attr_Log on the old sheet could all get compiled into one attribute, attr_notes2 . These are all free-text areas which could be combined into one free text area on the new sheet.

joesinghaus:
Okay, that's easy!

`<script type="text/worker">
on('sheet:opened', () => {
	const attrs = ['Badge', 'Gear', 'Rumors', 'Notes', 'Log', 'version'];
	getAttrs(attrs, v => {
		if (v.version) return;
		let setting = {};
		if (v.Badge) {
			setting.name = v.Badge;
			setting.Badge = '';
		}
		if (v.Gear) {
			setting.notes = v.Gear;
			setting.Gear = '';
		}
		let notes2 = '';
		if (v.Rumors) {
			notes2 += v.Rumors + '\n';
			setting.Rumors = '';
		}
		if (v.Notes) {
			notes2 += v.Notes + '\n';
			if (!v.Gear) setting.notes = '';
		}
		if (v.Log) {
			notes2 += v.Log;
			setting.Log = '';
		}
		if (notes2) {
			setting.notes2 = notes2;
		}
		setting.version = 1;
		setAttrs(setting);
	});
});
</script>`

timdenee:
You're a legend, thanks! I've added that in now, so this new sheet can cleanly replace the old one.

wing-it:
Happy to see authors working together! Since this can now replace the old sheet, could you submit this as a direct update to those files?
Also, this includes some files for the Sprawl sheet, I'm not sure if this was intentional or not, but if it is, we generally prefer each sheet have a seperate pull request. Thanks!

PR #2928, opened 2017-08-16
Updated new attribute selection for attacks and spells.

shukinhekon:
Many bugfixes that appeared after playtest involving negative attribute modifiers, with the addition of choosing which attribute to roll when using different Attacks or Spells.


wing-it:
Github is detecting some conflicts in these files, can you resolve these so this can be merged? Thanks!

shukinhekon:
I added more information between the first pull request and the merge, that's what caused the conflict. I've updated the requested pull with the newest version of the code.

PR #2936, opened 2017-08-19
Blades in the Dark v1.12

joesinghaus:
Can you regenerate all the translations? I've noticed that changed (as opposed to added or deleted) translation strings are not reflected in the individual translation files. There aren't actually any translations yet, so it's safe to just scrap them all.


wing-it:
Translations are replaced with a new file every time we update. If you could give me a specific example of what you're talking about, I can look into it.

joesinghaus:
If you look at my translation.json, I changed "light": "light" to "light": "Light", but that's not actually changed in any of the files in the translations folder. There's some more cases like that.

PR #2946, opened 2017-08-21
Delta Green sheet update worker+layout

uriele:
Modified attr_sex into attr_gender
Modified attr_v1 and attr_h1, into attr_violence_1 and attr_helpness_1

Modified attr_name into attr_character_name and attr_player_name

Add modifiers to role (difficulty/general/combat/status)

Add auto-computation of bond projection, breaking point computation, attr X5.
Add attack and skill rolls for all skill/attack/stats
Add attack modifier based on strength for melee and unarmed attack
Add WP reduction when using stimulants
Autocheck death/unconsciousnell/low wp status when HP/WP goes below a treshold
Modified layout and add tabs


uriele:
@wing-it : mostly all the name were standard from the sheet, so I had only to change the position of chk in the skills and modifying the weapons attr_names and bonds attr_names in my sheet to match, all the other attributes name I modified are in the  description up there. Let me know if everything is in order. Marco :)

wing-it:
Looks like all you have left to do is to make a sheet worker to update the old sheet to the new sheet.
For example, a player currently using the old sheet that updates to this one will not see their character name, because your sheet is looking for "attr_character_name" and their data is saved in the database under "attr_name".
Here's a quick sample of what that would look like(just for the name), adapted from another sheet:
```
on("sheet:opened", function() {
  var updates = {};
  getAttrs(["version", "name"], function(attrs) {
    if(attrs.version != "1.1") {
      if(attrs.name) {
        updates.character_name = attrs.name;
      }
      updates.version = "1.1";
      setAttrs(updates);
    }
  });
});
```
This may not be the best script but should give you an idea of what I'm talking about. You'll only have to do this for the attributes whose names you've changed, if the attribute name is the same the sheet will find the old value just fine. If you've only changed a few of the names like you said, this should be a pretty simple function.

You can easily check if your update works correctly by setting a game's character sheet to the current Delta Green sheet, then making a new character and filling it out at least partially. Then, set the sheet to 'Custom' and paste in your sheet, and when you refresh the game that character should have all of the values still in place.

It wouldn't hurt to let the author of the current sheet know what you're doing, his github username is @dhatcher.

Also, please remove the 'DeltaGreenRPG' folder and the files in it from this request, as they are no longer needed.

Let me know if all of this makes sense!

uriele:
Thanks, I was looking for that user name on roll20 :)

@dhatcher: Danny, I'm submitting an update to your character sheet for delta green, let me know if it's ok or you have any suggestion/change. Marco

dhatcher:
Hey I'm excited to have an updated version of it!  I did a quick one just to have something for my players to use over a weekend.

uriele:
Great! thanks Danny. I was overthinking it and I made overly more complicated that it's supposed to be, but I liked to add all the sheetworked. I'm trying to solve the last problem of compatibility with your special fields (in my case it's a repeating field and I couldn't copy the data easily), beside that it seems your old data are updated smoothly to the new version, so you can just change sheet to this one without loosing anything :)

PR #2947, opened 2017-08-21
Delta Green sheet update v1.2 worker+layout

uriele:
Modified attr_sex into attr_gender
Modified attr_v1 and attr_h1, into attr_violence_1 and attr_helpness_1

Modified attr_name into attr_character_name and attr_player_name

Add modifiers to role (difficulty/general/combat/status)

Add auto-computation of bond projection, breaking point computation, attr X5.
Add attack and skill rolls for all skill/attack/stats
Add attack modifier based on strength for melee and unarmed attack
Add WP reduction when using stimulants
Autocheck death/unconsciousnell/low wp status when HP/WP goes below a treshold
Modified layout and add tabs

Following wing-it indication I updated almost all the fields to the new sheet. There are two things that cannot be resolved:

1) Special training is now a repeating section instead of having six fixed field. I couldn't copy the old properties into the recurring section (I tried to use $1 to $6 in the attribute name, but it doesn't work)

2) Weapon skill is not a flat number anymore, but it's a drop down menu that link the weapon skill with one of the skills defined in the skill part according to the attack skill used in the Delta Green manual (i.e. Instead of copying the attack value for firearms or unarmed combat every time, I select from the drop down menu the right skill for the weapon and the roll will use that value to roll for damages)


uriele:
@wing-it: I couldn't copy the values for special1 to special6, these value are used to remember the player about what skill they should roll for special skills like parachuting or picklocking. In the original sheet the field was given as a fixed six row table, now it's a repeating value table (I tried the code you suggested me and I used the $n to refer to the new value, but it doesn't work).
The other difference is in the weapon skills: in the old sheet you should put the skill value every time, now you choose between the attack skills in your character sheet with a drop down menu

uriele:
@dhatcher : hi danny, I tried to add new functionality to your sheet. Let me know what you thing. Marco

joesinghaus:
Hey @uriele, try this function for your sheet workers: [Pastebin](https://pastebin.com/GfNppMhM), adapted from one I've used in my sheet.

It's meant to automate filling a repeating section of a sheet with several entries. Its first argument is the section name (e.g. "repeating_training"), the second argument is an array, and each element of the array is an object containing all the data for the row, for example:

`let data = [`
`  {newtraining: values.special1, skilltraining: values.special_skill1},`
`  {newtraining: values.special2, skilltraining: values.special_skill2},`
` ...`
`];`

You can of course you the values you get from special1,... special6 to define your data! Then

`fillRepeatingSectionFromData('repeating_training', data);`

should do what you want (I may have gotten some details wrong).

uriele:
@joesinghaus : wow! It's a beautiful and elegant solution but... it still doesnt work :(

````
const fillRepeatingSectionFromData = (sectionName, dataList) => {
    const createdIDs = [];
    const setting = dataList.map(o => {
            let rowID;
            while (!rowID) {
                let newID = generateRowID();
                if (!createdIDs.includes(newID)) {
                    rowID = newID;
                    createdIDs.push(rowID);
                }
            }
            return Object.keys(o).reduce((m, key) => {
                m[`${sectionName}_${rowID}_${key}`] = o[key];
                return m;
            }, {});
        })
        .reduce((m, o) => Object.assign(m, o), {});
    setAttrs(setting);
};                                
                                

on("sheet:opened", function() {
  var updates = {};
  getAttrs(["version", "name","sex","v1","v2","v3","h1","h2","h3","special1","special2","special3","special4","special5","special6","special1_skill","special_skill2","special_skill3","special_skill4","special_skill5","special_skill6"], function(attrs) {
...  
let data = [{newtraining: attrs.special1, skilltraining: attrs.special_skill1},
      {newtraining: attrs.special2, skilltraining: attrs.special_skill2},
      {newtraining: attrs.special3, skilltraining: attrs.special_skill3},
      {newtraining: attrs.special4, skilltraining: attrs.special_skill4},
      {newtraining: attrs.special5, skilltraining: attrs.special_skill5},
      {newtraining: attrs.special6, skilltraining: attrs.special_skill6}];
      fillRepeatingSectionFromData('repeating_training', data);
      
    }
  });
});`
```
The rest of the sheet seems to work, but It doesn't generate a single new row. I've created several test characters with the original sheet, populated the special1-6 fields and switched to the custom one, nothing. :/

The annoying thing is that the field per se is not used in any worksheet scripts, and it doesn't come in play often, but it's still ruining the compatibility between the two sheets :(

@dhatcher : do you use the special training in your sheet or you think we might upgrade to this new repeating field?

@wing-it : I tried @joesinghaus script... but the function doesn't create new rows. It's the last incompatibility between the two sheets to solve

joesinghaus:
I'll take another quick look if I've messed up anything while transcribing
the function,  in 30 minutes or so.

uriele:
Dude, you are my savior. Thanks so much :)
I'm not used to javascript but the function seems correct, and the fact that the rest work it's a good indication (usually when I make typos or it doesn't compile correctly, the rest of the functions stop working)

joesinghaus:
Alright, got it. It was my mistake, mostly (but note that you wrote special1_skill instead of special_skill1 in your getAttrs argument). The problem was not taking care of default values in case any attribute on the old sheet had not been filled in. The following should work better:

`let data = [{newtraining: attrs.special1 ||'', skilltraining: attrs.special_skill1 ||''},
      {newtraining: attrs.special2 ||'', skilltraining: attrs.special_skill2 ||''},
      {newtraining: attrs.special3 ||'', skilltraining: attrs.special_skill3 ||''},
      {newtraining: attrs.special4 ||'', skilltraining: attrs.special_skill4 ||''},
      {newtraining: attrs.special5 ||'', skilltraining: attrs.special_skill5 ||''},
      {newtraining: attrs.special6 ||'', skilltraining: attrs.special_skill6 ||''}
      ].filter(o => (o.newtraining || o.skilltraining));`

The .filter() method also takes care of filtering out any completely empty rows, they would be pretty pointless :).

uriele:
I deeply deeply love you @joesinghaus. There was another small mistake but it was all on me (I forgot a [ in my code). Thank you so much

@wing-it : the character sheet is ready, now the only difference is the weapon_skill attributes because I don't use a flat number but a link to a skill to avoid putting back the number (and risking putting in the wrong one) every time

wing-it:
Nice to see people working together, great job!

uriele:
@wing-it @joesinghaus @dhatcher : thank you everyone

PR #2949, opened 2017-08-21
Ghost Lines - Graphic character sheet

timdenee:
A new sheet for Ghost Lines that better matches the print character sheet, with sheetworker to convert old attributes to the new sheet.


kreuvf:
@wing-it This PR added a whole bunch of Windows link files. I am quite sure this does not work as intended.

wing-it:
@kreuvf You're correct, thank you for reminding me, I meant to come back to this. There were a bunch of those link files, which I have removed, but all of the links in the sheet itself are to imgur, and everything seems to work as intended, at least as far as I can tell.

That being said, @timdenee we do recommend that the images themselves be added to the repository and that the sheet link to them on github. Aside from being convenient, this makes it easier for others if they want to contribute.

Your sheet seems to work fine as is, but if you would find the time to submit another pull request with the image files and updated links, we'd appreciate it!

PR #2952, opened 2017-08-24
Added Shadow of the Demon Lord in spanish

Nixitro:
Translated character sheet of the Shadow of the Demon Lord


wing-it:
Since we now have a translation system in place, we really don't want to add multiple versions of sheets for different languages (you may have seen some sheets that have multiple versions, generally these were created before we implemented the translation system and were grandfathered in). We would appreciate it if you would instead add a translation to the current sheet instead, you can find instructions here: https://wiki.roll20.net/Character_Sheet_i18n

PR #2957, opened 2017-08-28
SWN Revised v1.2.0

Karlinator:
Used a conditional statement to apply a -2 unfamiliarity malus if the combat skill was at -1.

Also resolving some conflicts. It appears me and Kevinsearle have been working similar problems at the same time.


wing-it:
Looks like you're removing a bunch of fields and replacing them with a repeating section. Seems like a positive change, but would it be possible to create a sheet worker to migrate the old data into the repeating fields, so that existing users wouldn't lose anything?

Karlinator:
That should be doable. I'll get right on that.

Karlinator:
Done. The data will now be migrated and then cleared from the old fields.

svmnotn:
Could you make burst default to no? it makes more sense for that to be the default since you can't burst all weapons.

kevinsearle:
@Karlinator Can we make these version changes stick to semver type versioning? I just submitted a PR a few days ago for a 1.1.2.

kevinsearle:
@Karlinator Also, perhaps we should either make you the maintainer or coordinate better.

Karlinator:
Actually, if we consider the attributes to be the "public API", then I did deprecate things this update. I don't know if we should do that, but yeah.

kevinsearle:
It's yours!

Karlinator:
@kevinsearle I'll make some changes to how we organise development and stuff. If you want I can give you write access to my fork so we can coordinate there. That's probably better than both of us just working on things individually and then submitting pull requests here.

kevinsearle:
@Karlinator I will probably just move on to other projects. I would still like to have some credit in the ReadMe.md and sheet.json files but otherwise redirect all bugs and feature requests to yourself.

PR #2960, opened 2017-08-29
AGE System 1.16

GhostMatter:
- I'm hoping this fixes the Default Sheet Settings once and for all.
(it would be helpful to be able to test it with a custom sheet instead of waiting for the merge each week)

### DIFF
--- a/AGE System/sheet.json
+++ b/AGE System/sheet.json
@@ -4,7 +4,7 @@
    "authors":"Olivier L.-L.",
    "roll20userid":"594486",
    "preview":"AGE System.png",
-   "instructions":"A sheet made for the AGE family of games: Dragon Age, Fantasy AGE, Titansgrave and Blue Rose. A tab with the data from the old sheets will be available until the end of 2017, allowing migration to this new sheet. An API script (TotalRolled for AGE) is also available to totals rolls and calculate stunt points.",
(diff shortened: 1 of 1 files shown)


Karlinator:
I'm just going to jump in here and agree that there should really be a way for people without Pro accounts to test the changes they are making to character sheets. Some kind of system where you can set up custom sheets in a game, but only in games where you are the only player. Still keeps the ability to have custom sheets for a campaign as a Pro feature but allows Free and Plus users to actually contribute.

GhostMatter:
Oh, that's wasn't what I was telling about. I was talking about Sheet Default Settings. I can't test them at all, Pro or not. On top of that, they are barely documented.

But... I definitely agree with you. My campaigns are on hold right now, so I'm basically paying Roll20 to create content for their website. I enjoy it and appreciate the opportunity, but it's still bothersome.
As for making custom sheets as a single player, that still has potential for abuse. I'd suggest giving that option for Sheet Authors and contributors only.

GhostMatter:
Any further discussion should go in the forums I think, as not to clog GitHub!

GhostMatter:
[Here's a forum thread](https://app.roll20.net/forum/post/3457170/allow-sheet-authors-to-make-custom-sheets)!

PR #2966, opened 2017-09-02
Updates & Corrections to Star Wars Revised RPG

traec:
I'm following up on the pull request #2893 I've deleted the additional html file. Submitted a new file with correct line endings and the same name as the original html so it will overwrite with the changes. Sorry for the delay on this. Let me know if there are any further issues.


Hotbloodhypno:
looking at the character sheet .. while in a game.. 
i found 2 issues...

1  where it states MELEE WEAPON BONUS and RANGED WEAPON BONUS  the boxes where the calculated Total is displayed..  is always empty  no numbers !

2  when pressing the ROLL for ranged and melee weapons  .. the printed results  states  
ATTACK:  0 vs.  AC  (the result always prints '0' ) even tho if you move the curser over the '0'  you see the caculated roll.. and result ..  which is not '0'..

PR #2982, opened 2017-09-07
Marvel Heroic Roleplaying: fixing specialties

wyvren2000:
Adding specialty name and fixing specialty dropdown.


wyvren2000:
Fixed an issue with a failed sync on my fork that resulted in several months of commits to be identified as requiring a merge.

PR #2985, opened 2017-09-08
CoC 5e script changes

Gussancho:


Gussancho:
ad9ebde are changes to the fate core with rolls sheet.
1. The image was an html with the format changed!
2. No roll worked outside the sheet.
3. I added i18n. When I see it working I will add the spanish translation myself.

PR #2995, opened 2017-09-11
Corrected NPC Sheet + Size Modifiers

shukinhekon:
Added some new Javascript code to correctly handle the size modifiers, while making bug fixes with NPC sheet as well.


wing-it:

PR #3000, opened 2017-09-13
Fate Core with rolls small error

Gussancho:
I wrote Extra where goes stunts. Solved


wing-it:
In the future, it would be helpful if you would submit changes to each sheet as a separate request. Thanks!

PR #3012, opened 2017-09-20
MERP Character Sheet

LyricalStreetPoet:
New sheet for Middle Earth Role Playing, which is a simplified Rolemaster ruleset published by ICE back in the day.


wing-it:
There doesn't seem to be anything included in this commit. Can you add the files for the sheet?

PR #3031, opened 2017-09-26
MERP CSS

LyricalStreetPoet:
Stylesheet to go with HTML.

### DIFF
--- a/MERP/MERP.css
+++ b/MERP/MERP.css
@@ -1,0 +1,59 @@
+#number {
+  width: 5em;
+}
+
+/* Tabs */
+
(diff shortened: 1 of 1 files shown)


wing-it:
Could you resubmit this sheet with the HTML and CSS in the same pull request? That makes it a little easier for me to review. Also, could you please include the sheet.json? Instructions can be found [here](https://github.com/Roll20/roll20-character-sheets#contributing), and it may be helpful to look at the sheet.json for another sheet to see an example.

If you could also include a preview image, that would be great. Once you make these changes, this sheet will be ready for me to merge. Please let me know if you need any help!

PR #3041, opened 2017-10-02
Bugfix

Zakarik:
Bugfix for all autocalc who bug after the last update of Roll20.

### DIFF
--- a/Knight/knight.html
+++ b/Knight/knight.html
@@ -1,4 +1,4 @@
-	<script type="text/worker">
+<script type="text/worker">
 	/* ---- BEGIN: TheAaronSheet.js ---- */
 	// Github:   https://github.com/shdwjk/TheAaronSheet/blob/master/TheAaronSheet.js
 	// By:       The Aaron, Arcane Scriptomancer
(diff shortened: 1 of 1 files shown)


wing-it:
I'm a little confused about what's going on here. for example, this change: `name="attr_energieBarbarianMax" value="@{energieBarbarian_max}"`. Since you changed the attribute name from "energieBarbarian_max", I don't see where that attribute is being set now.

Do these fields work with brand new characters? I'm thinking this will work for existing characters because that attribute is already set, but I'm concerned that new characters will not get that attribute set and those fields will break. Does that make sense?

Zakarik:
The truth is, they're already set earlier. If you look at line 1628, you will see for example <input type="number" class="center" min="0" style="width:100%" name="attr_energieBarbarian_max" value="60+@{energieBarbarianModif}" disabled />.

Sorry about my English, I use an automatic translator.

wing-it:
Ok, that makes sense! Sorry, I just wanted to make sure since I didn't see it.

We have a fix going out today to address the autocalc bug. Would you like to hold off on this update and just see if that fixes it, or would you rather me go ahead and merge this in?

Zakarik:
I'd rather you merge. Maybe today's update will fix it or not. I prefer to be sure, by my update, that the bug will be fixed for this week. The players are counting on me for that.

wing-it:
Sounds good to me!

PR #3046, opened 2017-10-04
Fixed Defenses rolls

dougansf:


Solarswordsman:

Please change the line endings to Unix-based and submit a new commit.

dougansf:
I have received this error response after nearly every commit.  Most of the time it is fixed by the Roll20 crew and pushed forward.

The last time, I copied the Roll20 fixed version into my working version before making edits.  Those edits made no new lines or carriage returns.

I do my work inside the Roll20 website.  The only “text editor” I use is Github for pushing to you.

I am confused how this is still happening, and have not found an easy way to fix it to make all our lives easier.

Solarswordsman:
Gotcha.  Sorry, I'm filling in this week so I hadn't kept up with these changes.  I'll do it for you this week, no problem.

The point at which this is happening is likely in GitHub Desktop if you're using that.  There might be useful information here, after a quick google search: https://lydiapchung.wordpress.com/2012/06/01/github-for-windows-line-endings/

I will try to assist you with this issue going forward as much as I can.  I definitely get why this is frustrating.

dougansf:
Thank you very much.  I will try this when I get a chance.

dougansf:
My chance came earlier than I expected.

I tried to update Github Desktop to 1.0.7 but I'm already running the most
recent version, 1.0.3.

I've contacted GitHub to see if they can walk me through this process.

Solarswordsman:
It's possible the process you're using to copy from the repository files (in GitHub Desktop) -> Roll20 editor -> GitHub Desktop is the problem. 

It could also be that your browser is not respecting the Roll20 editor's line endings when you copy from it, and using Windows-style instead, in which case you'd need to configure GitHub Desktop to always use the correct line endings whenever saving the file, even overwriting them when you copy/paste.  I'm not familiar with the GUI of GitHub Desktop so I can't assist you there.

The "simplest" solution to the problem might be to use a mature text editor (e.g. Sublime Text), manually set the line breaks directly in the file on your computer once you're done making changes, and then either use git's command-line or the Desktop app *only* to push those changes to your fork, never opening the files themselves.

Feel free to continue responding here if you need more help.  I am going to merge this, as I've fixed the line endings here.

PR #3060, opened 2017-10-09
 Pathfinder Old/Legacy small syntax fixes

vince-roll20:
PF old/Legacy; 
a couple small fixes for syntax errors
- removed backslash on input
- changed inline css from align:left to text-align:left


vince-roll20:
Thanks

PR #3076, opened 2017-10-13
Non-bfrpg input field removed

pstubbs1020:
Discovered a non-bfrpg input field on the sheet and removed the code.


pstubbs1020:
Had to go in and fix more errors. Will do a new pull once it is fixed.

PR #3123, opened 2017-10-26
[KURO] Initial commit: design only

slezier:
Hello,

following the initial commit of Kuro character sheet.

This commit cover only the original design of the character sheet.

Thanks


wing-it:
Your sheet.json was missing a few required fields, which I went ahead and added for you, along with a preview image. However, I don't have your roll20 user ID#, so I couldn't add that. It's optional but is what we use to give your username the 'Sheet Author' tag. Go ahead and add it in the next commit!

More about the sheet.json file here: https://github.com/Roll20/roll20-character-sheets#contributing

PR #3132, opened 2017-10-31
Remade the sheet from scratch, fixing layout, improving compatibility and optimization.

Nimgaard:


wing-it:
This is a pretty big update! I just wanted to get an idea of how this is going to affect current users of the sheet. Mainly, I'm concerned if any of the attribute names have changed, which could cause current users to lose data. This can be solved with a sheet worker to convert the old attributes over to the new sheet.
I apologize if you've already considered and accounted for this. Normally, I look through the changes and try to determine this for myself, but there were too many changes here for me to be able to do that without spending a long time on it.

Nimgaard:
It’s already taken care of. This remake is based on my update from last month, which already updated all the old attributes (from the sheet from 2016) to the new ones. This remake has the same conversion sheet worker. I also tested it with a sheet from the 2016 version, my October version and a new sheet from this version. As far as I tested nothing was missing in the conversion (except for inputs in the 2016 sheet which didn't have attributes assigned, which would be unsaved anyway)

On 9 Nov 2017, 20.51 +0100, wing-it <notifications@github.com>, wrote:

This is a pretty big update! I just wanted to get an idea of how this is going to affect current users of the sheet. Mainly, I'm concerned if any of the attribute names have changed, which could cause current users to lose data. This can be solved with a sheet worker to convert the old attributes over to the new sheet.
I apologize if you've already considered and accounted for this. Normally, I look through the changes and try to determine this for myself, but there were too many changes here for me to be able to do that without spending a long time on it.

Nimgaard:
I noticed an attribute that didn't get copied from the October version (because the old attribute was a in a repeating_field and the new is a simple textbox), but I got it working. Committing new changes.

PR #3144, opened 2017-11-06
Final Fantasy d20 (FFD20) v1.1 Update

Azurift:
__Version 1.1 (released 11/06/2017)__
Small update. Changes were:
*	Fixed CMB and CMD autocalculations
*	Added a d4 option for class HD
*	Added a quantity/charges (wands) field for inventory items
*	Changed gil, monster XP and CR, to text fields to accomodate for commas
*	Added craft, perform, and profession skills to the monster skills section
*	Fixed the formatting for the monster skills section -- should look better for Firefox users


wing-it:

PR #3146, opened 2017-11-08
New Impulse Drive sheet.

tandroy:
With the blessing of game and original sheet author, Adrian, I have replaced his version of the character sheet with an exciting new version.


madadric:
This version is fantastic, I can't wait to see it go live.

wing-it:
Could you give me an idea of how this update will affect current users of the sheet? Mainly, I'm concerned if any of the attribute names have changed, which could cause current users to lose data. This can be solved with a sheet worker to convert the old attributes over to the new sheet.

Normally I look through the changes and try to determine this myself, but the way you've done this request (deleting the old files and adding new ones with a different name) makes it difficult to compare your changes to the previous version on github.

tandroy:
Unfortunately, this sheet was not created with the intention of replacing the official sheet. It was only going to be something that I used with my own group.

As a result, I paid very little mind to the original sheet and its attribute names. It's a safe bet that most everything has changed. I can put together a list of the original attribute names and their corresponding new names, but it will take a bit of time.

wing-it:
If that's the case, I'd rather not merge this right now because I don't want to run the risk of interrupting people's games and making them lose their characters, so I'll close this pull request for now.

If you're willing to go through and make this compatible with the old sheet so that current users won't lose their data, that would be great. If are able to make those changes, just resubmit another pull request.

If you start looking at this and realize that it is going too difficult to convert easily, let me know and we can discuss some other options.

PR #3153, opened 2017-11-10
Update Cyberpunk2020.css

viperc46:
The previous dark grid image was removed. This is just an updated image so that way it should work.

### DIFF
--- a/Cyberpunk 2020/Cyberpunk2020.css
+++ b/Cyberpunk 2020/Cyberpunk2020.css
@@ -1,7 +1,7 @@
 /* ------------- Background ------------- */
  
 .charsheet {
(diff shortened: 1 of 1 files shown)


wing-it:

I'm going to go ahead and merge this in, but thought I'd let you know what we consider best practice fore images.

wing-it:
Just to let you know, the original sheet author submitted a pull request (which I didn't get to until after I did yours) fixing this same issue, so I've used his changes.

Just didn't want you wondering why your changes weren't showing up on the sheet!

viperc46:
That’s more than fine. Thank you for letting me know about the images though. I’ll keep that in mind for the future! 


Best,
Viper

PR #3166, opened 2017-11-14
Corrected file type for image file in JSON

spacebandito:
Preview file type was incorrectly listed as JPEG instead of PNG, as the original was a JPEG type file. The JSON file has been updated to reflect correct file type to prevent conflict. Apologies for the slip.


spacebandito:
Issued was already resolved by admin

PR #3167, opened 2017-11-14
[Pathfinder] sheet not working at all paste error?

plutosdad:
something went wrong with the last file
pathfinder sheet not working AT ALL!!!

i did a diff, the timestamp is exactly the same, but i see differences in the code. IDK how that happened.
here is the file i meant to load, i did a diff with the build and verified it is exactly the same


plutosdad:
@wing-it  can you look at this and reload it to prod ? Pathfinder sheet not working at all for the users. just the html/javascript file. i don't know if i pasted wrong but the timestamp (built into every build we run) is correct but some problems in old code. this is the correct file. IDK how this could happen except something bizarre with pasting it in

wing-it:
These changes should be live now, check and make sure it fixed it!

plutosdad:
thanks for the speedy help!

PR #3172, opened 2017-11-16
2E_Tabbed

tormeron:
Added tabs, more dynamic fields and a proper spellbook, memorization tab with ability to cast.


wing-it:
Since we already have 2 different sheets for 2E, and this is heavily based on one of them, would you consider upgrading the existing sheet rather than submitting this as a new sheet? These all seem to be positive changes, so as long as you keep attribute names the same (to avoid current users losing any data), I bet most people currently using the sheet would enjoy getting an upgrade.

You might try to contact the original authors of the sheet to see if they mind. If you don't hear back from them at all within a few days, I'd go ahead and submit the new pull request. If they strongly object to this upgrade, we'll reconsider having this be its own sheet.

We have so many sheets on the one-click menu, we try to avoid having multiple sheets for a single system whenever we can. I'm going to close this pull request for now. If you have any problems re-submitting, let me know.

PR #3175, opened 2017-11-17
Shadow of the Demon Lord v4

joesinghaus:
* Finesse and Perception attacks
* Fix some of the spacing and made the code more transparent
* Fix attack roll 20+


joesinghaus:
@wing-it I don't exactly know how to do default values for checkbox for sheet default settings. My first attempt of "default": "desired-value" didn't seem to work. I'm now trying "default": "checked", but I have no idea what I'm doing. Is this the correct way to indicate that a checkbox should be checked by default in the sheet default settings?

BuildAndDie:
checked="checked"

wing-it:
This was actually kind of hard to find, but I did confirm that the correct syntax is `"checked" : "checked",`

I'm adding sheet.json documentation to my to-do list, this should be on the wiki.

wing-it:
Also, feel free to let me know in the pull request if there's something in the sheet defaults you're trying to do. I can test sheet.json changes locally for you and let you know if your changes are doing what you want them to. I can't promise I'd always get to it right away but hopefully it would be faster than waiting a full week for the next code push.

PR #3182, opened 2017-11-19
Update 1ESheet.html

cowtankerous:
fixed the typo on exceptionalcharisma (it was mistakenly placed as exceptionconstitution).

PR #3184, opened 2017-11-21
Ad&d1e

vince-roll20:
- this pr includes Cowtankerous pr #3182 for exceptional  CHA and CON
- fixed some issues with the last update
- hit table attributes were converted to THAC0, but not all users desired that.  Sheet now includes the ability to enter to-hit ac values based off of DMG as well as THAC0.  
- newly added non-weapon prof repeating section had issues where the name attribute was not being set.  same problem with repeating ability section.
- adjusted widths of the exceptional ability attributes so the sheet windows would not have to be stretched to accommodate the additional width.
- fixed alignment/layout of non-weapon prof section
- sheet-wide typos/syntax fixes


vince-roll20:
Thanks

PR #3192, opened 2017-11-23
Ad&d1e text attribute fix

vince-roll20:
- added input type="text" to text input fields.

### DIFF
--- a/AD&D_1E/1ESheet.css
+++ b/AD&D_1E/1ESheet.css
@@ -1,4 +1,9 @@
-/* background toggle*/
+/* abilities */
+div.sheet-abilities tr td.sheet-center:first-child {background-color: #1f1f1f; color: #fff; font-family: serif; width:4.5em;}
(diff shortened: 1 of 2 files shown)


vince-roll20:
I have a big favor to ask.   Would it be possible to run this PR https://github.com/Roll20/roll20-character-sheets/... ahead of the regular cycle?  History; although not the original sheet author for the 1e sheet I have primarily been the author responsible over the last few years.  Another user(Cowtankerous) made some changes to the sheet a few weeks ago without contacting me or informing the community on the forums.  These changes caused some problems with the sheet and miffed some of the sheet's users.  I intervened and made an update to the sheet last week to accommodate the changes and address some of the issues with the update.  In doing so, I made a HUGE mistake and accidentally deleted all the type="text" attributes on the text input fields.   An errant find/replace sheet-wide no doubt.  I know... Major faux paux on my part.  Anyone that opens their 1e characters with this version of the code will more than likely overwrite the text-based input attributes with default values!  I understand if you cannot expedite this update, but I believe it would at least help limit any collateral damage suffered by the community at my expense.  Thank you. - Vince

PR #3212, opened 2017-12-01
L5r 4e sheet

rdefendi:
I haven't been able to get anyone to test my changes on the 4e sheet, so I've created a new version, like Khendar did with the 5e Shaped sheet. This will allow users to switch to the sheet and roll back to a backup if there's a problem with the conversion. If I did this right, it should only pull the new L5R 4e 2.0 Sheet, not the L5R Fantasy Flight Games sheet.


wing-it:
Could you give me some details on why this is necessary? Do you think it particularly likely that this update will break people's characters?
Generally, we prefer not to do things like this as it's somewhat confusing to users who don't spend time reading the character sheet forums.
If you've tested this yourself as much as you can and have found it to be pretty stable, and that it upgrades in place very well, I would say that we could just go ahead and push this out as an update to the main sheet. In this case, its likely that no one at all will be negatively impacted, and if some are, it's likely it would be a small enough number that you can assist them on a case by case basis.
If you think it very likely that this will break, but that this is still a necessary update, we can submit this as a new sheet and deprecate the old one. This will basically remove the old sheet from the list of available sheets without removing the sheet itself, meaning current users will keep using the old version, but any new sheets created will use the new version. The drawback of this approach is if any of the current users switch to the new sheet, they will be unable to switch back to the old one.
I'm going to close this request for now. When you decide how you would prefer to move forward, you can submit a new request or I can re-open this one.

rdefendi:
Do you receive replies to this email?

I’ve been trying to get a second test forever. I’ve tested pretty well, but one set of tests is far from conclusive and these changes are pretty extensive, internally. I don’t want to break hundreds of characters in long-standing campaigns. I’ve sent PMs to some people at Roll20 to see if there was a quick rollback option if this did happen, but I’ve never received a reply. So this was the only solution I could come up with. It’s how they handled the 5e shaped sheet, which is where I got the idea. Without buyin, I’m really hesitant to force this change on people without their consent.

Could you give me some details on why this is necessary? Do you think it particularly likely that this update will break people's characters?
Generally, we prefer not to do things like this as it's somewhat confusing to users who don't spend time reading the character sheet forums.
If you've tested this yourself as much as you can and have found it to be pretty stable, and that it upgrades in place very well, I would say that we could just go ahead and push this out as an update to the main sheet. In this case, its likely that no one at all will be negatively impacted, and if some are, it's likely it would be a small enough number that you can assist them on a case by case basis.
If you think it very likely that this will break, but that this is still a necessary update, we can submit this as a new sheet and deprecate the old one. This will basically remove the old sheet from the list of available sheets without removing the sheet itself, meaning current users will keep using the old version, but any new sheets created will use the new version. The drawback of this approach is if any of the current users switch to the new sheet, they will be unable to switch back to the old one.
I'm going to close this request for now. When you decide how you would prefer to move forward, you can submit a new request or I can re-open this one.

wing-it:
Sorry your previous messages got overlooked! I can understand why private messages could go unnoticed as some of us get so many that they're hard to keep up with. In the future, emails to team at roll20.net are a better way of contacting us, or you can comment on a pull request here and I'll be sure to see it.

If you've tested this yourself and haven't found problems, and posted on the forums informing people of the change and asking for testers (even if you didn't get many responses), you've already followed our suggested procedure for a big update like this. If you believe your changes to all be positive, or necessary for future development of the sheet, I would say let's go ahead and push it out as an update to the main sheet. If you do so I can keep a close eye on GitHub comments so that if you find a problem and need a rollback, I can get to it as soon as possible. Typically if we find out about a breaking character sheet change during work hours we can push out a user's fix almost immediately, though it may be longer depending on when we find out about it.

We have done what you're asking in the past, but these are typically special cases and are done because of the sheer numbers of people using the sheet, as is the case with the 5e shaped sheet.

rdefendi:
No worries. What’s the best way to contact you for a quick rollback if
there’s a problem?



*From:* wing-it [mailto:notifications@github.com]
*Sent:* Wednesday, December 6, 2017 5:21
*To:* Roll20/roll20-character-sheets <
roll20-character-sheets@noreply.github.com>
*Cc:* rdefendi <robert@robertjdefendi.com>; Author <
author@noreply.github.com>
*Subject:* Re: [Roll20/roll20-character-sheets] L5r 4e sheet (#3212)



Sorry your previous messages got overlooked! I can understand why private
messages could go unnoticed as some of us get so many that they're hard to
keep up with. In the future, emails to team at roll20.net are a better way
of contacting us, or you can comment on a pull request here and I'll be
sure to see it.

If you've tested this yourself and haven't found problems, and posted on
the forums informing people of the change and asking for testers (even if
you didn't get many responses), you've already followed our suggested
procedure for a big update like this. If you believe your changes to all be
positive, or necessary for future development of the sheet, I would say
let's go ahead and push it out as an update to the main sheet. If you do so
I can keep a close eye on GitHub comments so that if you find a problem and
need a rollback, I can get to it as soon as possible. Typically if we find
out about a breaking character sheet change during work hours we can push
out a user's fix almost immediately, though it may be longer depending on
when we find out about it.

We have done what you're asking in the past, but these are typically
special cases and are done because of the sheer numbers of people using the
sheet, as is the case with the 5e shaped sheet.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Roll20/roll20-character-sheets/pull/3212#issuecomment-349623339>,
or mute the thread

wing-it:
I'm sure to see comments on pull requests (like this one).

PR #3220, opened 2017-12-02
[VtM 5e Alpha] translation tags updated

Anduh:
Please do not revert these changes when updating the approved.yaml file like last time. Earlier I only bothered to change the translation.json and en.json, but now fixed all the rest too. 

**Changes:**

- Health, willpower and other stats formatted for clarity
- added dice pool rolls(slightly moddified) back to the sheet
- (redundant) Other Traits removed
- Backgrounds section changed to a textbox
- WOD roll template added, still unused
- translation tags updated for all languages
- minor tag, textfield improvements
- json file fixed


wing-it:
I apologize if your changes were reverted last time! This was not intentional, but happens occasionally, because our process of updating translations is automated and can potentially overwrite changes if things aren't done properly.
I happened to notice in this pull request that your translation files have some invalid json (looks like some missing commas). This is likely what caused your changes to not get updated with the last request. I suggest you fix the errors and use [jsonlint](https://jsonlint.com/) or a similar service to make sure your json format is valid, then update this pull request.
I'll also point out that it's only necessary to change the translation.json file unless you are actually adding translations. Updates to the translation.json will be automatically propagated to the other files.

Anduh:
Okay, thought it was something like that, it wasn't that bad. I'll check the validity of json files in the future.

PR #3227, opened 2017-12-04
[KURO]  Design change + first roll for characteristic

slezier:
Hello,

here a new version of Kuro sheet.

Adding, font thru face-font instead of @import, embedded svg, roll for primary characteristics

Thanks
Sylvain


slezier:
Error while commiting.

Sorry

PR #3240, opened 2017-12-07
New TormentaRPG Sheet

johannborges:
A work in progress, but beautiful character sheet comparable with the original PDF one. More user friendly.


wing-it:
We typically try to avoid having multiple sheets for the same system, and currently we already have two separate sheets for TormentaRPG. Is this sheet a significant upgrade to the other available sheets? Has the community requested the changes in this sheet? Please give me a little more info on why there needs to be this third version of the character sheet.
Have you considered changing one of the existing sheets instead? You might even get help from the original creators if you contact them and let them know what you plan to do.

johannborges:
Yes, it is a upgrade. Just by seeing the preview image you can see. Last year I attempted to create this sheet, but had no time, and gave up on the project. But the community was very excited.
I mean, the first version is not very good and could use a upgrade, but as you can probably see, this sheet is entirely different. Many players would lose their information. About the second one that came out a few months ago, I have to say, I think mine does a better job at organizing the character information.
I have no contact with the authors, but knowing the community, I think the authors would not be open to a "brutal" change, as my sheet has made.
De: wing-it <notifications@github.com>
Enviado: terça-feira, 12 de dezembro de 2017 00:04
Para: Roll20/roll20-character-sheets
Cc: johannborges; Author
Assunto: Re: [Roll20/roll20-character-sheets] New TormentaRPG Sheet (#3240)


We typically try to avoid having multiple sheets for the same system, and currently we already have two separate sheets for TormentaRPG. Is this sheet a significant upgrade to the other available sheets? Has the community requested the changes in this sheet? Please give me a little more info on why there needs to be this third version of the character sheet.
Have you considered changing one of the existing sheets instead? You might even get help from the original creators if you contact them and let them know what you plan to do.

johannborges:
So... are there any news on this?

wing-it:
Sorry for taking so long to get back to you, the holidays have messed up our schedule somewhat.

After some discussion, we've decided to deprecate the old version of this sheet, as it appears that the original author has stopped maintaining it, and we'll replace that sheet with this one. What this means is that the old sheet will no longer be available to select from the character sheet menu, but any games that are already using it will continue to use it. New games that select "Tormenta (Portuguese)" as their character sheet will use this sheet.

We have one request. Could you please add a warning to the description in your sheet.json file that informs players that this sheet is different from the previous version, and that characters made with the old version of the sheet will probably not be compatible with the new sheet? If you can add this, I'll go ahead and merge this in next week.

johannborges:
Yeah, sure. I can do it. But I would ask you to keep both of them until I can say this one is 100% ready to be played with. Because I don't have any premium account, and I can't test if the fields are working as intended.

wing-it:
Ok, I'm going to go ahead and add this then. When you've confirmed that everything works, make a new pull request adding the warning to the sheet.json, which will let me know its ok to deprecate the old sheet.

PR #3252, opened 2017-12-11
SportsBrawl Sheet

LostWestMedia:


wing-it:
Is there a publicly available ruleset available for this system? I can't seem to find "Sports Brawl".

LostWestMedia:
There will be as soon as they are released on the Marketplace.  I have product manuals created in the individual campaigns, but nothing has been made public yet.  Dean has examples of the product manual in the Lost West Media demo campaign.

PR #3272, opened 2017-12-18
Shadowrun (advanced) bugfix

darkxyro:


inwils:
Hi there,
Not sure if I am using this correctly - but we use your character sheet for SR on roll20 and there seems to be an issue about the initiative roll - are you aware of this?
Thanks
inwils

PR #3278, opened 2017-12-20
Blades in the Dark: Sheet v2.6 and Korean translation

joesinghaus:


joesinghaus:
@wing-it About the translations again — I made some adjustments with this commit, most of it superficial (exchangings hyphens for emdashes and such), one of them substantial (changed translation for "set_up_a_teammate" to help with Korean translation). Anyway, as previously witnessed, none of the changes to translated values are actually showing up in the xy.json language-specific files. So at this point, only the new Korean translation is correct.

PR #3281, opened 2017-12-25
Addition of HarnMaster 3 character sheet

Streamweaver:
Hello.  I would like to to submit this character sheet for use with Harn Master 3.  Credits and links to the RPG are included in the readme.  Info in sheet.json as requested.  The sheet has been fairly well tested with my group and has no known bugs at this time.


suntzu777:
You do know that there is already a harn sheet on Roll20 which I created , perhaps you may want to modify that one rather than submit an additional sheet for the same system

Streamweaver:
I'm aware, this sheet came out of trying to use that one.  The differences are so large as to not be remotely compatible.  In my groups opinion (and mine) the other is extremely difficult to use and it would take a total rewrite, so I created one from scratch.

suntzu777:
Ok well it may not be accepted as Roll20 does not like multiple sheets for the same system

Streamweaver:
I didn't see you're comment that you created it.  Thanks for the reply.  They may not and that's up to them.  The two sheets are so different I don't think they could be reconciled.

suntzu777:
Well good luck , personally I don’t mind that there maybe two sheets and merry Christmas

mjthurston:
I tried to use the roll20 harn character sheet and scrapped it in 5 minutes.  It does not work.  It also puts out mystery numbers in addition to other numbers that do not add up correctly.  So I had to make my own character sheet in the biotab section.  I then added skill totals and other items to the abilities tab.  It worked very well and I would have continued to use it until Streamweaver came up with a solution to the problem.

I have personally tested the new sheet and it works great.  I have uploaded it into my custom character sheet and converted my player characters over.  Streamweaver's sheet looks like a professional character sheet and the rolls that break out in color for CS, MS, MF, and CF are over the top.

Great job Streamweaver.  I hope your charcter sheet is accepted.  Harn DM's will be very happy for a character sheet that is usable.

suntzu777:
Mjthurston do you have specific instances of why it didnt work for you as i would be interested in hearing your observations , personally i ran a harn weekly campaign for 4 players with it for 3 months with it , there are some issues but those relate to roll20 dice and how the rolls are related to the results but a hover over the roll in the chat does allow those results to be resolved using the quick reference sheet included in the rules , with regard to the sheet and the rules there are options in combat that require player choices so there are multiple buttons to reflect those choices which is what i prefer rather than having multiple popup windows, obviously if the wrong numbers were put into boxes then that effects the results as there are multiple behind the scenes calculations , by the way this sheet is for harn3 rather than harn gold as there are some subtle differences between the two rulesets which are mutually exclusive

mjthurston:
It does not add up EML correctly.  It has a funky second number added  to the rolls.  I have no idea what this UG thing is supposed to do.  It doesn't calculate conditioning correctly.  The only thing it did was calculate encumbrance and a good place holder for skills and stats.  Rolling from this sheet was not fun.  I had to make my own macros for attack rolls.  Here is an example from the charcter sheet.  No way of knowing your target number.  A second number that was confusing and a write up that didn't sound fun.  

Marcus T. (GM) rolls 87 1470 Efez Egum Attacks With Sword Weapon using 16 Skill.

This is how it looks for us below and we are molding this to HM3.  

 
| Caelynd attacks with a Spear |
| Target Mods: ML(102), Penalty(-10), Mod(0) |
| Target EML: 92 |
| Roll(1d100): 48 |
| Marginal Success. |
| B(4) E(0) P(7) |


Give's a ML, a Penalty and a Mod for aiming.  Then gives you a target number and a roll.  It also gives you a result.  Much cleaner.

I also wish the sheet that is on the roll20 game was cleaner looking.  A nice clean character sheet with black letters on a white background.

If you are wanting to fix your character sheet here are my suggestions.

1.  Fix the calculation.  Conditioning can be improved with training.  You have no way of doing this unless you added it to the physical skills.
2.  Redo the attack rolls.  You can use the above as an example.
3.  Redo the sheet.  Black and white.  You don't have to get fancy.
4.  Not sure if you had a notes section, but that would be nice.
5.  Not sure if you had an inventory section, that would be nice also.

Marcus


    On Monday, December 25, 2017, 5:25:31 PM EST, Rob <notifications@github.com> wrote:  
 
 
Mjthurston do you have specific instances of why it didnt work for you as i would be interested in hearing your observations , personally i ran a harn weekly campaign for 4 players with it for 3 months with it , there are some issues but those relate to roll20 dice and how the rolls are related to the results but a hover over the roll in the chat does allow those results to be resolved using the quick reference sheet included in the rules , with regard to the sheet and the rules there are options in combat that require player choices so there are multiple buttons to reflect those choices which is what i prefer rather than having multiple popup windows, obviously if the wrong numbers were put into boxes then that effects the results as there are multiple behind the scenes calculations , by the way this sheet is for harn3 rather than harn gold as there are some subtle differences between the two rulesets which are mutually exclusive

Streamweaver:
Thanks everyone for their comments.  If I could bring this back to topic. I have a submission I'm making.  I think the world would be a fine place with two versions of Harn sheets as they do different things.  One does one set of things, this does another.  I believe the fan community would be well served by having both sets.

suntzu777:
Thank you for comments I will have a look at the calculations when I ave a chance to do so , by the way “UG” is for upgrade and allows a temporary modifier to be put into the roll calculation and not effect the base calculations. As to the look of the sheet that is down to personal preference from the authors perspective and is outside the scope of these comments , I too see no reason why both sheets cannot exist

wing-it:
@suntzu777 is correct, we generally avoid having multiple sheets for a single system. However, reading through your conversation has answered most of the questions I would normally ask when a sheet like this is submitted, namely, what is different about this sheet as opposed to the other one, and why does it need to be a separate sheet. I'm going to go ahead and merge this, glad to see that it's already been discussed!

mjthurston:
Is there anyway to change the tag on the character sheet?  Harn Master 3 (Simple) makes it sound like it doesn't have everything.  Harn Master 3 would be just fine.  Thank you.

Marcus    On Monday, January 1, 2018, 9:14:49 PM EST, wing-it <notifications@github.com> wrote:  
 
 
Merged #3281.

PR #3309, opened 2018-01-03
Version 2.0

dougansf:
Defense Conditions, Toggles, Auto-Calculating Values, Armor & Shield Worn, Jump Up, Magical Equipment, Chat Notes, Whisper confirmation GM Rolls.


wing-it:

Have you considered how these changes will effect people currently using the sheet? Mainly, I'm concerned with attribute names changing, which can result in lost data for current users. Normally I can look through and see if you've done this in GitHub, but your line endings make it very difficult to do myself, so I'll just ask you instead!

dougansf:
I appreciate you fixing the line endings.  I have not been able to find a good way to manage it myself, despite multiple attempts to research it.  It seems that my lack of experience in coding means I need a little more instruction than I am finding.

I have considered it, and the two games I am in are aware of it.  I have no idea if it’s being used by more than that.  I’m uncertain of my options to communicate with any users I don’t know about.  However, I think the changes will significantly improve the functionality of the sheet.

PR #3310, opened 2018-01-03
Updated to version 1.1.0

KnowlegeRhino:
- Updated translations (3 were missing)
- Fixed an issue with Story Point Tracking
- Updated 2 sections to remove missed Star Wars references.
- Updated the Dice Suggestions to default to "none"
- Made a two sections in the GM sheet make more sense.


KnowlegeRhino:
Errors were found in this version, will upload a new version when they are fixed.

PR #3312, opened 2018-01-03
[Uncharted Worlds] Far Beyond Humanity expansion

RhOmicron:
Added new Origins and Careers
Added Images


RhOmicron:
I've contacted Oliver and got permission.

PR #3322, opened 2018-01-06
Blades: update translations

joesinghaus:


joesinghaus:
@wing-it All translations changes I've made to translations (except Korean) have been reverted again by commit ef7498b...

joesinghaus:
If you don't see this before the next cycle, I guess I'll have to brute-force the update by changing all the relevant translation keys...

PR #3324, opened 2018-01-07
Heroquest Glorantha, Runequest G QS, Over the Edge 2e

Lockbox313:
Heroquest Glorantha, Runequest Glornatha  Quick Start, Over the Edge 2e


wing-it:
I have some questions about the Runequest Glorantha Quick Start sheet. It seems from the description that it's not a fully functioning character sheet and is only intended as a demo. Would this sheet be usable in an actual game? I'm hesitant to add a new sheet that has such limited usefulness. Is there a reason this isn't a fully functional sheet?

On a side note, I would go ahead and merge the other 2 sheets in, however it's difficult for me to separate the files out. In the future, this can be avoided by submitting a separate pull request for each sheet.

Lockbox313:
Yes it is use able  here a game setup with the sheet.

wing-it:
The readme makes it seem like it's only usable for demo purposes and would not work for an actual game, is this not the case?

Lockbox313:
The only difference is category modifiers which are not used in the Quick Start.

Lockbox313:
Here is a link with a game setup
https://app.roll20.net/join/2825996/tTE1wg

PR #3325, opened 2018-01-09
Dungeon World Official - Two small quality-of-life improvements

timdenee:
Here are two small tweaks:

- Makes the Load input manual instead of auto-calculated. While not as smart, the auto-calculation doesn't seem to be working at the moment so manual entry would be preferable. 
- Add an option under "global modifiers" to let a character roll 1d6+1d8 instead of 2d6; this is crucial for moves like the barbarian's Herculean Appetites.

### DIFF
--- a/Dungeon_World_Official/Dungeon World.html
+++ b/Dungeon_World_Official/Dungeon World.html
@@ -382,6 +382,12 @@
 								<input type="number" name="attr_gmod_ongoing_mod" value="0" />
 							</span>
 						</div>
+                        <div class="flex-row flex-left">
+							<input class="enable-mod" type="checkbox" name="attr_barbdice_on" value="1" />
+							<span>
+								<span class="label">Roll 1d6+1d8 for Moves </span>
+							</span>
+						</div>
 						<div class="flex-row flex-left">
 							<input class="enable-mod" type="checkbox" name="attr_dmg_on" value="1" />
 							<span>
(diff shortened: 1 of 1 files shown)


Solarswordsman:
Hi Tim,

Auto-calculating load seems to work fine for me.  If you can spot a problem or have a resolution that makes the auto calculation work for you, then we can definitely take that, but I would rather not remove that functionality from the sheet.

The barbarian change is good, though I don't think it makes much sense as a global modifier.  Perhaps you should move it to the settings page as another section there, much like level 0 spell name.  This is what the `rolltype` attribute was made for anyway, so you can just set it there with a simple toggle instead of using sheet workers.

(Note that this change--Herculean Appetites--and more are coming in a future sheet update, but we're happy to accept tweaks like this in the interim since our hands are quite busy with other things as well!)

timdenee:
Hey Jeffrey, totally fair - I'll revise the pull request in a bit. I was also going to add in a "notes" field at the end of the sheet, are you OK with that? I just find it's always helpful to have a generic notes field... 

Didn't mean to step on your toes btw, I just needed Herculean Appetites for an upcoming game. Incidentally, I would still argue it should be under Global Modifiers since it's something you'll turn on and off quite frequently (rather than a general Setting). You might use it for one roll, not use it for the next, and then use it again on the third roll in quick succession. It would be a lot easier if it could be toggled right next to the Move/Stat buttons. 

Just saw your edit, never mind! Haha.

Solarswordsman:
Sure, definitely throw in a notes field.  I see no downside there, the only difficulty is location.  The CSS for Moves is a bit finicky (the repeating section uses flexbox) and its container is slightly unpredictable in size, so keep that in mind.  Perhaps it deserves its own tab, as there is in the 5e OGL sheet for character backstory, notes, etc.

And yeah, as my edit mentions, agreed; quickly toggling Herculean Appetites is good.  The gmods section is appropriate for that, at least for now.

What issues are you having with the Load attribute?  I'm more than willing to help troubleshoot.

PR #3329, opened 2018-01-09
SWN Revised v1.5.1

Karlinator:
A few fixes that I kind of forgot to submit a while back, and a few I did now.


wing-it:
Just a heads up: 'for="<something>"' will not work as expected in character sheets, as 'for' looks for an element with a matching id, and id's are sanitized out of the sheets to prevent conflicts.

PR #3335, opened 2018-01-09
Dungeon World Official - Herculean appetites and minor tweaks

timdenee:
- Added Herculean Appetites button under Global Modifiers
- Added "notes" section under Sub page (I've used two text areas to simulate two columns; the shorter line-lengths are easier to read than one wide box).
- I added a text area for Race to make it consistent with Bonds and Alignment. In my playtesting, some of the racial features are quite long and it's fiddly trying to read them in a one-line text input.

### DIFF
--- a/Dungeon_World_Official/Dungeon World.css
+++ b/Dungeon_World_Official/Dungeon World.css
@@ -32,7 +32,19 @@
 .sheet-compendium-drop-target.dropping {
 	background-color: yellow;
 }
-
+.sheet-notes{
+        border-radius: 0;
+    box-shadow: none;
+    border: 0;
+    padding: 5px;
+    width:49%;
+    box-sizing: border-box;
+    resize:vertical;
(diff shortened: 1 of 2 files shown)


Solarswordsman:
I put a border on the notes section so that it matches the rest of the sheet.  When I push out version 2 of the sheet I'll probably look into cascading this style a little more cleanly, but for now it's fine on the class.

I also put a minimum width there matching a single-line for the line-height; again, will definitely want this in the cascade later.

As for Races, I thought you should know that there are Moves in the compendium for each Class's available Race(s), which was sort of the intended use when designing the sheet; then one would simply just place the name of their Race in the simple input field.  Still, I'm not at all against both or either use cases being present.

Otherwise, looks good to me Tim!  If you'd like, feel free to add yourself to the authors in the `sheet.json` before this gets merged next week.  Thanks for your contribution!

PR #3337, opened 2018-01-10
[Infinity 2d20] Checkboxes to radio buttons; other minor additions

DrakeRichards:
Changed the checkboxes in each stress track to radio buttons. This will allow the tying of token bars to each stress track.
Changed "Trauma" to "Metanoia" to reflect the change made when the RPG went to print.
Added a "Harm Effects" section and moved the "Traits" section to improve page flow.
Removed the "sheet-check-on" class: no longer used in sheet.

### DIFF
--- a/Infinity-2d20/Infinity 2d20.css
+++ b/Infinity-2d20/Infinity 2d20.css
@@ -141,10 +141,6 @@ button {
 	margin-left: 2px;
 }
 
-.sheet-check-on {
-	border: 1px solid black;
-}
-
 .sheet-check-off {
 	background-color: #E3E3E3;
 	border: 1px solid gray;
--- a/Infinity-2d20/Infinity 2d20.html
+++ b/Infinity-2d20/Infinity 2d20.html
@@ -1,3 +1,7 @@
+<head>
+	<link rel="stylesheet" href="Infinity 2d20.css">
+</head>
+<body class="charsheet">
 
 <div class="sheet-section-basic sheet-section">
 <img class="sheet-logo" src="https://raw.githubusercontent.com/Roll20/roll20-character-sheets/master/Infinity-2d20/Infinityheader.png" alt="Infinity" />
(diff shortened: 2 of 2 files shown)


suntzu777:
I have no issue with you modifying my sheet but it would have been nice to have been asked first . Normally someone would contact the creator of the sheet asking them if they could perform the modifications

DrakeRichards:
My apologies. I meant no offense with this pull request and I am unfamiliar with github etiquette.

wing-it:
It looks like you've changed the names of the FW and RS attributes, which would cause people currently using the sheet to lose the data stored in the old version of the sheet. As I'm not familiar with this system, could you tell me how this might affect games already in progress with this sheet?

It looks to me like you're changes are generally for the better, but if the loss of this data would be a serious setback to players currently using the sheet, you might consider adding a sheet worker to migrate the data from the old version of the sheet to the new one.

DrakeRichards:
The changed fields are the three "health pools" for each character. Losing them would be akin to losing your current HP in D&D: annoying but certainly not debilitating. I don't believe there's a way to make a migration with them either- there are so many possible configurations that a player could make that it would be nigh impossible to make a consistent migrator.

wing-it:
I see what you mean, the current implementation of the checkboxes make it difficult to convert to the new system.

The only thing I can think of would be to move the current layout of those sections to a separate part of the sheet that is hidden by default (hidden in a similar way that other sheets to a settings page, or even just placed at the bottom of the sheet below everything else). You could add a note that players should move their data to the new section because the old one will be removed in a future update. That way, the players will still have access to their previous values and should know that they need to move them. After you've given everyone a couple of weeks to switch to the new section you can remove the old section from the sheet entirely.

This is just a suggestion. If you think this is more trouble than this inconvenience is worth, I'll go ahead and merge this request as is.

DrakeRichards:
It's a good request, and I'll have to add that to the pull. Unfortunately I've encountered other problems when putting the code into roll20: The Momentum radio buttons became vertical and the damage tracker tables extend into the next column. It should be an easy fix, but for now I'll have to hold off on this pull request until those are working properly.

wing-it:
Ok, sounds good. I'll go ahead and close the request for now, you can resubmit when you've worked out those problems.

DrakeRichards:
Resubmitted here: https://github.com/Roll20/roll20-character-sheets/pull/3359

PR #3339, opened 2018-01-10
Updated to version 1.3.0

KnowledgeRhino:
Fixed issues with old SWRPG stuff


wing-it:
Could you please remove the old version folders? This is generally how we structure API script folders, but it's unnecessary for character sheets as you cannot select a different version. The changelog will be saved on Github so you won't lose any data.

KnowledgeRhino:
I copied the format that the SWRPG API Compatible had which is what i engineered this sheet from. If this one is wrong, so is the SWRPG sheet.

PR #3343, opened 2018-01-14
Fix some French translations in the Warhammer character sheet

Nhacsam:


Warhammer-Fantasy-Roleplay-2nd-Edition/WHFRP2.html
```diff
@@ -1977,7 +1977,7 @@ <h4 class="sheet-center">
       <div class="sheet-col-1-8 sheet-center sheet-pad-r-md">
         <input class="sheet-center" type="text" name="attr_height"/>
         <div class="sheet-small-label">
-          <span data-i18n="height">Height</span>
```
Nhacsam: I had a weird bug. The translation in french was "Hauteur de la page" (Page Height in english) instead of "Taille".
My best guess is there was a conflit with one of the own roll20 translation key.
I am not sure that changing the translation key will fix that

joesinghaus: Changing the key to something different than "height" will fix it.

PR #3352, opened 2018-01-16
Minor Changes to the NPC Sheets

KPhan1212:
•	Changed the NPC’s Defenses section to make it more like the PC’s
•	Added an Armor Section for the NPC sheet
•	Changed the NPC’s Attacks Section to make it more like the PC’s.
•	Changed the Repeating Skills Section on the NPC Sheet to be more like the PC skills.
•	Fixed NPC Sheet’s Damage Threshold 
•	Fix NPC Sheet’s Condition Tracker
•	Fix NPC Grapple and Skills
•	Fix PC and NPC Sheet’s Weapon Damage, it should not be counting Non-Heroic Levels for damage.
•	Changed the NPC Grapple to function more like the PC Grapple

### DIFF
--- a/Star Wars Saga Edition/StarWarsSagaEdition.html
+++ b/Star Wars Saga Edition/StarWarsSagaEdition.html
@@ -15,7 +15,7 @@
 	<div class="table left classHeader">
 		<div class="table-row">
 			<span class="table-data long"><label data-i18n="level">Level:</label>		</span>
-			<span class="table-data">	<input type="number" name="attr_level" title="Total number of Heroic levels @{level}" value="1" /> <input type="number" name="attr_level_max" value="[[(floor((@{level}+@{nonlevel})/2))]]"  title="Half Level @{level|max}" disabled="true" /></span>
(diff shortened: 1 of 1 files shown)


wing-it:
How will this affect current  NPCs made with this sheet? For instance, I see that the attribute NPC-CT is being changed to CT, which will cause current users of the sheet to lose that data. I'm not familiar with this system, will losing this data be a major setback/inconvenience?

If so, consider adding a sheet worker to convert the old data to the new sheet. In this case, it should be pretty straightforward.

KPhan1212:
The changes to the NPC sheet was to make it more like the PC sheet, it would automatically calculate all the same things as the PC sheet. 

The specific example you mentioned will not affect current users of the sheet. It's just an improvement to have the Condition Tracker affect all of the sections of the NPC sheet.

Some of the changes I've made to the sheet are pretty radical and would be difficult to convert over. The NPC's defenses and weapons section originally just have you manually code the values onto a text box. For the Weapon's section I've replaced those text boxes with drop down menus that automatically change the values for them. For the Defenses section I had them automatically change with the NPC's level as well as a new section for Armor, which adds to the defenses section. Every other change I've added would automatically calculate itself.

PR #3355, opened 2018-01-16
Add new Eclipse Phase Alternate sheet (v1.1.0)

Bathtor:
Since the other Eclipse Phase sheet hasn't been touched in like 3 years, this is a new one that makes heavy use of sheet workers and roll buttons to automate playing Eclipse Phase.


joesinghaus:
Nice! 👍  You might want to submit the translation.json in a non-minified form for diff purposes?

wing-it:
As a general rule, we don't like to have multiple sheets for the same system unless there is good reason. Since it sounds like the other sheet is pretty out of date, is there any reason to keep it as an option on the drop down menu?
If there's no reason to keep the old one around, I'd like to deprecate it and replace it with this new sheet. This will not delete the old sheet, it will simply make it unavailable to select from the character sheet menu. This sheet would be the only Eclipse Phase sheet on that drop down, but existing games that use the old sheet will keep using the old sheet without interruption.
Let me know if you can think of any reason why I shouldn't do this. Thanks!

Bathtor:
I don't know of any advantage of using the old sheet over this one, but then I am somewhat biased.

If there are any issues that would prevent the use of this sheet as default, I would like to know about them as well, so that I could maybe integrate the missing behaviour.

PR #3361, opened 2018-01-18
Correctios for tagmar charsheet

fabiocmg:
Uodate of Tagmar system character sheet.  There are new functionalities and corrections from previous inactive fields.


wing-it:
Just a quick note, there were a few errors with your sheet.json. First, you had the file named tagmar.json, but it's required to always be sheet.json in order to work properly. Second, you had your description on a few different lines, which isn't supported. Instead, this needs to be all on one line in the sheet.json file, with "/n" to add line breaks.

I've gone ahead and fixed these for you myself, but I thought I would make you aware of them for future reference. Thanks!

PR #3373, opened 2018-01-22
Update to the Impulse Drive Character Sheet update to rules version 3.5

madadric:
The Impulse Drive sheets have been updated for a better look, new
functionality, and to bring features in line with the 3.5 version of the
game rules. While many elements have changed, relevant Attributes have
been tested and kept the same for compatibility with older versions of
the sheet.


wing-it:
> While many elements have changed, relevant Attributes have
been tested and kept the same for compatibility with older versions of
the sheet.

I just wanted to say that I appreciate you testing this, and letting me know that it has been tested. It saves me a good bit of head scratching while staring at the code to see if it will be compatible.

PR #3385, opened 2018-01-23
Add files via upload

BluSponge:
Alternate 7th Sea 2nd edition character sheet, based on original by Coal Powered Puppet and modified in consultation with original author.  Credit provided.


wing-it:
Since this is based on the previous sheet, would it be possible to submit this as an upgrade to the existing sheet? We prefer not to have multiple sheets for the same system wherever possible.

BluSponge:
Sure, I suppose. Do I need to do that on my end, or is it something you can do with the files I’ve already submitted?

wing-it:
Basically, you'd need to check if an existing user can smoothly transition to the new sheet without losing data (data loss occurs when attribute names are changed are deleted). In some cases, a sheet worker may be necessary to upgrade from the old style of sheet to the new one, though since your sheet seems similar to the other sheet I think this shouldn't be too hard in this case. After you've checked this, just create a new pull request with your code overwriting the html and css of the other sheet.

This is a little extra work for you, but we would appreciate helping keep our overflowing character sheet list a little tidier, and existing users will appreciate the upgraded sheet!

BluSponge:
How much data loss is permissible? I don’t recall changing much in the way of attribute names, but there are lots of changes to the infrastructure and some new data in the mix.

I’ll test it tomorrow.

PR #3387, opened 2018-01-23
Revert "[Pathfinder] Fix damage dice changes by size"

plutosdad:
@wing-it  

Please revert these, sorry about this. I don't know what is wrong, i must have uploaded a wrong file not the working file. 

Reverts Roll20/roll20-character-sheets#3374


plutosdad:
@wing-it  it is just the html file that needs to be changed back to the old one

wing-it:
I'll merge this in and pass it on up the chain, but I'm not sure what the timeline will be for pushing it live.

PR #3390, opened 2018-01-23
[Rogue-Trader] Ship weapons fix

Hesko:
Fixed issue with adding ship weapons table

### DIFF
--- a/Rogue_Trader/RogueTrader.html
+++ b/Rogue_Trader/RogueTrader.html
@@ -892,7 +892,84 @@
                 </tr>
             </table>
         </div>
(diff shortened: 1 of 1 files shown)


wing-it:
Can you give me an idea of how this will affect current users? it looks like some of the attribute data will be lost since the attributes were previously not in a repeating section. If this loss will be a disruption to players' games, you might consider adding a sheet worker to update the old attributes to the new ones.

Hesko:
>  it looks like some of the attribute data will be lost since the attributes were previously not in a repeating section.

Oh, thank you, that's true! It seems that problem should be solved if I add the same table before repeating section.

PR #3393, opened 2018-01-25
Change of sheet creator

Lockbox313:
Change of sheet creator


wing-it:
I'm assuming you've already discussed this with the other author, but all the same we'd probably prefer you keep his name and id on the sheet, and add yours after his (separated by a comma). Your notes on the readme and sheet.json files should be enough to direct users to the right place for bugs/maintenance, and we want to make sure the original author still gets credit for his work.

Lockbox313:
I am the original author

PR #3394, opened 2018-01-25
Eclipse Phase V2 : Fix bug & add french translation

Zakarik:


Zakarik:
Capharnaum : http://legrog.net/jeux/capharnaum

PR #3409, opened 2018-01-30
D&D_4E character sheet power change

Bimny:
Made major changes to power section:
-Made 4 power sections; at-will, encounter, daily, item
-Removed second column for powers
-Removed 23000 lines of code by making powers repeating
-Changed the basic macro for powers
-Added multiple text fields for power cards (emote, keywords, target, custom crit, hit effect, effect, special, miss effect)
-Added option to forgo attack and damage rolls
-Added a "None" option to the attack and damage roll dice


wing-it:
Hi there! Looks like these files aren't in the correct folder. I'll go ahead and close this so you can try again, let me know if there is something I can help with.

PR #3411, opened 2018-01-30
D&D_4E Character Sheet Power Update

Bimny:
Placed files correctly (hopefully)
Made major changes to power section:
-Made 4 power sections; at-will, encounter, daily, item
-Removed second column for powers
-Removed 23000 lines of code by making powers repeating
-Changed the basic macro for powers
-Added multiple text fields for power cards (emote, keywords, target, custom crit, hit effect, effect, special, miss effect)
-Added option to forgo attack and damage rolls
-Added a "None" option to the attack and damage roll dice


Bimny:
I made this sheet to be used by my friends and I since we started playing 4E and I realized the sheets needed some work. The problem with this version that I have realized is that because I took out 85% of the hard code any sheet that switches from the old one to this one will lose all of their current data on the powers sheet since I removed powers 1,2,3,4...100 in favor of 4 powers that can be repeated. I don't know if it is possible, but I would prefer if this sheet can be a separate choice and allow games that currently use the old one to keep theirs since they can have upwards of 20 of these they might lose.

joesinghaus:
You can write a sheet worker to take the data from all the existing sections and put it into the repeating ones.

Bimny:
Thank you for putting up with my lack of experience. I'll learn some javascript and try again.

PR #3415, opened 2018-01-31
WIP: V20

jgoppert:
This is a V20 sheet that simplifies the end-user experience and integrates well with health bars etc. I'm using jinja to generate all of the HTML code.

One of the major problems I'm finding running V20 is that it is a pain to keep track of experience and health in game, so this is really focused on fixing those two aspects. Also adding bells and whistles like support for discipline sheets.

**Tasks**
- [x] Freebie Points Spent
- [x] Backgrounds
- [x] Base Points Spent (7/5/3 etc.) (maybe want user to specify primary/ secondary attr/abil to make sure points are correctly spent?)
- [x] Experience Points Spent (any points past (freebie + base = j) get summed with sum_j:n(n*X) type formula
- [x] Handle case for attr/abil with no free points or base points (new disciplines etc, and cost diff)
- [x] Background/discipline xp cost
- [x] Add base points for disciplines/ backgrounds/ virtues.
- [x] Willpower (base = courage)
- [x] Humanity (base = conscience + self_controll)
- [ ] BloodPool (Set max blood pool, trait max, and blood per turn based on generation).
- [ ] Handle lower cost for in-clan discipline (need to check if discipline is in clan dict)
- [ ] Merits/ Flaws (tie in to freebie calc)
- [ ] Health Macro for Carrying Damage from Bash -> Lethal -> Aggravated
- [ ] Compendium Loader (Load a user supplied file of disciplines, avoid copyright issues by not embedding it in the sheet, and making the user type it out from the book in a separate file, but at least something they can keep around and use for games) Include format in docs.


jgoppert:
image
Current status.

jgoppert:
image

jgoppert:
image

Anduh:
wing-it will probably tell you to consider integrating these changes to the existing sheet, as Roll20 has the policy to only have one sheet/system, except for the biggest few. If it is easy to do is a whole another matter.

jgoppert:
image

jgoppert:
@Anduh Thanks for the heads up. I'll keep working to make it functional and then see how best to merge it in with the existing sheets. There are a lot of radical changes here, so not much similarity to existing sheets.

wing-it:
To expand a little on what Anduh said, yes we do prefer to keep it to one sheet per system. If it is at all possible to submit this as an upgrade to the existing sheet, we'd appreciate that.
If that is what you'd like to do, you can go ahead and close this pull request, and re-submit it when you're ready to replace the old one.
Thanks!

jgoppert:
@wing-it OK, will do. I'll close and reopen as an overwrite on the existing sheet. I'll try to pull in the cool parts of the other sheet as well.

jgoppert:
@wing-it If you don't mind, going to leave this up until I finish my remaining tasks, then try to merge it in and open a new PR for that then, just easier for me to stay organized that way and review changes with other authors.

jgoppert:
@wing-it @Anduh  this is now setup to replace CWOD-V20, rebased as 1 commit. Still need to finish tasks above.

jgoppert:
image

Need to tie in calcs for discp/ background but layout is done.

image

jgoppert:
@Anduh @wing-it One argument for making this a new sheet and leaving the old one is this:

The old sheet just records strength, so there is no way to know how you got there xp wise, which is why I have the new input fields. So I can't take the old attribute data and upgrade to the new sheet and calculate the xp of the character since I don't have the data.

For instance, strength 5 could mean (1 base point + 2 freebie points + 3 xp points) or (2 base point + 0 freebie points + 3 xp points) which would both have different impacts on the xp for the character.

jgoppert:
@alanman @Smil3yRiley @JackNexus @Twibbit @mantonyz thoughts?

Anduh:
I'll start with I do not myself use this, so consider more users opinions on this if possible.

Would one way be to use tabs? first tab have the regular view, and the second would be this. Then you could use the "base" variable shared between the tabs, so the old view would have all as it used to, and anyone who want to switch to the new way could divide the points however it's supposed to be done.

jgoppert:
@Anduh Yeah, at start I need to cycle through and make sure strength = base + free + xp, and add to base if it doesn't I guess. Is it best to do that on sheet load? On start I get a lot of warnings about undefined attributes on sheet load, so I'm not sure how to wait until it connects to the sheet properly and fixes the undefined values into zeros.

jgoppert:
image

Current status.

jgoppert:
image
Current status.

Anduh:
I don't know how it would be best to resolve that, sorry. I no experience with sheet workers. But put up a thread on the roll20 forums and people will have suggestions, maybe even someone using the sheet.

joesinghaus:
There's no way to delete attributes via sheet workers, unfortunately. Setting them to the empty string should reset them to the default, but that may interact weirdly with autocalc fields (do not use autocalc fields, ever). A good solution would be to do this instead:

1. Let strength (and whatever other fields this apply to) be calculated by sheet workers from strength_base etc instead. This will also improve your sheet's performance.
2. On sheet opening, if the **version** attribute is undefined, perform a conversion function that takes the old strength if present and puts it into the new strength_base. This conversion function should probably be set up in such a way that it does nothing if run on an empty sheet.
3. Also on sheet opening, set a version attribute (not defined in the HTML, only in sheet workers) that tells you what version of the sheet a particular character has been last opened with. If undefined, you know it's either a new sheet or an old sheet that needs to be converted, which tells you to run the conversion function in 2. If not undefined, you're good (or you can run some other upgrade stuff if you ever change attribute names between sheet versions).

jgoppert:
Thanks @joesinghaus  I'll go ahead and finish removing all of the autocalcs then.

jgoppert:
image
Working on a new layout to support dots and table style @Anduh.

jgoppert:
Closing this for now until we get something more tested by our group.

PR #3423, opened 2018-02-03
Bugfix Eclipse Phase V2 & Bugfix Knight NPC

Zakarik:


wing-it:
I noticed the Knight sheet has changed a few attribute names (for example `attr_listeAO` to `attr_listeAOPNJ`). I'm unfamiliar with this system, could you explain how losing the data stored in the old attributes might affect ongoing games? If this would be a significant inconvenience, you might consider adding a sheet worker to update the old attributes to the new ones.

Zakarik:
It's not gonna affect the current games. Actually, I caused a bug by assigning similar names to two fields. The field mentioned, for example, appears in the PC version and the NPC version of the form, I forgot to change the name on the NPC version. But this field is self-filled and hidden in plain sight. It only serves me to concatenate several fields that will be included in the die roll, and to better find myself there, rather than having the value of the roll that makes a line of an endless length and difficult to read. So the only effect is that it will prevent the die roll on the NPC version from taking, by mistake, the fields of the PC sheet.

Translated with www.DeepL.com/Translator

PR #3438, opened 2018-02-07
Elite Dangerous Role Playing Game - Added a missing checkbox next to each skill

tacroy00:
In this system at the end of an adventure you are allowed to improve skills which you have used during the adventure. A checkbox next to the skills is used to keep track of this on the character sheet. I've replicated this functionality with these checkboxes.


tacroy00:
Fixing another bug then submitting a new pull request

PR #3458, opened 2018-02-12
Update translation.json

rdefendi:
Fixed the label on Fitness, to include (Defense) for campaigns where you break fitness out by subskills.


rdefendi:
I forgot to save a change I made to the translations file, adding a subskills note to the fitness skill.

rdefendi:
Out of curiosity, how long before the change shows up?

Solarswordsman:
Hi rdefendi,
It will not ever appear for English viewers (or any other language that has translated that key), because the key already existed before you changed it.  You will need to update `en.json` in the `translations/` folder as well in order to change the key's translation.

Jeff

PR #3463, opened 2018-02-13
Bugfix barabarians for Knight & fix Eclipse Phase V2 Translate & fix Capharnaum for Google Chrome

Zakarik:


wing-it:
I noticed you're removing a large section of the sheet, including several attributes. I'm not familiar with this system, is the loss of this data going to affect ongoing games?

Zakarik:
Yeah, it was a mistake on the sheet.

I had put some skills of the V1, I didn't realize that they had disappeared... An inattention error. So I removed the offending parts.

PR #3472, opened 2018-02-17
Starfinder Hud - Bug Fixes, community use disclaimer

Kurohyou:
Bug fixes:
* Importer will now auto skip tactics section
* npc rolls that were not working due to misformatted private roll attribute

The Paizo community use disclaimer has been added as well.


wing-it:
Your sheet is now live on the site, but I thought I'd let you know that it doesn't seem to be working right, I can't get the Core sheet to show up, it's stuck on the settings tab. If you can figure out the problem and submit a fix, we may be able to push out a fix outside of our normal update cycle.

Kurohyou:
Thanks for the update Wing-it. I've submitted a possible fix. I think this was because the dropdown sheets are limited to ~13000 lines of html/js code. I've removed a bunch of line breaks in the html to reduce the lines of code. See pull request #3482 .

Kurohyou:
Hey wing-it, don't know if this will reach you better than the github
comment (or may just be another github comment). Just wanted to notify you
there is a fix in as a pull request.

PR #3481, opened 2018-02-20
5th Edition OGL French translation

MrPadRpg:
Some fix in the french translation of the 5th edition OGL


wing-it:
Looks like this request contains an unrelated file that was accidentally deleted, could you add that back?

MrPadRpg:
I want to cancel this pull request, something is wrong in my branch

PR #3484, opened 2018-02-20
Add files via upload

suntzu777:
Bug - found 2 attributes with the same name , one renamed


suntzu777:
don't know why i haven't got write access to my own fork of the depository so not sure how to fix it

PR #3504, opened 2018-02-26
SWN Revised v1.6.1

Karlinator:
Fixed the Default Settings (thanks for making me test that on production servers btw), a couple of text changes, and a small improvement tot he NPC tab.


wing-it:
@ me on github if this doesn't work. I realize that it can be frustrating to try to nail this down, I may be able to help test any changes locally if it's still not working after this.

PR #3513, opened 2018-03-01
New Sheet Synthicide by Will Power Games

blakeburns2194:
Hello,

I tried submitting this new character sheet once before but I think I put it in the wrong place. Hopefully this is the right place. It should include everything.

Thank you!

[Synthicide by Will Power Games.zip](https://github.com/Roll20/roll20-character-sheets/files/1773091/Synthicide.by.Will.Power.Games.zip)


Anduh:
This won't be approved, trust me. You are supposed to only make changes(or add) the  files relevant to you sheet, but you somehow managed to try re-add what seems to be all existing sheets.

Fo back and look at the Github guide on the wiki and for example https://github.com/Roll20/roll20-character-sheets/pull/3511 which is a better example of submitting a new sheet.

wing-it:
To clarify a little on the best way to fix this, I'd suggest the easiest (though not the only) way would be to delete your local branch, pull down a new one, then add your files back and re-submit this request. That way your branch should be completely up to date so it should only include your new files. I'll close this now, let me know if you need any help.

Issue #3421, opened 2018-02-02
[Open Legend] Character sheet doesn't allow for invisible GM rolls


seppe2468:
There is currently no option to use the roll macros on character sheets and make them invisible to players.

BuildAndDie:
1st) Which Open Legend sheet do you mean, though I don't think either one does this (nor a lot of charactersheets)

2) there isn't a lot of need for it in most cases, though for the sheet I have considered it, but it isn't a priority right now

3) This isn't really an issue, but a feature request, see: https://community.openlegendrpg.com/t/roll20-character-sheet/162

seppe2468:
1) The charactersheet on Roll20 simply named "Open Legend", we're not using the minimalist one.
2) While it's perfectly possible to GM the game without it, it would make fudging the numbers as the GM a lot easier.
3) My apologies, it's my first time using these channels.

BuildAndDie:
For #3, no worries, just harder for a person that works on sheets to see issues posted here unless the author is mentioned/pinged, hence why it took me 25 days to notice it.

For #2, not a personal fan of fudging numbers, BUT, I understand some people want to. I was planning to implement a way to send various rolls to the GM, though not for fudging, for players that want to do something without the party knowing about it. It of course can be used by the GM in a similar fashion.

I'll actually be working on the sheet some over the next 2 weeks, though I will probably be working on other things first, so not sure If I'll get something implemented for GM rolls this stretch, but possibly after I get back from my next trip.

Again, feel free to post in the forum topic I linked above, I take lots of requests there, and it will help remind me.

PR #3521, opened 2018-03-04
[Open Legend] Fixes to Speed & New Char Calcs

BuildAndDie:
### 1.8.9.98 on 2018 March 4th
* Fixed a new character sheet not auto calculating Feat & Attribute Point Totals
* Fixed Speed changes based on Max Heavy amount
	- Based on Might Score
* Fixed Speed not calculating when you entered amounts from Feats
* Changed GUI for Actions, Inventory, etc to be Gray lines instead of Teal


BuildAndDie:
Close #3421 
not an actually issue, but suggestion, plan to implement in future update

PR #3536, opened 2018-03-08
Blades in the Dark v3.0

joesinghaus:


joesinghaus:
@wing-it Something weird is happening here that I cannot explain. The very first rule in my CSS (.charsheet{/*STUFF*/}) is not being applied on live Roll20. But I cannot see why; when I copy the exact same code that is on here into a custom sheet, it works!
The difference is easy to see, since all the fonts become sans-serif fonts when that first CSS rule is missing.
Apart from that, everything seems to work, so I really don't know what is going on. Any ideas?

wing-it:
interesting. I'm not sure what would cause that. I'll investigate a bit when I get a chance and see if I can narrow down the problem.

joesinghaus:
Thanks! It's not so catastrophic, since it just looks a bit worse, but it would still be great if it could be fixed once you get a chance.

joesinghaus:
If you think it will help, I can just submit a non-minified version of the CSS for next week for now. Don't know if that makes a difference, but since it is impossible for me to test anyway, maybe that's the best choice.

PR #3542, opened 2018-03-10
Update to Mutant Year Zero Alternative

nodos-7:
Added a roll template and uses default stats. This cleans up a lot of the hassle of playing MYZ online, since tracking which dice get which results is important in the system.

Added missing fields and added new logic to shift the titles out to match other expansions of MYZ.

### DIFF
--- a/Mutant Year Zero Alternative/MutantYearZeroAlternative.css
+++ b/Mutant Year Zero Alternative/MutantYearZeroAlternative.css
@@ -10,6 +10,10 @@ div.sheet-tab-content {
    margin-top:2px;
 }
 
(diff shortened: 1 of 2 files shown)


wing-it:
Before I merge this, I just wanted to confirm that you've kept all of the attribute names the same, as these changes are a little too extensive for me to check manually.
If any of the attribute names have been changed, players in existing games will lose the associated data. I just want to make sure this update isn't going to interrupt anyone's game in progress.

nodos-7:
I've actually checked it on one of my players games, i didn't change any attributes, just added new ones for labels

wing-it:
Ok, thanks!

PR #3547, opened 2018-03-12
DG Sheet v1.5

uriele:
Fix the issue with CONX5 roll
Fix the issue with the strength modifier applied also to firearms
Add double damage for critical attack rolls
Add Lethality


wing-it:
This branch seems to be deleting all of the old files for this system and uploading a new html file in a different location. Instead, you should commit your changes as modifications to the existing files and leave them in place.
If you need help with how to do this, let me know and I may be able to walk you through it.
I'm going to go ahead and close this request for now, please resubmit after you've made these changes so I can review.

PR #3550, opened 2018-03-12
Item modifiers skills minor fix

oukag:
Previously item modifiers for animal handling and sleight of hand did not properly apply skill bonuses. The expected string to trigger the changed used spaces whereas determining the value of the bonus was looking for the name of the skill to have underscores. This meant that it was impossible to use the item modifiers for this field.

This change will require the underscores between words for these skills so that the correct bonus can be applied.


SteveKoontz:
Hey Oukag, thanks for working on this. Unfortunately your change doesn't actually solve the issue. Doing an underscore in the mods section of an item was always working. What wasn't working was using skills with a space and not an underscore in them in that section. Your change doesn't change that problem because the breakdown was happening in the update_skills function.

I found the source of the issue and I've patched it. It'll go out next week.

oukag:
Great job! I tried testing it on my own, but was unable to get any of the sheet workers to work actually. I'm glad that you were able to use the information I provided and find the true cause of the issue.

PR #3560, opened 2018-03-14
TFT Sheet.html

Jhar226:
Complete change of the entire sheet, updated for 2e.

### DIFF
--- a/The-Fade-Tabletop/TFT Sheet.html
+++ b/The-Fade-Tabletop/TFT Sheet.html
@@ -1,163 +1,182 @@
 <div class="sheet-1colrow" style="text-align:center;">
   <div class="sheet-col">
       <p style="text-align: center;">
-            <img src="http://s10.postimg.org/o5frg51cp/BFS_Logo.png" style="max-height: 100px;" alt="The Fade Tabletop"/>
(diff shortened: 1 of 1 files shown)


wing-it:
Could you give me some details about this sheet? The main thing I'm concerned about is compatibility with ongoing games. Would existing characters lose data when upgrading to this sheet? Have you tried upgrading an existing game to make sure no data is lost?
Secondly, the comment states that this is updating the sheet for 2e. I'm not familiar with the system, are the differences between editions enough that some people may want to keep using the 1e rules, or will they remain completely compatible with each other?

Jhar226:
It would cause 1e games to lose data and for the sheets to not function, the two editions while compatible are two different die systsms.

wing-it:
In that case, we have 2 options.

1: Have an option to toggle between the two editions, so that this sheet supports both. Doing this in a way that gives current users the option to continue without interruption or to upgrade to the new edition is probably do-able with sheet workers, but it may be a lot of work.

2: Re-submit this as a separate character sheet. That way, the current 1e games can continue uninterrupted, but those who want to use 2e will have that option.

Either way, we want to make sure that people currently using this sheet don't have any changes forced on them that they do not want.

wing-it:
I'm going to go ahead and close this pull request. If you make the changes mentioned above, you can simply re-submit this pull request or submit a new one.

PR #3562, opened 2018-03-14
Blades in the Dark: fix buggy roll template & CSS workaround

joesinghaus:
- Fix for missing opening curly bracket, rendering the second roll template nonfunctional
- Added a workaround for the font part of the weird "first CSS selector does not work" problem. This cannot be used to fix the .charsheet padding, but at least it solves everything else.

PR #3568, opened 2018-03-15
Revert "Blades in the Dark v3.0"

wing-it:
Reverts Roll20/roll20-character-sheets#3536


wing-it:
@joesinghaus We reverted this because of numerous reports that people's games were breaking: https://app.roll20.net/forum/post/6181674/images-or-game-not-loading/?pageforid=6182355#post-6182355
Hopefully you can sort out the problem and have a fix ready for next week.

joesinghaus:
Wait, so I can ask you to revert things? Cool, thanks for doing this! I did indeed break one of the roll templates with this change.

Anyway, yes #3562 would have fixed it (and partially fixed the weird "CSS does not apply" issue by moving some CSS around so it has less of an impact, though I still cannot reproduce that one).

PR #3574, opened 2018-03-16
New Deadlands Classic Sheet

pyromanci:
New Deadlands Classic Sheet.

This sheet is a improved version. Cleaner layout and roll buttons.


pyromanci:
I can, Though while I tired to make it swap friendly there are some fields that just are not compatible between the 2.

PR #3581, opened 2018-03-20
Deadlands Classic Redesign

pyromanci:
The original sheet was designed by [Kevin Justus (Rats)](https://app.roll20.net/users/129882/rats). The sheet was upgraded to include rolling, skill modifiers, multi damge die per weapon, and an improved layout.

Prequest https://github.com/Roll20/roll20-character-sheets/pull/3574 submitting changes as a ungrade to the existing sheet. Orginal sheet design credited.


pyromanci:
The Fields that will not migrate cleanly are Fate Chips, Ammo, and Weapons.

**Fate Chips**
Were changed to be by the rules of the system. The originally sheet had it so you could only get 5 of each type. This is wrong the system allows for up to 10 whites, 5 red, 4 blue, and 2 Legend. As a result I lowered the max on Red, Blue, and Legend chips and increased the max on white. The impact on the game should be extremely minimal. Most players will have spent their chips for EXP at the end of the session.

**Ammo**
Ammo on the sheet, was not very friendly and extremely limited and restive. The sheet only allowed you to track 3 ammo's. With 2 fields being a restive 0-6, and 1 field 0-15.  The Ammo fields on the official printed character sheets could be used a couple different ways. Tracking how much ammo you had in total (via boxes) or how much ammo was in a gun. 

While this method is fine for pen and paper, it's restive in a digital system. Which forces you to track those sorts of things in different places. So the ammo section was converted to repeating field with custom names, unrestricted capacity and notes. This allows the players to track their ammo in any way they wish. You wouldn't be able to track a belt feed Gatling gun with the current sheet.

This should have a minimal affect on existing games.

**Weapons**
Originally the fields were just weapon#name, weapon#shots, weapon#rof, weapon#range, weapon#damage. weapon#def, and weapon#speed. Where # was 1-8 (1-4 being ranged, and 5-8 being melee). Though not all of those fields were available for each weapon. So creating macro's to roll things becomes confusing. As well the damage for most weapons were variable. So the damage field was mostly just a description.

So to make macro's easier to create and allow for the weapons to be expanded. As 4 for each category and 1 forced slot (giving you 7 total weapons) were not enough. So i broke the fields down into their seprate groups. Range and Melee, and converted the damage field into a repeat field.
rngweaponname#, rngweaponshots#, rngweaponrof#, rngweaponrange#, rngweapondamage# (repeating field),  where # is 1-6.
meleeweaponname#, meleeweapondefense#, meleeweaponspeed#, meleeweapondamage# (repeating field),  where # is 1-6.

This will be the largest impact on a game. As the players will have to reenter their weapons. Though their macro's should still work. As i did not reuse the old field names. I could possibly put a read only reference to the old weapon values under the misc column.

wing-it:
From your description, it sounds like it may be possible to use a sheet worker to import the old weapons into the new repeating sections. If you aren't familiar with sheet workers, I may be able to find an example to point you toward.
However, if it is simply not possible to convert the old data this way, I would definitely recommend putting all of the data that won't carry over in a new column as you suggest, so players will at least have it to refer to when converting to the new system.
Preferably, there would be a warning dialog at the top of the sheet that will inform users of the change and instruct them how to convert their data to the new system. You can make this dialog only show up the first time the sheet is opened by having them click a checkbox to dismiss it. You can also plan to completely remove the old data from the sheet after everyone has had plenty of time to convert (at least a month is a good rule of thumb), but be sure to inform users when it will be removed in that warning dialog.

pyromanci:
If you have a sheet work explain I can take a look. I've not worked with the sheet workers before beyond minor tinkering. 

Meanwhile i'll just add a new tab and warning.

pyromanci:
I did just add in a new tab, and notices for the older fields.

PR #3582, opened 2018-03-20
Pathfinder-Neceros/translations/de.json

PhilHoffm:
solve little syntax errors. Translation about 51%.


PhilHoffm:
there is a new update. next update will upload in the next days.

vince-roll20:
There are reports that users that have their roll20 language set to German are unable to load the sheet.

PR #3596, opened 2018-03-25
Initial commit for SW30

clflo:


Anduh:
Nice! This looks neat and minimalistic, just as it should be.

I had considered reworking the existing [Star Wars D6 sheet](https://github.com/Roll20/roll20-character-sheets/tree/master/D6StarWars) (more geared to 2E) to be more compatible with 1E/30th Anniversary Edition, but haven't come around to it yet. It's better this way, with having a sheet each for the 1E/30thAE and 2E.

Are you considering adding a section for ships?

clflo:
Thanks, I’m quite happy with it. I want to add some more nifty features but haven’t had the time to do it yet. Yes, I’m planning to add a tab with ship section.

Anduh:
Those sheetworker seems nice, am I reading them right that they also automatically gives the die penalties if you are wounded? And does the Armor code automatically reduce the die code in dexterity?

There you aware of the existing sheet?

Anyway, the SWD6 community will surely rejoice with this great addition, I can drop a message or two about this when it's published.

clflo:
I realised now that I committed some unfinished code. I'm a bit unfocused with a 3 month old baby at home...

The roll button for armor is only rolling strength plus armor code. I'm planning to update the code so dexterity skill rolls will get the armor penalty automatically. Also, I want to add som more worker scripts that fills in empty skills when you change the attribute value. And, of course a ship section.

I wanted to have an initial commit so it could be reviewed and maybe merged as a first draft.

Anduh:
Isn't armor soak roll only armor+strength, what's missing in your opinion, is it the possible wound-penalty?

I know it probably nothing, but maybe 1d6 is a smarter default value for attribute/skills, regardless of the sheetworkers works perfectly or not.

PR #3615, opened 2018-04-01
[Pokemon Tabletop Adventures] Bug fixes

AdrienGAIRE:
Minor bug fixes & changes:
 - Changed the name of "attr_pokemon_HPXmax" to "attr_pokemon_HPX_max" for consistency
 - Changed the name of "attr_Pokemon_HPX" to "attr_pokemon_HPX" for consistency
 - Corrected a bug where nature-related stat where not kept inbetween sheet openings
 - Corrected the roll for damage to include the specificated damage bonus (movedmgbonusXX)
 - Added the "N/A" option in the damage move stat, for non-damaging moves

### DIFF
--- a/PokemonTabletopAdventures/PTA.html
+++ b/PokemonTabletopAdventures/PTA.html
@@ -247,7 +247,7 @@
     <div class="sheet-col">
         <h3 style="margin-bottom: 10px;">HP</h3>
         
-        <label>Current</label><input type="number" name='attr_Pokemon_HP1' class='sheet-short'><br />
+        <label>Current</label><input type="number" name='attr_pokemon_HP1' class='sheet-short'><br />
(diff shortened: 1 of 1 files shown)


Anduh:
I'm sure you're aware that existing games using those stats will "lose" them, and have to be manually re-entered unless you make a sheetworker to migrate the attributes. But it's probably not a problem if there exists only very few players.

Regards, someone who have made the mistake in the past

AdrienGAIRE:
To be honest, this is the first time I try to contribute, so no, this problem kind of flew over my head ^^'
But yeah, my group already had to remake part of their sheets with the last update...

Should I first only do the fixes? (i.e. the last three points)

For the max HP, it shouldn't be a problem as this stat is automatically calculated, but for the current HP, this can be an inconvenience...

Anduh:
Well, I'd suggest to re-submit the pull request without the attribute-namechange until you add a sheet-worker that does the attribute transfer. It' should be relatively simple but I haven't done it myself nor do I really know or understand them yet. And there is another update by someone else, it might be smart to coordinate you efforts with the sheet author.

AdrienGAIRE:
Well, let's do that!

PR #3626, opened 2018-04-03
[Pavillon Noir 2]  Add character sheet & files

Sheershym:


Anduh:
The sheet.json file seems to be empty, maybe you forgot to fill it?

[This]( https://github.com/Roll20/roll20-character-sheets/blob/master/kitchensink/sheet.json) is an example of a correctly filled sheet.json

PR #3628, opened 2018-04-03
Paranoia - Fix the translation

MrPadRpg:
Error in the EN translation file


Anduh:
You got some conflicts here, won't be approved until you fix them, I can almost guarantee it.

Are you doing translation corrections for multiple sheets? It would be smarter to make them separate pull requests so if one have some problems, the others won't be hindered if it doesn't get approved.

You might want to update your repository against Roll20's, this weeks pull-requests were just merged, would be the source of the conflict.

PR #3630, opened 2018-04-03
[Pokemon Tabletop Adventure] Fixes & Changes

AdrienGAIRE:
First of all, I would like to apologize for the mess my commits are... I had some problems with Github Desktop, it won't happen again!
I solemnly swear my next contributions will be cleaner!

Anyway, here are some fixes and changes to the Pokemon Tabletop Adventure character sheet:
 - Corrected a bug where the selected nature-related stat changes were not kept in-between sheet openings
 - Corrected spelling of "Capabililties" into "Capabilities"
 - Fixed the damage roll
 - Added a "Misc. Skill Bonus" to the skill checks
 - Added the "N/A" option in the damage move stat option, for non-damaging moves


AdrienGAIRE:
@wing-it,
sorry if I'm being pushy, I know you're just back from your retreat, but are you planning to do the merge request this week? Without the damage roll fix, we can't roll and I'd really like it if it could be fixed before my session tomorrow ^^'

Thank you in advance!

wing-it:
since this sounds like a vital fix, i'll try to get it pushed out today. I can't make any promises, though!

AdrienGAIRE:
Thank you!

PR #3637, opened 2018-04-06
5e OGL fixed spellcasting lvls of non-full casters

oukag:
Half casters (Paladins and Rangers) and third casters (Arcane Tricksters and Eldritch Knights) will now calculate the correct number of spellslots. Previously the spell slots were rounding up to the next level of spellcasting when it should instead be rounded down.


joesinghaus:
Actually, if you look at the book, they **do** round up, not down.

oukag:
Ah you're right, I was looking at multiclassing with these classes, which do round down. I'll modify the change so that it rounds up when not multiclassing but rounds down when multiclassing.

PR #3638, opened 2018-04-07
Over the Edge sheet typo fix and Shadows Over Sol sheet update

tmtabor:
Fixed typo in the Important Person attribute of the Over the Edge 2e sheet. Also updated the Shadows Over Sol character sheet for compatibility with the Siren's Call supplement.

### DIFF
--- a/Over the Edge 2e/Over_The_Edge.html
+++ b/Over the Edge 2e/Over_The_Edge.html
@@ -58,7 +58,7 @@
 		<br />
 		<div class="sheet-field_box">
 			<p  class="sheet-field_label sheet-field_label_r">Important Person</p>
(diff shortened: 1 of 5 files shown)


wing-it:
These are some pretty extensive changes to the Shadows over Sol sheet. Will this upgrade affect users currently using the sheets? I'm mostly concerned about data being lost due to attribute names being changed.
If any attribute names have been changed, you can use sheet workers to update the old sheet to the new version.

tmtabor:
No attributes were changed on the Shadows Over Sol sheet. Most of the big changes you see are to the structure of <div> tags to make the sheet flow better when it is resized. The way the sheet previous used CSS floats made it look weird when you resized the sheet to make it more narrow.

PR #3655, opened 2018-04-13
Small fixes + Small changes

Medusa:
TIH sheet. Small fixes & changes


wing-it:
Before I merge this, I thought I would point out the misspelling in `att_positive-detail`, which should be `attr_positive-detail`.

Medusa:
You're right, thank you.

PR #3656, opened 2018-04-14
Donjon de Naheulbeuk 2.1

Kaias26:
- minor bug fixes
- minor UI improvement


wing-it:
Good job fixing the merge error. I'm going to go ahead and merge this, but here is something to try to fix for next time:
Your code was using the default Windows new-line characters (\r\n) when we need it to use the default Unix character (\n). Our website runs on a Unix system, and requires the Unix line-endings rather than windows. To avoid this in the future, you can change the settings in your text editor to use Unix line-endings rather than Windows.

PR #3669, opened 2018-04-19
Update WOIN.HTML to fix two typos, add new fields

Newguy12354:
This change would be to Fix the typo that when you roll Logic, it states you roll Agility. It will now say you roll Logic.
Reputation was written repuation everywhere. Made it Reputation again.
The sheet needed a field to add additional damage to a weapon roll, for example, when you want to roll 2d6+2. There was no way to add the +2. I added a field called Damage Add. for that.
The sheet needed a field to record current health and max health. Left the Health field as is and added a max Health field.


wing-it:
Unfortunately, the places where "reputation" was misspelled were all names of attributes, and changing them will cause players to lose data.
If you want to change the name, you'll need to add a sheet worker to update the old data to the new attribute names.

Newguy12354:
Hmmm. Instead, I'll update only the text Value of Reputation so that it appears properly, but it will remain named repuation for coding purposes. This should fix the issue without changing the attributes. I'll update and test later tonight.

Newguy12354:
Okay, my second Commit should only be:
Text change so that it properly says you are rolling Logic when you click Logic (instead of Agility, like it did before)
Text change so the coded attribute "repuation" is shown, in text, as "Reputation", but no changes to the attribute name has been done, so no need for a sheet worker.
The new field for additional damage under equipment, for example, when you need to a 2d6+2, to be able to have the +2 in the damage roll.

Newguy12354:
This has also been tested using a custom sheet in Roll20 to verify the code. I didn't find any broken area.

Newguy12354:
Oh, and of course, the new field for Max health.

PR #3694, opened 2018-04-25
Starfinder HUD compendium, hp, init fixes

Kurohyou:
# Bug Fixes
- NPC initiative not rolling when set to 0
- Internal Compendium not importing
- temp hp max being set to 0 would not properly set temp hp max to 0
- temp hp max being updated did not properly set temp hp

# Feature Updates
- Temp HP will now carry over to stamina

@wing-it, Any chance for an out of sequence merge on these fixes? They're two things that break for people.


Kurohyou:
@wing-it  The initiative bug is actually something that prevents the sheet from being useful in the common event of combat breaking out.

PR #3710, opened 2018-04-30
new style char sheet

Hotbloodhypno:
taken from the original Scion sheet..

fixed errors and made it easier to use for combat...

and a bit more colour!


Solarswordsman:
This is a significant rewrite of the original sheet.  Will existing users of the sheet have their user experience significantly degraded in some way?  What's most important is whether attribute names have been lost or changed and if so that the prior attributes are migrated gracefully through something like a sheet worker.

Hotbloodhypno:
the transfer to or from this sheet to the original should be smooth... no major change in name of attribute titles or layout

Solarswordsman:
OK, I can merge in the Scion changes, but you've made significant changes to the CWOD V:tM 20th anniversary sheet, some of which are breaking changes and arguably make the sheet harder to use (whatever symbol you're using for Abilities is not working, I am greeted with this: � everywhere).

If you remove the CWOD stuff from this branch then I'll merge it in.

Hotbloodhypno:
on vampire 20th..
any roll that uses Strength, Dexterity, or Stamina..  is set up to auto add any blood enhanced stat or Potence Celerity or Fortitude.. look in the combat section the sheet.. its controlled there..(when characters are most likely to burn Blood to increase stats..)

PR #3711, opened 2018-04-30
created for Scion 2 (Kickstarter)

Hotbloodhypno:
this was made to playtest the game.. 

tho there ATM no official char sheets drawn up ... so it might be totally wrong in places (and some sections have altered names)


Solarswordsman:
Hi @Hotbloodhypno, because this is a separate sheet, its files need to be placed in a separate directory than Scion/, which already has the first sheet.  Scion2/ would do just fine.

Also, since there is no publicly available form of the sheet, I can accept the code into the repository for use in playtesting (and for tracking, then release down the line), but you will need to use a pro account to copy the code and put it in your game as a Custom sheet; we don't support adding sheets to the dropdown list unless they are publicly available in some form.

Hotbloodhypno:
ok.. thanks for the info...

i will resend the scion 1 alternate sheet...  in its own folder..

as to the actual attribute names...  nothing changed .. it should be  (mostly) be exactly the same...   except where.. the option menus at the top (choose pantheon.. and god!) use a script/writer.. to infuence a few lines on the pages  i.e.  you choose god/pantheon  it auto sorts your Virtues, and the pantheon Purview ...


was trying to save time and page turning for players



Solarswordsman:
You should put the Scion 2 sheet (this one) in its own folder, leaving the first one as an update so long as those changes don't totally drop the attributes as you've said.  The first one doesn't need those separate `_1E` files either.  Since one of those is a CSS file and it's not referenced in the sheet.json, I'm going to assume it's needed so I'll hold off on merging both of these until those issues are resolved.

Solarswordsman:
You still need to move this version to a new directory before I can merge this, as it is a separate sheet for a new version of the game.

Hotbloodhypno:
have been trying to create a folder for it and post it up to my on line github..  but cant seem to get it to co-operate

its sitting on my system in its own folder.

PR #3713, opened 2018-04-30
Command and Medicine fixes

Hotbloodhypno:
sorted the errors in the code for the two buttons  Command and Medicine


Solarswordsman:
Like I mentioned in the other thread, simply having the additional html and CSS files aren't going to do anything here.  The fix for Command and Medicine is fine, but it should be in its own PR by itself.  This should be easy enough by reverting 4b70b28 or some other method.

We don't generally want multiple sheets for the same game, so you should be treating the new sheet as an upgrade (though it is a significant one) to the former one.

Hotbloodhypno:
i am finding it a bit difficult  using the github app on my computer to push the required files (and folders) to my online copy.. then to request a pull request to merge or create the files for use in the master files ...  trying to create a folder in the online copy is a task...


thanks for informing me whats happening tho... (wrong or right)
i know you have a difficult job sorting thro all the requests ...

Solarswordsman:
It's been a while, so I'm going to close this PR.  Let me know if you have any questions on how to do what I specified above.

Thanks!
Jeff

PR #3716, opened 2018-04-30
Add Sheet "Les Royaumes de Midgard" and "Tiny"

Lyraelle:
Les Royaumes de Midgard
http://www.legrog.org/jeux/royaumes-de-midgard

Tiny
http://www.legrog.org/informations/actualites/11-06-2017-tiny

Thanks you !


Solarswordsman:
In the future, it's preferable to have two separate Pull Requests here, because you're adding two new sheets entirely.  I'll go ahead and merge this though, as it is all formatted properly; it just helps with record-keeping.

PR #3723, opened 2018-05-02
modified conan 2d20 sheet without tickmarks

rschweiker:


Anduh:
You submission will probably not be approved as the [guidelines](https://github.com/Roll20/roll20-character-sheets#guidelines) says:

> Don't Duplicate Work

We only need one sheet for each system. For some major systems like Pathfinder or D&D, we may allow more than one sheet if they are different enough, but contact us in advance before you spend a lot of time working on a new sheet for a game we already have a sheet for. We should focus on improving existing sheets, not creating totally new ones. Along those lines, this is a community effort, so if you want to help improve an existing sheet, just jump in and do so.

Is there any reason this change can't be approved to the original sheet, or made a customization option on it? Have you contacted the original sheet author?

rschweiker:
What if sheet  creator disagrees with suggested edits?

Solarswordsman:
@rschweiker: as mentioned above by @Anduh, we only want one sheet per system unless there are exceptional circumstances.  I will note that we hold accessibility in very high regard, so the suggestions above (making this accessibility mode an improvement to the sheet as a toggle-able option, so that the original style is preserved for those who wish to continue using it) would be my own as well.

This is a community repository, the creator does not own the sheet, and I agree that they would likely have no issue with your improvements anyway so long as they are non-breaking for existing users.

Thank you for your hard work, and if you need help getting this done, please don't hesitate to ask.

(P.S. make sure if you end up changing the `sheet.json` for the main sheet that you format it correctly, the one here doesn't have quotes around the values so it's not valid JSON)

Anduh:
Did it turn out too hard to integrate these improvements as optional in the existing sheet?

You can stop by the char sheet forum if you got stuck with something.

PR #3728, opened 2018-05-06
Shadowrun Anarchy sheet

clevett:
New Shadowrun Anarchy sheet that holds closer to the style in the book. It has built in rolling & an NPC sheet.


Solarswordsman:
I'm adding this as the new Shadowrun Anarchy sheet and deprecating the previous one.  Current users of the previous sheet can continue to use it so long as they don't change their character sheet.

PR #3732, opened 2018-05-07
[Open Legend] Actions Section Major Update

BuildAndDie:
### 1.9.0 on 2018 May 7th
* Massive update to Favored & Other Actions sections
	- Select if action is an Attack (Damaging or Bane), Boon, or Effect
	- Effect isn't quite fully functional yet in that there is a pop-up that doesn't do anything, but will still roll the dice you choose
* Backgrounds for the borders updated
	- Also white background for things sent to chat
* Prep work for NPC/Alt Form/Companion tabs
	- Nothing functioning on the user side yet

### DIFF
--- a/Open Legend Basic by Great Moustache/ReadMe.md
+++ b/Open Legend Basic by Great Moustache/ReadMe.md
@@ -8,6 +8,7 @@
 ### 1.9.2 on 2018 May 8th
 * Added Whisper to GM option
 	- Global button to make any roll on the sheet go to the GM only
+	- Public to send to public
 
 ### 1.9.1 on 2018 May 8th
 * Made Background Tab for the sheet


BuildAndDie:
Dang it, that last one "Roll Public or GM" I think is the same readme file in "GM roll option" just need to delete that I guess? Everything else should be fine. Was adding it to mark the issue fixed.

joesinghaus:
You can probably just use the "resolve conflicts" option of Github to make this work.

Solarswordsman:
I was so very close to fixing this for you. :P.  The problem is that you put the readme file in the root of the repository rather than in the character sheet's directory (or I guess, in both places).  `git rm ReadMe.md` in the root of the repo and commit that and submit a new PR and it should work fine.

BuildAndDie:
just resubmitted it about 10 minutes ago :(  dang, wish I had known you were working on it. Darn my impatience.

PR #3733, opened 2018-05-07
Uploading an official City of Mist character sheet - Add files via upload

wmkertz:
I'm uploading a character sheet for City-of-Mist.  This sheet has been commissioned by the game's designer and publisher.


Solarswordsman:
This looks really good.  It overwrites the existing sheet though -- are current users going to be able to continue with their games using this sheet? (e.g. have you provided sheet workers to convert from the former version, and/or maintained attribute names)  

If not, it'd likely be best to submit this sheet as a new sheet (in a different directory) and I will deprecate the old sheet (so existing users can keep using the sheet, but it will no longer be able to be selected).

wmkertz:
Okay, uploaded the files to a new folder.  The new request is here: [https://github.com/Roll20/roll20-character-sheets/pull/3745](https://github.com/Roll20/roll20-character-sheets/pull/3745)

Solarswordsman:
Sounds good, I'll take a look.

PR #3743, opened 2018-05-09
Fixed images

clevett:
Images were broken. Updated the URLS to fix them.


clevett:
I had another fix that needs to go as well.

PR #3750, opened 2018-05-12
Update Multiverse Sheet.css

Jhar226:
Fixed the background to be less painful on the eyes.

### DIFF
--- a/Multiverse/Multiverse Sheet.css
+++ b/Multiverse/Multiverse Sheet.css
@@ -20,7 +20,7 @@
 }
 
 .charsheet {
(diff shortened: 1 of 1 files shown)


Solarswordsman:
I would recommend that you (sooner rather than later) upload this image and any other images you're using to the github repository and reference them there (You'll need to predict the absolute url, which will be something like `https://raw.githubusercontent.com/Roll20/roll20-character-sheets/master/Multiverse/IMAGENAME.PNG`).  This will work for now, but in the event the remote hosted image is moved or deleted, your sheet will have problems.

Jeff

PR #3774, opened 2018-05-21
SWN-Revised: v2.1.0

joesinghaus:


joesinghaus:
@Solarswordsman This version now seems to be live, yet sheet workers aren't doing anything, despite the fact that they work just fine when I copy the code into a test game. Also, I have updated the translations with this, but the individual translations were not updated (for any sheet). Something is wrong, could you look into it?

joesinghaus:
@jacksty @wing-it for good measure.

Solarswordsman:
@joesinghaus I'm looking into it now.  From a cursory glance it looks like it's hitting a snag on MAXIMUM_HIT_POINTS:
`Translation Error: the key [MAXIMUM_HIT_POINTS] is not in the translation object.`
Likely this is killing the processing of the sheet workers and causing them to fail.  Interestingly, MAXIMUM_HIT_POINTS is definitely defined as a valid translation in all the languages.

Perhaps there is a bug in a sheet worker (possibly on our end, that would require a workaround at least for now?) processing the translation key.  I'll look into it deeper.  If you have some insight (e.g. a function that uses this key) it might help.

joesinghaus:
It is *not* defined, since it seems you forgot to update the individual translation files this time around. That's probably the problem then?

Solarswordsman:
You are absolutely right.  I updated the translation files, but I missed the rather important step of pushing it to the repository.  My mistake!  I will get this sorted out ASAP.  Sorry!

And thank you for reporting this!

-Jeff

joesinghaus:
Thanks for responding so quickly and resolving the issue!

Jakob

PR #3777, opened 2018-05-22
update to v1.2

Azurift:
Updated to v1.2


Solarswordsman:
Hey @Azurift,

I've made this change on your branch so that I can merge the changes, but please try to use Unix style line endings for future changes.

Jeff

PR #3793, opened 2018-05-24
French translation of the Masks - a new generation character sheet

FkinGuy:


FkinGuy:
Sorry, I made the same mistake as previously; I will modify and create a pull request of a version with translation tools used.

PR #3796, opened 2018-05-24
Masks a new generation: add translation capability ; add french trans…

FkinGuy:
Masks a new generation: add translation capability ; add french translation


Solarswordsman:
Thanks part 2 for another translation!

I'll add the `translation.json` here that is needed as well (note my comment on the Dungeon World translation).   Also note that that the `zu.json` is actually a metadata file we use for crowdin, so there's no use in touching it.

PR #3803, opened 2018-05-26
[Eclipse Phase V2 / Knight / Terre 2] Bugfix for Knight & Eclipse phase / Add New Sheet Terre 2

Zakarik:
Hi.

A New Sheet again for the RPG Terre 2 (http://www.legrog.org/jeux/terre2).

I've corrected a little bit the translation of Eclipse Phase V2 & little others corrections.

I made a correction for Knight and made it transparent to users.


Solarswordsman:
Hi @Zakarik.  All looks mostly good, but in the future, could you please put changes to sheets or new sheets in as separate pull requests?  It helps greatly in record keeping (and my sanity :P).

A note: you don't need to update individual translation files (`af.json`,`el.json`, etc.) unless you have actual translations for them (though you can REMOVE translation keys when necessary if you so choose and it reflects the loss of those keys on the sheet itself).  The important part is that the key exists in the base `translation.json` file.  Also, the `zu.json` file is metadata used by crowdin, so you never need to edit that.

Thanks!
Jeff

PR #3819, opened 2018-06-04
[Eclipse phase V2] Update for last Playtest

Zakarik:


Solarswordsman:
Due to the conflicts in the translation files, could you please revert all changes to these files?  They will be auto-regenerated when I run the translation script.

If you like, you can try to resolve the conflicts, but I don't imagine it will be easy.

PR #3831, opened 2018-06-10
debug  &  Character Sheet  - ShinobiGami

zeteticl:
New Character Sheet
Shinobigami - Modern Ninja Battle Tabletop RPG from Japan
UI base on PF official character sheets


Solarswordsman:
Hi @zeteticl, because we now have a full translation platform, we've moved away from the standard of having separate sheets per language.  This isn't an issue for your merge, but I will name it language-agnostically and I encourage that you support translation keys for the sheet in the future.

Also, please in the future make a separate pull request for each character sheet you modify.  It makes it easier to organize and differentiate different contributions and sheet progress.

Thanks!
-Jeff

PR #3839, opened 2018-06-14
Undoing the PTGS additions just added

GeneralPith:
The expanded PTGS that was recently added is appreciated, but it's not ready yet. The added rows are edge cases used in only a fraction of games, but they dominate the sheet now and take up way too much space. Until they can be hidden in some way (drop downs perhaps?), I think we should stick to the original PTGS.

### DIFF
--- a/Burning Wheel/Burning Wheel.html
+++ b/Burning Wheel/Burning Wheel.html
@@ -1214,24 +1214,24 @@
                 </tr>
                 <tr>
                     <td style="background-color: black; color: white;">Coordinate</td>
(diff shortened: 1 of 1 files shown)


Solarswordsman:
@LeeRhodes Pinging you for awareness.  I'm going to merge this in and revert the changes, I believe @GeneralPith is correct in their assessment of the situation, if the tracks were optional / toggle-able in some way that would be ideal.

LeeRhodes:
Yep I agree. The method I was thinking of was simply have some buttons that can toggle them when needed.

PR #3841, opened 2018-06-16
Additional Very Simplified Sheet AD&D2e

aquaalex:
A very simplified AD&D 2e PC sheet with NO Tabs / Roll templates / Worker Scripts / Buttons


Anduh:
Roll20 prefers to have only one sheet per system, so don't be supprised if this doesn't get added.

aquaalex:
there is already two sheets for AD&D 2e :) And problem is nobody likes to compromise. SOme people want an extremely simplified sheet, some wants soem extra stuff and otehrs wants templates and worker scripts etc

aquaalex:
Actually I see there are many systems with multiple sheets

Anduh:
Yes, both true, but it doesn't change the current guidelines. Recently only official sheets(made by Roll20) have been added as secondary sheets, and here and there some smaller sheets have been replaced by newer and better sheet. Even a modified sheet with colorblindness support where turned down as a separate sheet as the improvements could be added as options to the main sheet.

Solarswordsman:
@aquaalex As Anduh said, our policy is only to have one sheet per system, that the community can contribute to in order to make it the best possible experience rather than fragment the player base and risk problems like unmaintained sheets, difficulty of user choice of sheet, etc.  While duplicate sheets do exist in the repository, they are only there from before this policy took effect (with some very particular exceptions) in order to retain backwards compatibility for those using said sheets.

I applaud your work in improving this sheet and working with the community to try and make it the best it can be, but I can't accept another sheet for the same game system.  Perhaps you can implement different modes for the sheet which are relatively easy to toggle, thus allowing the sheet to have both a "Simple" and "Advanced" form.  As you progress with the sheet, perhaps even its name in the character sheet select drop-down might be altered to reflect such a change.

If you have any questions, feel free to contact me here, or on ([Roll 20 by direct message](https://app.roll20.net/users/1464775)).

Thanks,
Jeff

aquaalex:
Well I am not for more sheets, but some users dont want any scripts or templates or buttons so as a community member I was trying to accomodate them as I do not see any harm or penalties arising from this. And give none pro members some options. But I have removed my additional sheets. Please make sure to at least approve my other changes to the current sheet.

BaldarSilveraxe:
Isn't "community sheets" something I support with each monthly payment to Roll20? Its says I get that right on my subscription. Seems to be something I am paying for. And ADD is 6.6% on the ORR Group report 2018.

PR #3842, opened 2018-06-18
[Eclipse Phase V2] Update for last playtest

Zakarik:


Solarswordsman:
In the future you should try to merge from upstream (this repository's master) before making changes so that you get all the auto-generated translation files into your fork.  They'll generate again, but it makes the commit history deceptive as they don't need to be removed.

PR #3846, opened 2018-06-19
Initial Deployment

sir-argo:


Anduh:
So it's a replacement for [this sheet](https://github.com/Roll20/roll20-character-sheets/tree/master/Hero%20Games%206e)?

sir-argo:
While that other sheet is for the same game system, I was not intending to replace it.  The two sheets look similar, but the underlying functionality is 99% different.

I also notice that I put the contents of the README.md file into the sheet.json instructions, because I thought I was supposed to.  But when I’m looking at others’ sheets, they only put a short blurb into that field of the sheet.json file.  Should I alter my sheet.json and change my instructions to just a short paragraph?

sir-argo:
Sigh... I just realized I forgot to update the version # in the character sheet.  I left it as 0.9.30 and it should be 1.0

I'll see if I can fix that.

Anduh:
I think it's better to shorten thejson text and possibly link to the readme at the end.

And you might want to read the comments, regarding the submission of duplicate sheets for existing systems...
https://github.com/Roll20/roll20-character-sheets/pull/3841

sir-argo:
Ok.  I shortened the instructions and put in a link to the README.md file.

sir-argo:
I've read the thread to which you've linked.  There is also a similar discussion in a thread I started.

Whew.  That's a huge problem and one that can probably only be resolved by Roll20.  Basically, there is a policy to only have one sheet per game system.  While I understand that policy from the Roll20 side, it does restrict the users to only one option.  So if some users want a full featured sheet and others want a sheet without roll buttons or sheetworkers, somebody is not going to get what they want.  I do see that character sheets are grouped into categories.  Most of these categories are for game systems, but some are not (Official Roll20 sheets).  What I'd like to see is a new category called "HeroSystem 6 Games" and both the old sheet and my new sheet added as members of that group.

Solarswordsman:
Hi @sir-argo,
Because of the explosion of character sheets on Roll20 and a number of other reasons, we avoid having more than one sheet for any given game system.  There are some exceptions, such as when games are extremely popular or for sheets that existed before this policy was enacted but have not since had their functionality merged / mirrored in a single sheet to a sufficient degree.
The "Official Roll20 Sheets" section of the dropdown is in fact a meta-category that lists sheets authored directly by us; you'll still find them in their respective sections, as well.

For this sheet, if you feel it is a complete improvement upon the existing sheet, there are two options.  
1) Support upgrading from the previous sheet for users of that sheet via sheet workers, transitioning their attributes into the ones this sheet uses.  This is likely to take some work, but is the best for the community.
2) We can deprecate the old sheet, removing it from the list but allowing players using the sheet in their games to continue using it, while at the same time adding your sheet as the HeroSystem 6 sheet.  They will not be able to make any new games using the old sheet after that point, unless they are a Pro user and use the old sheet's code directly from this repository as a Custom sheet.

The previous Herosystem 6e sheet has not been maintained in some time, so I leave the decision here to you as to the best course of action, as you have clearly gone to great effort in creating a new sheet for this system.  As I said, implementing upgrading the old sheet is likely the smoothest and best to keep this gaming community in one piece, but if that isn't feasible, we can go to option two.  Please let me know what you'd prefer.

Thanks,
Jeff

sir-argo:
I don’t like the idea of auto-upgrading others’ games to the new sheet.  I wrote a message in the forums about how some people prefer minimalistic character sheets and others prefer character sheets that are full-featured.  You can’t have one sheet to satisfy both.

Doing option 1 would force users of the older, minimalistic sheet to the new one.  I don’t want to do that.
I did send an email to the author of the old sheet and he has not responded.

So if I have to choose, I think option 2 is the best course of action.  Do I withdraw my pull request and rename the sheet “HeroSystem 6” or will this take special handling from you?

Solarswordsman:
Fair enough.  In that case, go ahead and leave the pull request, and I will deprecate the old version and take care of the rest.

Thanks very much,
Jeff

sir-argo:
I will help in anyway I can.  Do you want me to edit these files and change references from "Hero6Playable" to "Hero System 6e"?

Solarswordsman:
No need; indeed, the folder name must be different in order for me to deprecate the existing sheet properly.  Leave everything as it is and the sheet will go out with the next sheet push, which should happen on Tuesday morning.

Thanks!
Jeff

sir-argo:
Ok.. something weird is going on.  I went into the game I use to develop the sheet and the sheet is all messed up.  The font is tiny and everything is mashed together.  I didn't change anything.  I had given a copy of the sheet to someone in my group, so I went and checked his game and it is the same.  I then read someone posting something about "Roll20 CSS changes", but I can't find an official posting in the forums.

Did Roll20 change something?

Solarswordsman:
Indeed, we've updated bootstrap and jquery.  We've tried to override and maintain backwards compatibility where we can, but most sheets will be affected by some changes to line-height and font size in particular.  The relevant post on the forums is here: https://app.roll20.net/forum/post/6532499/bootstrap-and-jqueryui-updates#post-6536343

sir-argo:
Ok, this really sucks.

Solarswordsman:
I looked through your code, and it seems you're making a lot of use of `rem` for sizing.  This is reasonable on websites where you control the entire site's html container, but this is not the case for character sheets.  Changing to `em` will  use proportions relative to the `.charsheet` container whose font-size has been corrected for and more importantly can be modified by you.  The new font-size of the html element is 10px, whereas before it was 14px.

I can merge this as-is, and then await styling updates from you and push them as soon as they come in (as the post says, we are pushing out styling updates for this breaking change on a daily basis for a time).  I can wait to replace it in the dropdown until you make those styling changes.

Sorry again for the inconvenience.  Let me know if the above works for you.

Jeff

sir-argo:
Ok.  I have made changes to all 5 files.  At this point, I assume it would be easier to cancel this pull request and submit a new one using the old name "Hero Games 6e".  Should I wait for you to save off the old sheet as _legacy first?

Also, there is a companion api script that I submitted three weeks ago that is already up, but since we're not going with "Hero Games 6e", that api's name will no longer match.  So I need to rename that too.  Would the best method for that be for me to just submit a new pull request for the api under a new name, and in the comments, indicate that version under the old name be removed?

Solarswordsman:
You should keep the name of the directory different from the previous one, as I will be deprecating it so that it is still usable by folks that already have it in their game (In other words, the previous sheet still has to exist, it just won't be selectable in the dropdown).  The directory name doesn't matter for its name in the dropdown (that name is specified in the `approved.yaml` file), I will name it appropriately.

You can change the directory name however you like before submission (though this name is just fine, it would be easiest for you to simply push your changes to this branch before I merge the PR).  Just keep it different from the existing one.  I'll handle the rest. 

As for the API script, if you want it to have a different name, just change its in the script's `script.json` file and submit a PR on that repository.  Again, the name of the actual directory does not matter.

sir-argo:
So keep this pull request and update the 5 files?

Solarswordsman:
Yes.

sir-argo:
Ok.  I hope I got that all of this correct.

I updated the 5 files (actually 3 were removed and replaced with different names).  And I changed the parent directory name.  Like I said, I hope I got all of that right.

My intent is to go over to the api directory and change the directory name for Hero6Playable to HeroSystem6e, but I assume that will be handled separately.

PR #3847, opened 2018-06-20
Insane Character Sheet - Korean

avocadoedition:
Insane Character Sheet - (lang)Korean


Solarswordsman:
Hi @avocadoedition, now that we have a translation system in place, we no longer use directly-named languaged-versions of character sheets.  I would ask that you please rename the sheet to simply the name of the game, and then submit it as is.  It is fine to be in the Korean language for the upload, it just should not be labeled as such; if for example someone wanted to make an "English version" of the sheet they could then add translation keys and strings.

Thanks,
Jeff

avocadoedition:
I am convinced.
I accepted your request.
These files have been modified.

avocadoedition:
Thank you.
That code is better than this code.
I am very happy to be able to use the character sheet of this system on Roll20.
If I add a Korean translation to the character sheet, when will the roll20 system be updated?

Solarswordsman:
If you're willing to drop this code and add a Korean translation to that sheet, then we can put out the sheet right away and then add your translation right on our next sheet update next Tuesday.  I think that would be the ideal solution.

avocadoedition:
Thank you for your kind advice.
I will update the Korean translation on the Shinobi_gami character sheet.
Although this code was not used, it was a pleasant experience.
I wish you a happy day today!

zeteticl:
@avocadoedition  
I also like your design style.
I already download your code, and try to make it together. 
.....in a month?

PR #3850, opened 2018-06-21
fixing advanced skills

abishaij1:


Solarswordsman:

Solarswordsman:
Once you've resolved this issue (One method would be to open the files in a text editor like Sublime Text, and change the line endings on the file, then commit those changes), please submit a new Pull Request so that we can review your changes.

If you have any questions, feel free to ask.

Thanks!
Jeff

PR #3853, opened 2018-06-23
Small fix

ap3h3ad:


Solarswordsman:
@ap3h3ad, your fork seems to be out of date and is submitting the deletion of files from another character sheet.  Please merge from this master repository into your branch to ensure you aren't committing removals that shouldn't be taking place.

PR #3854, opened 2018-06-23
Complete Redesign

Almos80:
This is a complete redesign and overhawl of the Godbound sheet to resolve several issues. There was sufficient desire from the community for me to publish my work, as shown in this forum post: https://app.roll20.net/forum/post/6269089/new-godbound-sheet/?pageforid=6514817#post-6514817. I also received the original author's blessing to publish this new sheet and replace his work.

### DIFF
--- a/Godbound/README.md
+++ b/Godbound/README.md
@@ -1,9 +1,31 @@
 # Godbound Character Sheet
 
-This sheet is created for use in Godbound games on Roll20. Based on the Character sheet design included with the game. Some reorganization done to make it easy to use on Roll20.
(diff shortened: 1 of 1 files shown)


Solarswordsman:
I've read through the thread and agree on the replacement.  Please note that since you've changed the sheet rather than created a new directory, I can't "deprecate" the old sheet so ALL existing games will forcibly update to this sheet and users _will_ lose data.  It seems everyone on the forums is OK with this, but I would simply recommend that you be prepared to handle bug reports if unaware users are seeing lost data.

Jeff

PR #3857, opened 2018-06-24
CWOD-DAV20 Character Sheet Spanish Translation

ekisster:
Translation into Spanish of the sheet [CWOD-DAV20]


Solarswordsman:
I'm sorry, but we [no longer accept new sheets that are just an alternate translation to an existing sheet](https://app.roll20.net/forum/post/3762625/policy-change-for-alternate-translation-character-sheets). If you would like to add a new translation for this sheet, please update the original sheet to use [Roll20's sheet translation] tools(https://wiki.roll20.net/Character_Sheet_i18n).

PR #3862, opened 2018-06-25
Blades in the Dark: 3 small fixes

joesinghaus:
- fixed faction ordering
- chat image version forced to "thumb" size
- fix line-height in order to deal with Roll20 changes


Solarswordsman:
Hey @joesinghaus, it would be super helpful if you could let us know about breaking styles that were introduced with the new update (we updated a lot of libraries, including CSS used in the editor), probably in the bug report forums or character sheet forums (we may make a thread dedicated specifically to this, but there isn't one for now).  We did our best to override these to prevent breaking character sheets, but if there is a CSS rule we've overlooked it'd be useful for us to know so that we can implement an override if appropriate.

Thanks for being on top of this, in any case.  Your work is always appreciated both by us at Roll20 and the community.

Jeff

PR #3865, opened 2018-06-27
Update de.json

PhilHoffm:
Updated version of this file


Solarswordsman:
Hi @PhilHoffm, I'd recommend you apply to become an official translator by sending an email to team@roll20.net.  This will make it easier to collaborate with other translators and to see a better picture of the translations as a whole, rather than doing them in the .json files themselves.

That said, thanks for the translation updates.

Jeff

PR #3867, opened 2018-06-27
Re-added the spell prep checkbox to the sheet

oukag:
The update for the charactermancer seemed to remove the spell prepared checkbox from all repeating spell sections. This update will put those back in place so that users can prepare/unprepare spells.


Solarswordsman:
The prepared checkbox is now the same as the indicator for preparedness (the red dot) -- the intent was so that you could quickly prepare and unprepare spells without needing to open the spell details, as you will be doing that much more often than changing how a spell works.  We are evaluating the impact of this UI change, but it is documented [on the wiki.](https://wiki.roll20.net/5th_Edition_OGL_by_Roll20#Adding_Spells)

PR #3873, opened 2018-06-28
iSANe : Translation - Korean

avocadoedition:
iSANe : Translation - Korean

and, UI Style (CSS) Modify (Fix element space and misaligned layout)


avocadoedition:
There is a word difference between actual English and Korean.
Because it translated based on [character sheet] published in Korea.
I think it would be more convenient to use the word [inSANe character sheet] that was officially published in Korea for Korean people to use.

ex:
Auto Skill List  자동 스킬 목록 → Specialty List 특기 리스트
Skill List 스킬 목록 → Ability List 어빌리티 리스트

PR #3895, opened 2018-07-02
Delta Green V2.0

uriele:
Added NPC/Monster character sheet
Added Creature attack for monsters


Solarswordsman:

Please change the files to use Unix-style (\n) line endings so I can properly compare and merge your changes.

Thanks,
Jeff

uriele:
Hi Jeff,
I'm going to close this pull request and submit another with the format changes. Thanks

PR #3911, opened 2018-07-06
Unknown Armies 3e v2.0

Rats808:
The last Roll20 update jumbled some things up with the old version of this sheet, and a new official sheet has since been released, so I decided to do a re-write.

Existing sheets will keep most of their information in-tact, with the exception of the 4th identity, but now all 4 identies are more flexible in their use(can all be freely changed between Normal/Supernatural/Adept/Avatar).


Solarswordsman:
Ok.  I would consider making a post on the Character sheet forums mentioning the loss of that single attribute.  Otherwise, as everything else will remain intact, I'll go ahead and merge this.

PR #3928, opened 2018-07-09
Creation of Character Sheet for Neuroshima

ddelpercio:
The character sheet for the polish RPG [Neuroshima.](https://en.wikipedia.org/wiki/Neuroshima)
There is no english version of the source book so the Character Sheet is done in polish. Sheet is pretty barebones, however it is better than nothing.

I don't have pro so I hope the rolls work, but everything else was checked with a mock version of roll20 so the layout should be fine.


Solarswordsman:
Hi @ddelpercio, you're going to run into a few issues here with your styling and structure.  Most importantly, html/body tags and selectors in CSS are going to be stripped out.  The character sheet exists purely inside the `.charsheet` container and any selector that specifies an element outside of that scope will be stripped.  Additionally, I should note that all classes in your HTML will be prefixed by `sheet-` when they go through our parser, so your CSS should select them as such (e.g. `sheet-abilities-container` for something like `<div class="abilities-container">Foobar</div>`).  You can find more details on our wiki: https://wiki.roll20.net/Building_Character_Sheets

ddelpercio:
The sheet is very barebones, there's hardly any css to speak of. I based the stule of the sheet on the ACKS sheet, a very plain sheet that has worked fine for me in the past. I gutted any scripts and roll templates I could find and added what is needed for neuroshima. Any css problems should be insignificant for the functionality of the sheet.

Solarswordsman:
OK.  I'll go ahead and merge it then, and you can tweak css as needed once it's live.

Thanks!

Solarswordsman:
@ddelpercio So, the main issue with the sheet as I'm testing here seems to be the referencing of attributes in your rolls.  Totally understandable since you haven't been able to test with the actual roll server on Roll20.  The problem is that you are referencing attributes and including the `attr_` in the roll. 
 You do not need this.  For example, `@{attr_zrecznosc}` should just be `@{zrecznosc}`.  I'm going to leave this merged, so that you can see the issue to rectify it.  You should be able to test changes with your browser's devtools by modifying the roll buttons to use the pure attribute names.

Jeff

ddelpercio:
Ok, seems simple enough to fix. I expected problems like these without access to the roll server. I'll fix everything I can to get it working for next week.

PR #3943, opened 2018-07-10
Fix textfield background-color

Rats808:
Turns out "#00000000" doesn't work for setting transparency in roll20. Switching it out for "rgb(0, 0, 0, 0)", which worked at the top of the sheet.


Solarswordsman:
To clarify for future authors who might notice this: `rgba(red, green, blue, alpha)` should indeed work in your sheet's CSS.  Incidentally, if you want `background-color: rgba(0, 0, 0, 0)`, this is the same as `background-color: transparent;`, which may be more legible.  In any case, I'll merge this.

PR #3949, opened 2018-07-12
[GURPS] Replaced all instances of the word Fear or fear with the word Fright.…

TankFarmer:
… In GURPS the proper term is Fright or a Fright Check, not fear. Also modified the Fright check macro to replace the word Will wuth the word Fright so it states it is making the roll vs Fright and not vs Will.

Replaced all instances of the word Fear or fear with the word Fright. In GURPS the proper term is Fright or a Fright Check, not fear. Also modified the Fright check macro to replace the word Will wuth the word Fright so it states it is making the roll vs Fright and not vs Will.

### DIFF
--- a/GURPS/gurps.html
+++ b/GURPS/gurps.html
@@ -301,30 +301,30 @@
 						<button type="roll" value="@{roll} [[3d6]] @{VS} Will [[@{willpower} + @{modifier}]]" name="roll_vsWill" />
 					</div>
 				</div> <!-- .sheet-row -->
-				<div class="sheet-row sheet-sub-row sheet-row-fear">
+				<div class="sheet-row sheet-sub-row sheet-row-Fright">
 					<div class="sheet-cell sheet-col0 sheet-label">
-						<div class="sheet-popup">Fear</div>
+						<div class="sheet-popup">Fright</div>
 						<span class="sheet-tooltip">
-							Fear Check
+							Fright Check
 							<br />
 							Based on Willpower
 							<br />
 							2pts/level
 							<br />
-							Unmodified: <span name="attr_fear_check_base"></span>
+							Unmodified: <span name="attr_Fright_check_base"></span>
 						</span>
 					</div>
 					<div class="sheet-cell sheet-col1">
-						<input type="number" name="attr_fear_check" value="10" readonly="readonly">
+						<input type="number" name="attr_Fright_check" value="10" readonly="readonly">
 					</div>
 					<div class="sheet-cell sheet-col2">
-						<input type="number" name="attr_fear_check_mod" value="0" step="1" />
+						<input type="number" name="attr_Fright_check_mod" value="0" step="1" />
 					</div>
 					<div class="sheet-cell sheet-col3">
-						<input type="number" name="attr_fear_check_points" value="0" step="2" />
+						<input type="number" name="attr_Fright_check_points" value="0" step="2" />
 					</div>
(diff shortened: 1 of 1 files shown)


Solarswordsman:
Hi @TankFarmer, while the terminology may need changing, it may be easiest for you to only change the user-visible indicators of the word.  The problem with your changes is that you are changing attribute names, and so the values in those attributes will be lost for current users of the sheet.  You could change the attribute names if you handle retrieving the previous attribute values and updating them to their new attributes via a sheet worker if you really wanted the underlying attribute names to change.

For now I'll close this, but feel free to make a new PR with either of the above modifications (either only change visible verbiage, or make a sheet worker to convert the attribute values).

Thanks!

TankFarmer:
Well I guess that is a problem as I have no idea how to do this vi a sheet worker.So for now I will resubmit and change the user-visible indicators.
Thanks
Tom

    On Tuesday, July 17, 2018, 6:51:32 AM PDT, Jeffrey Lamb <notifications@github.com> wrote:  
 
 
Hi @TankFarmer, while the terminology may need changing, it may be easiest for you to only change the user-visible indicators of the word. The problem with your changes is that you are changing attribute names, and so the values in those attributes will be lost for current users of the sheet. You could change the attribute names if you handle retrieving the previous attribute values and updating them to their new attributes via a sheet worker if you really wanted the underlying attribute names to change.

For now I'll close this, but feel free to make a new PR with either of the above modifications (either only change visible verbiage, or make a sheet worker to convert the attribute values).

Thanks!

TankFarmer:
JeffreyQuick question, if I may ask:Does changing the type of an attribute require any special code changes to retain existing data?
There is an attribute which is restricted to 'value' and I want to change it to 'text'.The current attribute only allows for a numerical entry BUT it needs to allow for several entries plus alpha characters as well.I am not changing the name of the attribute, just its type.Tom

    On Tuesday, July 17, 2018, 6:51:32 AM PDT, Jeffrey Lamb <notifications@github.com> wrote:  
 
 
Hi @TankFarmer, while the terminology may need changing, it may be easiest for you to only change the user-visible indicators of the word. The problem with your changes is that you are changing attribute names, and so the values in those attributes will be lost for current users of the sheet. You could change the attribute names if you handle retrieving the previous attribute values and updating them to their new attributes via a sheet worker if you really wanted the underlying attribute names to change.

For now I'll close this, but feel free to make a new PR with either of the above modifications (either only change visible verbiage, or make a sheet worker to convert the attribute values).

Thanks!

Solarswordsman:
You can change an `<input type="number" name="attr_attributename">` to an `<input type="text" name="attr_attributename">` with no issue, yes.  If it has a default `value`, I would keep it as such.  Note that if you change the input type you might have to do a bit of CSS to make sure the layout remains consistent.

TankFarmer:
JeffreyThank you for the quick answer.All seems fine here, I can make the CSS layout changes as needed, I figured that part out.This is exactly what I needed to know.As you can tell I am rather new at this BUT no one else is making the changes we need so I took it upon myself to give it a try.Thanks again,Tom
    On Tuesday, July 17, 2018, 6:47:38 PM PDT, Jeffrey Lamb <notifications@github.com> wrote:  
 
 
You can change an <input type="number" name="attr_attributename"> to an <input type="text" name="attr_attributename"> with no issue, yes. If it has a default value, I would keep it as such. Note that if you change the input type you might have to do a bit of CSS to make sure the layout remains consistent.

PR #3957, opened 2018-07-15
PF Community. HL-import update. Evaluation of macro(a) fields update

vince-roll20:
- HL-import feature updated; Fixed regex expressions that required content inside of parenthesis,
causing exceptions on strings like "Knowledge ()". (James W.)
- macro(a) updated to handle the inclusion of multiple grouping rolls.  ie " {x,y}kl1+{x,y}kl1 "
- more changes converting inline rem values to px for the "mancer" update.
- adjusted spell domains section to better fit sheet constraints in compact mode.


Solarswordsman:
Hey @vince-roll20, just as a heads up, while the root element font size has changed with our CSS updates, it's still totally feasible to use `em` units (which will base off of the font-size specified anywhere in the cascade from `.charsheet` down) instead of `px` to support different viewports / scaling.  Whether you do this is of course up to you, but as I'm sure you know these units are generally better for scaling.

Admittedly this means some more fiddling by you since font-sizes undoubtedly change throughout the various sections of this sheet.  For what it's worth, we have no future plans to change the root element font-size so you could adjust your `rem`s instead to keep using a global basis, though I fully concede I cannot foresee the future--however since this has caused problems across many sheets, I expect this to be something we look at very carefully if ever we perform another update of this kind.

Regardless, thanks for all your hard work as usual.

Jeff

vince-roll20:
Thanks for the tip Jeff.  I just needed to make a quick conversion to get the sheet back up for everyone.  Redoing rem or converting to em is on my lonnnnnng todo list...  ;-)  Cheers

PR #3965, opened 2018-07-17
Song of Swords - Update to beta 2.0

Lloydski246:
Large update to the sheet. My initial sheet was based on the previous version and therefore was not fully useable.


Lloydski246:
Had to stick a second update because I didn't commit the right html. Sorry if this has messed up the request.

Anduh:
Probably not, it's the end results that matter. I've more than once continued working and updating a pull request after my initial push.

PR #3972, opened 2018-07-18
Wrath and Glory Multi-tab with API v1

bts1967:


bts1967:
Hi Solarswordsman, This is on my list of to-do's.  Obviously that needs to be part of the initial release. Question though, there is at least one other sheet that is Advanced API only... Dark Heresy: 2nd Edition (Advanced API).  That is why I thought this was not an issue.

PR #3977, opened 2018-07-18
1.2.0

Meteox:


Solarswordsman:
Hi @Meteox, it looks like you've changed a lot of attributes and structure on the sheet.  How will existing users be affected by this change (e.g. will they lose data, does a sheet worker handle the update, etc.)?  I want to make sure that the update does not cause existing users to lose data or other character information.

Thanks,
Jeff

Meteox:
Only Attributes thats serve no function have been deleted. One Attribute that has a new name has a worker sheet to take over the old value. There will be no data loss

PR #4002, opened 2018-07-28
Add Karyu Densetsu sheets

trueshinken:
OK, I think this time I got everything right. Thanks for all the help!


Solarswordsman:
Looks good!  Only thing you didn't update is the `sheet.json`, it was referencing the generic `kitchensink.png` file and had the default instructions.  I've fixed that, so assuming the roll20userid is correct, you're good to go!  Thanks for the sheet!

Jeff

PR #4010, opened 2018-08-03
Add Undying sheet

Niduroki:
Add a sheet for the game [Undying](https://enigmamachinations.com/2016/02/13/undying-update-4/), inspired by the [original character sheet](https://enigmamachinations.files.wordpress.com/2016/02/undying_sheets_rev020416.pdf).

This is a PbtA game, so ideally it should be categorized as such in the sheet-list.


Anduh:
In the sheet.json file, you have no "instructions" section, might be good idea to add one, where or not it's mandatory. Haven't seen a sheet without it.

PR #4012, opened 2018-08-05
Initial dump

Medieve:
Full rehaul of Coal's sheet. Beautifying the skeleton sheet and filling out with more capability and auto-calc formulas. Practicality adjustments.


Solarswordsman:
Hey @Medieve, this looks _great_!  My only concern is the effect this overhaul will have on existing users of the sheet.  It looks like at least most of the attributes remain in existence, but due to the large scale of the overhaul I can't check every single thing.

Will existing users be able to open their current sheets without significant data loss?  As long as that's the case, I'm more than happy to approve this upgrade.

Thanks,
Jeff

(Note; I kept CoalPoweredPuppet's roll20userid in addition to your own--this is currently not well documented, but you can simply put user id's in comma-delimited fashion)

Medieve:
Hi Jeffrey, thanks for the kind words!
I've tried to keep as much consistency as possible but there are two
exceptions:

1. Minor exception: I added in a auto-formula for the individual
Abilities/Skills a character gets which only updates when the Score
changes, so players will have '0's in several *new* fields until they
fiddle with them.
2. Moderate exception: Several equipment attributes were combined and
simplified into new attributes, so players would need to re-enter equipment
information, though these are Table-referenced from source books.

I originally requested the sheet from puppet but I can't really guess how
many people are using the current sheet (is that even trackeable?) Is it
worth mentioning in the description about the data changes? I'm also
realizing I might want to add in some notes about how to use the sheet's
more advanced features.

Medieve:
Eh, the sheet is reasonably easy enough to decipher, if I get a message
about it I'll consider adding notes.

clevett:
Hello @Medieve
We're going to go ahead and push this sheet. As it as its a lot of changes there may be issues that crop up. If you could make a post or keep an eye on the character sheet forums that would be appreciated.

Thanks,
Cassie

PR #4014, opened 2018-08-05
Fixing the Multiplier Option

KPhan1212:
One of them had the old input multiplier instead of the drop down menu.
The other one had a misspelling that needed to be fixed.


KPhan1212:
One of them had the old input multiplier instead of the drop down menu.
The other one had a misspelling that needed to be fixed.

PR #4018, opened 2018-08-06
Formatting corrections

shukinhekon:
Corrected the positioning and spacing of most skills and miscellaneous text in the sheet.


Solarswordsman:
Hey @shukinhekon, I've noticed you changed several attribute names in the sheet.  This will cause a loss of data for existing users.  Please provide a sheet-worker to "version-up" from the previous attributes, or retain the existing attribute names, so that users do not lose data.

Thanks!
Jeff

shukinhekon:
Thanks for noticing that! It was something that I didn't even notice
changing. Every correction brought something else that needed to change
into light, so I'm sorry for adding so many different changes to the same
push request. I guess the attribute names are correct now. Please tell me
if there's anything else that might hamper the existing users!

Thanks for your time!

PR #4031, opened 2018-08-09
Add vampire 5th edition sheet.

jgoppert:
This is a significant improvement over the existing vampire sheets and incorporates the rules for the new official 5th edition Vampire release. It should also be easier to maintain since I use python scripting to generate all of the tedious bits. It eliminates the need for most html copy and paste.

All stat groups also have a checkbox that toggles between the dots layout and the table layout that show you how the dots were obtained (xp or initial character points).

It also calculates spent XP, so players and storytellers can easily upgrade their characters.

The health and willpower levels are also calculated (level = max - aggr - sup) , so you can display health bars and willpower bars on your characters that actually make sense. If you add a superficial damage level, it will automatically convert to aggravated damage if all health/will slots are filled as you would expect.


jgoppert:
screenshot from 2018-08-09 06-13-37

Anduh:
Whichever of these becomes are accepted, I think the [Alpha version](https://github.com/Roll20/roll20-character-sheets/tree/master/WODVtM5eAlpha) of the sheet can be removed from the menu as an option afterwards.

PR #4066, opened 2018-08-16
Wrath & Glory v2 Release

bts1967:
Update to core rules.


joesinghaus:
You seem to have opened three pull requests for individual files for the same sheet. You should probably upload all of them to the same branch so that you only need a single PR.

bts1967:
They are different sheets... readme, html, and css.

PR #4076, opened 2018-08-19
Eirendor character sheet Rev2, new layout

Seregras:
I have changed the general layout of the character sheet, to make it more user friendly. I designed and uploaded the original character sheet a year ago.


clevett:
Hello @Seregras 

I see that you changed a number of attributes from Spanish to English such as attr_raza to attr_race . This will cause character sheets to no longer display the information they had in the attr_raza input. You'll want to write a sheet worker to help users retainer their data on their character sheet. Directions for sheet workers can be found in the [Roll20 Wiki](https://wiki.roll20.net/Sheet_Worker_Scripts). 

The best way to handle this is to make inputs type="hidden"  with the old attributes names then have a sheet worker that triggers on [sheet open](https://wiki.roll20.net/Sheet_Worker_Scripts#sheet:opened) to take the old data and SetAttrs it into the new fields. If you have questions please let me know.

Thanks,

Cassie

Seregras:
Hi,

Thanks for the info. I wanted to change  the fields to English in case the game gets translated in the future. What if I keep the translated fields as they were? I mean, keep attr_raza as it is? Would that require a sheet worker? I just intend to avoid extra work.

Many thanks!

clevett:
Greetings @Seregras  user G-G-G provide a helpful answer. Yes, if you keep attr_raza the same and just change the label for the input that is okay as it will not effect their character's actual stats.

Seregras:
Thank you all!

clevett:
Close this request since we've not got, [4110 ](https://github.com/Roll20/roll20-character-sheets/pull/4110) with the new changes

PR #4081, opened 2018-08-20
Fixed Perception Button to use Wisdom Modifier and not Dexterity Modi…

Kognar:
…fier

Perception button was using dexterity_mod instead of wisdom_mod for the roll.  Visually it was fine but when rolling it was wrong.


clevett:
Greetings @Kognar 

Thanks for attempting to fix this bug. :) Two pull request went in to fix this so I am going to close this one. The issue was fixed in request #4082 as well.

Thanks,

Cassie

PR #4085, opened 2018-08-21
[GURPS] Inventory and Language Box reformatted

KlingonMike:
•	Updated the Language Box format – Cosmetic Only.
•	Added additional options for Native Language in regards to Spoken and Written levels as a Disadvantage.
•	Updated the Inventory Box format – Cosmetic Only.
•	Added fields for ‘TL’ (Tech Level) and ‘Ref’ (Reference) to the Inventory repeating section.
•	Fixed bug for Single Modifier Prompt and Extended Modifier Prompts if a ‘+’ is used for a positive modifier entry.


clevett:
This was a very helpful comment. Thanks for the details.

PR #4086, opened 2018-08-21
[Weaverdice] V.1.32

Anduh:
- fix skill rolls added in V.1.2. Added option for using stat bonus and skill bonus to rolls,plus a field to custom the roll. Default "k1" keeps highest dice

- Stat value added above the stat modifier

- adds sheet version in the corner

- hearts and dice changed to red

### DIFF
--- a/Weaverdice/README.md
+++ b/Weaverdice/README.md
@@ -3,7 +3,7 @@
 
 Any improvements to the sheet are welcome.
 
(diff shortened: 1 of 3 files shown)


clevett:
@Anduh Sorry for getting back to this late. Unfortunately this weeks request were processed early. We'll be back to our regular scheduled programming this week.

PR #4092, opened 2018-08-22
Pathfinder Playtest: updated untrained from -1 to -2

Ansrala:
Pathfinder Playtest: updated untrained from -1 to -2


clevett:
Greetings @Ansrala 

It looks like this got updated in two places. It was also fixed in pull request [4113](https://github.com/Roll20/roll20-character-sheets/pull/4113). I've merged that one alone for now since it had a number of other updates. :)

PR #4093, opened 2018-08-23
Earthdawn 4e: Version 2.3

dougansf:
Questors tab added


clevett:
Greeting @dougansf ,

I'm going to approve this merge request but I wanted to let you know that in the near future we have new standards coming that will effect this sheet. All sheets are going to need to be accessible without the use of the API. I see this sheet is using Power Cards to do all the rolls & ChatSetAttr . While these can be useful companion features, it should not be the default as it will make the sheet unusable for basic & plus accounts. 

Thanks,
Cassie

dougansf:
Thank you for informing me of the impending change.  Is it known when that
will happen?

Due to the complexity of the rolls involved with this sheet, I'm not sure
that will be possible, but will make the attempt.

clevett:
The stated changes will hopefully be posted in the near future but I don't expect to start 'enforcing' them on for the older sheets for another month or two at the earliest. I just want to let sheet authors know they are coming. :) 

A good way to keep your fancier rolling is to make a checkbox toggle that switches out the API roll for a more basic roll macro. I know Earthdawn's step system makes for a bit of an interesting challenge. The simpler rolling does not need to cover all the possibility but just provide minimal functionality. Let me know if you run into blockers and I'll be happy to try to help. I've a passion for the 4th World.

Thanks,
Cassie

dougansf:
You are my new favorite R20 rep.  :)

I'm teaching myself HTML by creating this sheet.  I've gotten help on the
R20 forums from some great people to get me this far.  I now have an
assistant from the Earthdawn West Marches campaign, who is easily my
largest user of the sheet.

We started figuring out a basic roll template.  I already had a basic roll
to give to folks that were not Pro Subs.
The idea of a toggle to activate the APIs would be great, though I have no
idea how to implement that.  Recommendations would be greatly appreciated.

Thanks,
Brett

clevett:
The new [Wrath & Glory](https://github.com/Roll20/roll20-character-sheets/tree/master/Wrath%20and%20Glory%20API) Sheet did this an interesting way.

First at the top of the **HTML** they used a checkbox to determine if you want to use standard rolls or API:

`<input type="checkbox" name="attr_roll-toggle" class="sheet-roll-useDice" value="1" title="check to use the standard Roll20 Dice Rolling mechanism vs the Wrath & Glory Dice Rolling API." checked="false"/><Span class="sheet-block-text"><b> Use inline dice (vs API)?</b></span><br>		`

Then in the sheet they made two separate buttons like this:

`<button class="sheet-diebutton-api" type="roll" title="Click to make a Ballistics skill test."  value="!wag {{[[[[@{BallTotal} + @{BallBonusRoll} - 1]]d6]]}} {{[[1d6]]}} --Name|@{character_name} --Skill|Ballistics --Modifier|@{BallRollMod}"  name="roll_ballistics" style="font-size: 175%;"></button>`

`<button class="sheet-diebutton-simple" type="roll" title="Click to make a Ballistics skill test." value="&{template:roll_dicepool} {{name=@{character_name} makes a Ballistics Skill test}} {{dice=[[@{BallTotal} + @{BallBonusRoll}]]}} {{roll=[[[[@{BallTotal} + @{BallBonusRoll} - {1}]]d6>4]]}} {{wrath=[[1d6]]}}" name="roll_dice4ballistics" style="font-size: 175%;"></button>`

Lastly in the **CSS** they included this:
`input.sheet-roll-useDice:unchecked 	~ div table button.sheet-diebutton-api { display: inline; }
input.sheet-roll-useDice:unchecked 	~ div table button.sheet-diebutton-simple { display: none; }
input.sheet-roll-useDice:checked    ~ div table button.sheet-diebutton-simple { display: inline; }
input.sheet-roll-useDice:checked    ~ div table button.sheet-diebutton-api { display: none; }`

What is happening is the "API roll button" is displaying on the character sheet unless that checkbox at the top is checked. When the roll-useDice checkbox is checked then the "API roll button" is hidden and the "simple button" is displayed. It would likely be a better practice to have the simple one displayed by default but this was a patch the author was nice enough to add when the sheet was first submitted. :)

dougansf:
Awesome.  I'll see if I can incorporate that.

For the standard Roll20 roller, I've figured out a macro to ask for a Step
value, roll the right exploding dice, add Karma dice and then compare that
against a target tokens' Defense to give the number of Successes (1 success
for meeting the Target Number, +1 for every 5 over the TN).

Is there a way to display the Roll Result AND the number of Successes at
the same time?  I can only seem to do one or the other.
This is the big reason I went with Power Cards.

clevett:
You may want to outline the rules, goal, and dice used on the [Character sheet forum](https://app.roll20.net/forum/category/277980) then ask for help. There are likely better people over there than I who can provide a suggestion. I have to relearn how dice & macros work with new every sheet. :)

Hopefully some one on the forums can come up with an elegant solution. If there isn't one with the current dice roller then keep in mind the simple roll just needs to be functional. This may mean you just show the dice result in a nice roll template and let users apply the rules as need. Just like if they were playing in the physical world with real dice.

PR #4110, opened 2018-08-27
Eirendor character sheet new layout

Seregras:
Following the instructions from last week, I have changed back the attributes and fields from this version to the names of the previous one. Unfortunately, the gear is now handled as a repeating section, and it was a textarea in the previous one, so it will have to be repopulated.


clevett:
This is greatly improved! The loss of the text area is manageable given how big a change these updates were. Thanks for fixing the other attributes.

PR #4122, opened 2018-08-28
Das Schwarze Auge 5: add of several data-i18n tags for the dsa5 character sheet

dyvim:
allow the creation of a translation.json file for this character sheet


dyvim:
@Meteox is this OK for you?

Meteox:
I already started work on the i18n Addition and i´m done with the first ~3000 lines, there are also updated in my current Version so this would be counter productive

dyvim:
OK.

If you see a way I can help you don't hesitate.

I'll work on the french translation as soon as the translation.json file
will be available if I can't help before.

Thomas Coquery

PR #4128, opened 2018-08-29
Update Mass Effect.html

SilverRahl:


ragnarocko:
So this is still not fixed yet its kinda weird but please keep try.

clevett:
Greetings @SilverRahl 

I see some attributes are being removed with this update. These will no longer be readily available to the users of this sheet. Can you give me a brief description of why they are being removed?

Thanks,
Cassie

SilverRahl:
I moved recoil upper weapon attributes category to below to even these to categories in the sheet view. Like 6 attributes upper side and 5 attributes below, so its clean.

Some of the attributes have been deleted like ammo capacity and so. We have a special properties so we can write it down there instead. So there will be less coding so less problem. Generally I changed roll coding line with working coding line from classic modern d20 and modified it. I am still trying to figure it out where is the problem.

PR #4129, opened 2018-08-30
13th Age neovatar: Fixes to CSS and HTML to display correctly with Roll20 base CSS changes

neovatar:


Anduh:
So the css file you removed(9000 rows) was redundant, and it's covered elsewhere?

neovatar:
Its not removed, just re-downloaded it from roll20. It is only used for the preview via grunt. The new file seems to be packed into just one line. Just double check, it's ok.

PR #4133, opened 2018-08-31
Eirendor character sheet corrections

Seregras:
The following corrections have been implemented, after the new layout included last week:

- Bonus per level formula changed from "ceiling" to "floor", as apparently was showing decimal numbers in rolls for levels higher than 2.
-Fields "rasgo1" was duplicated, changed to "rasgo1" and "conjuro1
-Rolltemplate had a word misspelled, removed.

Thanks

### DIFF
--- a/Eirendor/Eirendor.css
+++ b/Eirendor/Eirendor.css
@@ -432,6 +432,14 @@
 	box-shadow: 0px -2px 0px 0px #000000 inset;
 }
 
+ input[type=text].sheet-rasgo {
+	text-align: left;
+	font-weight: bold;
+	color:#7a0000;
+	font-size: 16px!important;
+	font-family: 'Contrail One';
+}
+
  input[type=number].sheet-underlined {
 	box-shadow: 0px -2px 0px 0px #000000 inset;
 }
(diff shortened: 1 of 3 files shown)


clevett:
Greetings @Seregras 

Check your **sheet.json**  . It is currently not a valid JSON due to a period at the end of the last line. We manually fixed this last week. 

Did you copy/paste the sheet before doing a pull request? It looks like the entire sheet is a resubmit which makes it hard to see what actually changed. Was there anything else besides your list?

Thanks,
Cassie

Seregras:
Hi clevett,

Sorry for the late reply, I have just seen your message. Thanks for the JSON remark, I will change it for next corrections.

I didn't understand your last point, but I see the sheet has been accepted.

Thanks!

clevett:
Greets Seregras,

Wing-it merged the request last week so it may not be an issue. He fixed the JSON for you. We wont need to worry about this pull request. :)

You may want to double check your text editor to ensure you're using Unix friendly Line Endings. Where this is set can very by text editor but you can probably find out how to fix it with a google search.

Thanks,
Cassie

PR #4147, opened 2018-09-04
[GURPS] Bug Fixes and added FoA

KlingonMike:
•	Fixed bug with calculating Current Move for Heavy and Extra Heavy Encumbrances and not using a Minimum of 1. Reported by Ray G (@raymond-g)  (UserID: 1600355).
•	 Fixed bug with calculating Total Weight from Inventory which at times created a weigh of more than 3 decimal places.
•	Added a ‘FOA’ (Frequency of Appearance) field under Advantages (& Quirks)
•	Modified the Tool Tip for SCN under Disadvantages (& Quirks) and Racial Traits to include the wording ‘OR Frequency of Appearance’– Cosmetic Only
•	Added Max Load to the Encumbrance table.


KlingonMike:
[GURPS] Version 1.4.0

PR #4149, opened 2018-09-04
The ADVANCED: Pandemonium, Wrong logo in the example (old logo)

odanion:


Anduh:
BTW might be a good idea to host the logo here on github rather that letting it be external link to imgur. Many or most sheets host their pictures here.

odanion:
Thanks, just uploaded the logo and changed the sheet to use it at least I believe the code is correct (found it used in another sheet)

clevett:
Greetings @odanion 

There appears to be a conflict in your merge request. This will need to be resolved before I can merge it into our code push tomorrow morning.

Thanks,
Cassie

odanion:
I do not understand why The ADVANCED Pandemonium/the_ADVANCED_Pandemonium.html shows as having a conflict, 1 line removed and 1 line added, even the preview shows line 2 removed and a new line 2

odanion:
Ok, that should have done it :D

clevett:
Greetings, @odanion 

I noticed one additional concern. Your code was using the default Windows new-line characters (\r\n) when we need it to use the default Unix character (\n). Our website runs on a Unix system, and requires the Unix line-endings rather than Windows. To avoid this in the future, you can change the settings in your text editor to use Unix line-endings rather than Windows.

This is an easy fix but may vary depending on your text editor. I use Sublime and all that I'd need to do is open the html View -> Line Endings -> Unix . 

These line endings will need to be fixed before we can merge tomorrow morning.

Thanks,
Cassie

odanion:
Ok, removed the old code and uploaded a new version (used notepad++ and set to unix line endings, so this should be correct)

clevett:
Awesome! Thanks for trying. If they don't work we'll fix them tomorrow so as to not delay the update. Line endings are a bit tricky.

odanion:
Thanks, I actually work with C#, HTML is like a strange alien, never even realized that unix used different line endings XD

PR #4156, opened 2018-09-05
Vampire V5 - Inline Roll enhancements

Kirintale:
Updated inline rolls so they now take into account hunger properly.
Renamed the JS and updated the images on it to link to Github. This should hopefully fix the broken link issue.


Kirintale:
Thank you ^_^.

Hopefully should be in a state where I can have a break now.

Kirintale:
Woo game on!

PR #4164, opened 2018-09-07
[ChrisPerrinMecha] Add sheet w/french translation

Al-tay:
This adds a new sheet for ChrisPerrin's Mecha with a French translation for all texts.


clevett:
Greetings @Al-tay 

I reverted this pull request since we have this sheet already in the repo. To translate it into French, the sheet author or yourself will need to add [i18n](https://wiki.roll20.net/Character_Sheet_i18n) tags. We crowdsource the translation using [Crowdin](https://github.com/Roll20/roll20-character-sheets#internationalization--translations). This method lets us use the original sheet in multiple languages.

Thanks,
Cassie

Al-tay:
@clevett I understand your point. There were already some sheets that existed in multiple versions depending on the language, so I assumed it was OK. Thanks for the pointers to the i18n tags.

PR #4167, opened 2018-09-09
AD&D1e - fix 

vince-roll20:
- max title text on equipment was wrong. changed to  @{repeating_equipment_$X_quantity|max}
- max attribute on repeating_ability changed from max to current_max in order to work with |max syntax for tokens.  Included sheet version check sheet worker to copy old max attribute data to current_max.
- fixed alignment issue with the repeating equipment fields and the column headers.


clevett:
Thanks for the comments. Those were very helpful.

PR #4177, opened 2018-09-11
[GURPS] Fixed Encumbrance Bug

KlingonMike:
•	Fixed Encumbrance Table calculations to always allow for a Minimum Move of 1 and a Minimum Dodge of 1.


KlingonMike:
[GURPS] Version 1.4.1

PR #4192, opened 2018-09-12
Added Sheet for The Witcher Pen and Paper RPG

HappyDamnTrails:
Character Sheet for new Witcher Pen and Paper RPG by R. Talsorian.


HappyDamnTrails:
Alright, updated. Shame there isn't a way for the preview images to work
like a gallery or something similar, would be pretty cool.

HappyDamnTrails:
That's an interesting work around. I reckon it's fine with just the first
tab preview since you can see there's other tabs too the sheet.

Anduh:
[Here ](https://github.com/Roll20/roll20-character-sheets/blob/master/D6StarWars/sheet.json) is an example of a "instruction" section with some more formatting and hyperlinks. HTML tags doesn't seem to works, like how in this [official sheet.json](https://github.com/Roll20/roll20-character-sheets/blob/master/New%20World%20of%20Darkness%20Official/sheet.json).

clevett:
@HappyDamnTrails 

I had to revert this change. Your pull request is for the wrong Roll20 branch. Your pull request needs to be for Roll20:master

Moonshaft:
Are you still taking feedback on the sheet at all? I was playing around with filling some of it out for characters on a game I've been waiting to do and some of the values under Magic need to be altered to accept values that aren't just numerical.
For example the option for "Duration". While some Hexes/Rituals/Spells last a set amount of rounds, others like "Glamour" have a duration of 1d6/hours and the boxes don't accept anything but a numerical value. Other spells like  "Telepathy" are always Active and cost 1 Stamina/round to upkeep. "Brand of Fire" is instant and immediate and the Ritual "Spirit Seance" is Permanent and doesn't really have a duration. I've put a 0 in there for now as a placeholder.
The "DAM" box when adding new spells is also significantly smaller than its first counterpart with no way to resize. I've attached a kind of example of what it ends up looking like.
magic

HappyDamnTrails:
Thanks for the feedback. As long as people are using the sheet and finding
things they'd like me to tweak I'll do weekly updates to account for them.
The sheet as it is represents how my personal group has come to like it
after about a month of use, but having a lot more eyes on it will be
helpful.

Moonshaft:
Awesome, thanks so much. I'd also like to suggest being able to create decimal values for items as well--especially components since a lot of them are either .1 or .4 or something depending on how _many_ of said item one has:
image

Also, when rolling a skill (ie Deciet, Awareness, etc.) what additional modifier is the pop up asking for?


> 
> Thanks for the feedback. As long as people are using the sheet and finding things they'd like me to tweak I'll do weekly updates to account for them. The sheet as it is represents how my personal group has come to like it after about a month of use, but having a lot more eyes on it will be helpful.

Moonshaft:
Huh, that's odd then. I know I don't have any other values in there but it's not very friendly to decimals. It might just be a box size issue or something:
image

Have you considered putting a kind of 'cheat box' in there for Resolve and other Defensive Totals, as well as an indicator on certain skills that they cost an increased amount when leveling up like a (2) at the end of the Skill name? i.e Crafting (2)

There's no separate box to add the Attack Bonus on the weapon of choice (Attr+Associated Skill). Is that bonus just being placed in the 'Damage' box to help save space? Just want to make sure I'm not missing something.

Having a dedication section in the **Personal History** section would be awesome too for those who rolled additional Friends or Acquaintances while making their characters and need to flesh them out via the provided tables in the Handbook.

PR #4224, opened 2018-09-20
The Fade Tabletop 2nd Edition Sheet

Jhar226:
I was working on the new sheet for The Fade Tabletop 2e. This is the final result.


Jhar226:
@Anduh File names for the html, css, and the internals of the json have been changed.

Anduh:
I'd change the preview name while you at it.

The Devs are doing the rounds on Monday and does merges on Tuesday, so you have a window there inbetween to fix it there appears some problems.

Jhar226:
Changes should be implemented now.

PR #4230, opened 2018-09-23
Feature/metal adventures

yuyuki:
Begin to add a CharacterSheet for Metal Adventures.

I want to try before going furthur


clevett:
Greetings @yuyuki 

I'm sorry to inform you this merge have been reverted. The sheet in app does not match the preview image and appears to be lacking a lot of basic functionality. You'll want to keep doing updates and submit it again when it is ready for access to the general public.

Thanks,
Cassie

PR #4243, opened 2018-09-25
Le Donjon de naheulbeuk 2.4.1

Kaias26:
Really sorry guys i made a little mistake in my previous PR Y_Y
updateWeapon and updateWeapons exist both so I had no visible error but right now, the wrong function is called during sheet:opened event and the weapons aren't updated and players see bad values.

### DIFF
--- a/Naheulbeuk/Naheulbeuk.html
+++ b/Naheulbeuk/Naheulbeuk.html
@@ -816,7 +816,7 @@
 			} else if( vrs < 2.4 ) {
 				// uptodate
 				updateSecondaryStats();
-				updateWeapon();
+				updateWeapons();
 				setAttrs( { "version" : 2.4 } );
 				return;
 			}


Anduh:
Stuff happens. Do you test the sheet by yourself, or do you have others who checks things out when you work on an update? It might be a good idea in the future as a second set of eye might spot bugs that you don't. Sure have saved me a time or two.

Kaias26:
Yes i've tested and asked for other players test.
With my pro account i create a new party with empty custom sheets and people can try the new version, but because they dont own pro account i can't let them try the update on their current game.

Sadly the bug only occur in already existing parties with already created sheet :
I've added a new attr_ witch is the sum of other existing stat (calculated by sheet worker on:change on those stats) but when you open an existing sheet, the new stat display 0 (because on:change is not triggered) wich look like a bug for players :s

this bugFix allow the sum to be calcultated on sheet:open if the attr_version < 2.4

Anduh:
Well, now we know a new parameter for with to test stuff in the future. 😄 

I've imported character from my other campaigns into my test games but that was because I didn't want to start filling a new one, but now I have a good reason testing on filled and new sheets.

PR #4246, opened 2018-09-25
Add "Vieja Escuela Cyberpunk" via upload

Zerel96:
Adding Vieja Escuela Cyberpunk Sheet to Character Sheets DB

It should work but i've didn't tested it on Roll20 because i'm not Pro Account

If it works, please, upload a preview on Roll20 on this folder and add it on the .json file


clevett:
Greetings @Zerel96 

Were you able to access a Pro account for testing? No pressure if you were not able to obtain one but we'll need to close the pull request without testing in Roll20 app.

Thanks,
Cassie

PR #4253, opened 2018-09-27
(Fate basico ES) new game

kinower:


kinower:
There are already the changes that you mentioned to me.

Thank you

kinower:
mmm, the problem is that I have not translated a Game already cloned,

https://github.com/Roll20/roll20-character-sheets/tree/master/Fate%20B%C3%A1sico%20BR

clevett:
Greetings,

These older sheets are from before we had a translation option. We now require games be [translated](https://github.com/Roll20/roll20-character-sheets#internationalization--translations) rather than cloned.

Thanks,
Cassie

kinower:
ok, so in order to translate this sheet, you would have to edit the original file of the sheet
https://github.com/Roll20/roll20-character-sheets/tree/master/Fate%20B%C3%A1sico%20BR
and add the translation.json file?

Anduh:
Yeah, thatcs the case. Apart from the wiki article you can find from clevett's links, it might be a good idea to take a look at one of the currently translated sheet & their translation.json so you get a hunch of how it's done.

[Vampires](https://github.com/Roll20/roll20-character-sheets/tree/master/CWOD-V20) is a faiely simple sheet that have been translated.

clevett:
Greetings @kinower 

I'm going to go ahead and close this pull request as its a clone of a current sheet. Translations keys can be added to the current Fate Basic sheet as an alternative to translation the HTML directly.

Thanks,
Cassie

PR #4262, opened 2018-09-29
Strain Sheet

CrossJustX:
Added 2 versions by accident


Anduh:
Yeah you need to remove either one. Otherwise you seem to have all critical things.

CrossJustX:
I've removed the incorrect files, it should just be the single one now.

joesinghaus:
Your Strain_Sheet.json file should be named "sheet.json".

PR #4269, opened 2018-10-01
Rough Draft Sheet for homebrew system Seven Deadly Sins

Kumatei:
Would like this in the repository so that my DM can use it =)


Kumatei:
Alright. Should I put the link to the rules in instructions? Or some other place in the json?

clevett:
Greetings @Kumatei 

You could link in the sheet.json if you wanted to but its not necessary. I just need a link in these comments so I can review the rule set & ensure the rules are available to the public in some sort of published format.

Thanks,
Cassie

Kumatei:
Alright, here's the rulebook. http://arsgothicus.com/rulebook.pdf

clevett:
Greetings @Kumatei 

FYI, your sheet.json was not named sheet.json. It is named sds.json which will cause the sheet to fail to load. I fixed this so the sheet could be merged and it is now live. You'll want to update your local copy.

Thanks,
Cassie

PR #4274, opened 2018-10-03
[D&D_BECM]

BiscuitCookies:
Fixed Skill rolls, ability rolls, saving throws to show the target for the roll and the actual roll.
Fixed attack rolls where they wouldn't work with negative modifiers
Also made the inventory auto-calculate totals of encumbrance for individual items and individual tabs.
Added rolls for turn undead in addition to the gm whisper
Changed two thief abilities to checkboxes since those are static %


BiscuitCookies:
I have a question regarding automation.
How far am I allowed to add automation to the sheet?
For example am I allowed to have experience needed and level automatically calculated based of which class is selected?
Or automatically input the saving throws depending on class and level?
Etc.

clevett:
Greetings @BiscuitCookies 

Automatic is a careful line to tread as you never want to add published rules that could result in you violating copyright. For example adding in a plethora of published content so the player could level themselves up would be a hard no. However doing some calculations based on inputs is perfectly acceptable. The examples you've suggested here are common levels of automation we allow.

Thanks,
Cassie

PR #4284, opened 2018-10-08
[Libreté] Add character sheet & [Headspace] Add character sheet

Sheershym:
Add character sheet


Sheershym:
Thanks ! I think that might be good now :)

clevett:
Greetings @Sheershym 

I had to revert this pull request in #4288 the Liberte sheet.json was causing my server not to load. You'll want to go over that file with a fine tooth comb to ensure its up to standards. Make everything lower case and rename files appropriate so the match exactly the entries in the sheet.json. The 'Librete.PNG' may have been what was causing the failure. 

Headspace was also removed since it was in the same pull request. You'll need to submit it again with Liberte.

Thanks,
Cassie

PR #4290, opened 2018-10-09
[Libreté][Headspace][Epyllion] Add files

Sheershym:


clevett:
@Sheershym 

Hello. I had to remove the Headspace sheet as it was failing to load in app. You may want to double check it by adding it to a table as a Custom game. 

As a best practice please submit the images to the sheet to GitHub. They should be hosted in the folder.

Thanks,
Cassie

PR #4291, opened 2018-10-09
Splittermond: Adding masteries

ap3h3ad:


ap3h3ad:
I don't know the difference, but I think the functionality inside eclipse did it.

Cheers.

clevett:
Its not a difference you can see looking at your sheet but the robots in our computers know the different. It does look like you  managed to fix it. :)

Thanks,
Cassie

PR #4299, opened 2018-10-10
V5: Fix extra space in roll template name

joesinghaus:


joesinghaus:
@clevett the most recent update apparently broke the roll templates for all the VtM 5th edition sheet. I took a quick look at it and it is easy to fix. I don't know if you can do anything about getting this live before next Tuesday.

Kirintale:
Good find.

Thank you.

PR #4302, opened 2018-10-12
add data-i18n tags to WFRP4, Quin and HollowEarth

dyvim:
I've added the tags so that a translation can be done after the merge for these 3 character sheets.

By the way there shouldn't be any modification in the DSA5 Character sheet by my side, I don't understand why conflicts appears.


dyvim:
By the way I'm not able to generate the translation.json files so if someone can do it, it would be great.

joesinghaus:
You did the translations wrong: it should be `data-i18n="xxx"`, not `data-i18n:"xxx"`. This is probably why you cannot genereate the translation files. BTW, with some bogus DSA5 changes in here, I doubt this will be accepted.

dyvim:
Thank you @joesinghaus 
I'll correct that and for the moment cancel this pull request.

PR #4303, opened 2018-10-12
Origin/master

falrun04:
I added a CSS file and updated the sheet to included roll buttons for different stats including attack and damage.

I also added a png file for previewing the sheet.


falrun04:
I've updated the sheet.json based on the sample file in the kitchensink directory and changed the LFs to be Unix style.


Thanks,

Paul


Sent from Outlook<http://aka.ms/weboutlook>

PR #4336, opened 2018-10-22
Metal Adventure : Move img in Github

yuyuki:
Fix

get img form github only.


yuyuki:
With pleasure.

Thank for your reactivity too.

clevett:
Hello @yuyuki 

FYI, we received an email from Github about a possible security vulnerability in the repo related to package-lock.json in the Metal Adventures folder. We went ahead and remove a few of the files in that folder which I assume are part of your build process. This came up as part of our weekly code push so we may have removed more files than we need just to in order to facility the process. You could add back the .gitignore file but the gulpfile, package-lock, and package should probably be kept locally.

Let me know if you have any questions. 

Thanks,
Cassie

yuyuki:
Hi @clevett ,

Thank to advice me. I will remove from « Metal Adventures » all unnecessary files.

PR #4319, opened 2018-10-16
add data-i18n tags in WFRPG4, Quin and Hollow Earth Expedition character sheets

dyvim:
allow a future translation via the generation of a translation.json file

PR #4343, opened 2018-10-25
Add translation.json files for #4319

joesinghaus:


clevett:
Thanks for helping with that project. :)

PR #4348, opened 2018-10-28
L5R Ring columns fix

olivier-grech:
## Changes

I made some small fixes to the CSS of the ring columns in the character sheet so they have a consistent height and bottom-padding.

### Before

before

### After

after

## Additional comments

Additionally, I ran js-beautify (https://www.js-beautify.com/) on the CSS file.

Note that I don't have a paid account on Roll20 so I couldn't test my changes. I manually edited the CSS and HTML from the Firefox console in order to do that.

PR #4349, opened 2018-10-28
L5R Use new ring symbols

olivier-grech:
## Changes

I updated the ring symbols, using more modern versions.

I think it looks better.

### Before

before

### After

after

## Additional comments

I do not have a paid Roll20 account, so I wasn't able to test this change.


clevett:
Greetings @olivier-grech 

Closing reasons started in #4348.

PR #4350, opened 2018-10-28
L5R Remove clan dropdown menu and use label instead

olivier-grech:
## Changes

There was a "Clan" dropdown menu with a list of major clans as well as a "Misc" option. I don't think this dropdown menu brings much to the player, as this value will only be written once anyway, and it prevents the user for using a custom clan (or a minor clan).

So I removed it and used a simple label instead, just like for "Family" and "School".

Additionally, I removed the colons that were present on "Clan:" and "Name:" but not on "Family" and "School".


clevett:
Greetings @olivier-grech 

Closing reasons started in #4348.

PR #4363, opened 2018-10-30
ChroniquesGalactiques: new version

Ulty:
*ChroniquesGalactiques.*

## Changes

Reorganization of the sheet, taking more game options into account.

There is one slight change for superior characteristics, but we added automatic update to the new version.


clevett:
Thanks for the comment. That was helpful. :)

Ulty:
@clevett : I don't get it. You seemed to approve my changes, and then you reverted the commit. What was wrong with it?

Ulty:
Ok, I read your comments in the revert (they were not sent to me, so I hadn't seen them before). I proposed a new pull request with hopefully corrected sheet.json.

clevett:
Sorry for the confusion. Usually when I have to revert changes like this I'm in crunch mode trying to get all the sheets ready to go just before the  deploy. I try to @ the author but I must have missed it here. I'll review the new pull request when its up. :)

PR #4366, opened 2018-10-31
Forbidden Lands character sheet

darkwuulf:
Initial upload of character sheet for Forbidden Lands campaign setting.

## Changes

No changes. Initial file upload.


darkwuulf:
Thank you for the suggestion, this is my first sheet and I vaguely remember reading about the i18n, but was focused on creating a functional sheet first.

darkwuulf:
Have a few fixes to push, will add the sheet image correction.

Anduh:
@clevett Replace what with what?

Think you wanna wrap you text parts in code blocks,  GitHub&Markdown ate up part of your comment...

darkwuulf:
Someone is going to have to show me a reference in my code to what this means. I do not understand.

Anduh:
Yeah as I said, anything like that isn't mentioned anywhere. This shouldn't be a hinder IMO, shouldn't anyway make a difference, exept for maybe code visualisation. I can admit that I have some code I've inherited that even have a mix and match, which is objectively worse.

There could be made added a .gitattribute file that changes /4spaces=tab at each new line or something, if we want to make this a thing a thing problem, but I have a suspicion it would have some annoying results in existing files. I personally use tab in the things I've participated in, but apparently spaces are more common among the most popular GitHub projects.

A good IDE can make the change between the two automaticly, but this haven't come up for me yet so haven't looked how NP++ does it(I'm using it as my my primary tool also.)

Anyway, Mondays are for review and pre-approval for PRs(and merging clear cases it seems), and Tuesdays are the actual day the rest of the PRs are merged. We'll get an official word in the coming hours.

clevett:
Greetings

I'm sorry. Indeed I forgot the `` so my code was eaten. You'll want to replace the tables with divs. Currently a ```<table>``` tags are being used as a core building structure in the html which divs would be suited better for. This the [minimum requirements ](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) encourage the use of 'Proper HTML Syntax'.

Thanks,
Cassie

darkwuulf:
Just to clarify then, existing sheets are grandfathered but new ones will absolutely not be allowed to use <table> anymore?

clevett:
Older sheets often did use ```<table>``` to layout sheets but this is less maintainable way of coding. The ```<table>``` element may still be used when displaying [tabular data](https://www.w3schools.com/html/html_tables.asp) however its not intended to be used for creating a basic html layout.

clevett:
Greetings,

As this pull request is now 20 days old since changes were requested, I'm going to go ahead and close it for now. Feel free to make a new one when you are ready. :)

Thanks,
Cassie

PR #4369, opened 2018-10-31
[Quin] add missing terms to translate

dyvim:
linked to pull request #4368 "correction of a bug in Quin character sheet"

add missing terms to translate that have been added by the correction of the bug in pull request #4368


Qin Warring States/translation.json
```diff
@@ -89,4 +89,32 @@
   "Wood-button": "Wood",
   "Yan": "Yan",
   "Zhao": "Zhao"
```
joesinghaus:
```suggestion
  "Zhao": "Zhao",
```
Comma is missing here to turn this into valid JSON.

PR #4385, opened 2018-11-05
Starfinder Official No Version Change

Kurohyou:
# Starfinder Official No Version Change
## Bug Fixes
- Fixed parsing of Buffs to handle attributes with more than one _ using spaces instead of _'s


G-G-G:
Is the starfinder.html file supposed to be deleted? Won't this cause issues with the sheet?

PR #4412, opened 2018-11-12
[Low Fantasy Gaming] Sheet Overhaul

dedurrett:
Expanded fields and features on the sheet to improve experience and compatibility with Low Fantasy Gaming system. Will convert from old sheet.

## Changes
Utilized a lot of the 5E Sheet that had been commented out to extend the sheet, and made it compatible with LFG where needed.
Added fields and features unique to LFG, and roll buttons where appropriate, extending the existing roll template.
Fixed several errors in attribute naming, but added version checking for bringing the old attribute data forward.

## Additional comments
Blessing received from old sheet creator (also creator of LFG), via MeWe LFG discussion page.


dedurrett:
Oh, bummer. I just use the editor on my roll20 campaign settings page, and then copy and paste it to github! Well, I think I fixed it. Just the html, right? Only changing the line feeds was giving me an error on commit, so I also just cleaned up an old comment, that seemed to do the trick.

clevett:
Interesting. Let me check it when I merge it in. It may be fine and just look the way it does since you copy/paste it from the Campaign Settings page. Typically when it looks like a brand new .html file its because the line endings may be wrong but that is not likely the issue now that I know your process.

PR #4417, opened 2018-11-13
Splittermond: Added some requested convenience features

ap3h3ad:
Inlcuding:
- Additional dice throws per melee abilites
- Additional dice throws per magical schools
- Taking into account the defensive capabilities of the PCs weapons while defending & styling the rolls
- Taking into account the abilitiy to heal / regenerate faster
- Added an option field to allow players to whisper all dice rolls to the GM
- Making sure that the new abilityroll is not empty when the sheet is opened
- Adding an information in the roll template if one of the dice shows a 1 or a 10
- Improving the interface to calculate LP and Mana loss (transforming checkboxes into button-like checkboxes)
- Extending roll macro for better spell support


clevett:
Thanks for the detailed comment. It was very helpful

PR #4418, opened 2018-11-16
(HITOS) Update hitos.html

kinower:
## Changes
-changed an error from a calculation of a skill
-changed the size of some boxes that did not show the numbers well


Anduh:
Soo what did you change? If you read the default message and folow it, Cassie knows faster what to look for in you changes.

kinower:
thanks!. Eres muy amable! ^-^

PR #4437, opened 2018-11-21
Scion 2nd Edition

salmelo:
New sheet for second edition of Scion.

## Additional comments

Plan to add i18n support eventually, but don't know how long it'll take me and wanted to get it out there in the meantime. I'm also exploring the possibility of adding roll buttons, but not sure if there's a good way to do it. Fortunately the system is simple enough for a generic macro to suffice, so hopefully they won't be too terribly missed.


Anduh:
I ran your sheet through a WIP script I've been working on for automating i18n-tag creations, and[ here is a passable first version](https://www.dropbox.com/sh/ejmchepsqw7plqm/AACW5Sdz0fQHvoitOOAbB434a?dl=0)

Had to do minor tweaks and use roll20 to generate the translation.json file, but imo a passable result from 15min of work.

salmelo:
The skills look a bit borked and the placeholder's need their translation attributes added, but other than that looks like a great starting point.

Anduh:
Yeah there is a reason it's still Version 0.1 :smile:

PR #4451, opened 2018-11-26
Splittermond: Small convenience functions

ap3h3ad:
- Adding the possibility to roll on distance fight spells via the interface for ranged weapons
- Improving the adding of fight spells with auto completion function for the necessary attributes
- Increased the size of some fonts for besser readability
- Adjusted the old credits, versioning and formatting


clevett:
Greetings @kreuvf 

It would be up to the contributor to squash all commits at this time. I generally do not monitor the commit histories but merely the changes to the files in the repo. My personal PRs even have this sort of commit histories. :( I will start trying to squash my own and see if we can come up with an easy process to recommend to people.

Thanks,
Cassie

kreuvf:
Thanks for the quick reply @clevett 

I understand the issue. New branches constructed using `cherry-pick` would work, but probably not from GitHub's web interface :(

Since you (= Roll20) are usually emptying the PR queue ~once a week, would it be possible for Roll20 to just put all new PRs in a (to be fast-forwarded) branch and by using `rebase --interactive` just remove unwanted/accidental commits? After rebasing, fast-forward merge the branch.

ap3h3ad:
I'm sorry for meddling in the history of commits. As the PR clearly states there is only 1 file changed and that is my own sheet. Therefore I was under the impression that there would be no changes to the DSA sheet, since they were reverted directly.

kreuvf:
Well, that's Git(Hub) for you: Showing you that no files were touched, although they were. The files were not changed, but I noticed that your PR in my fork of the repo fails right now and when I investigated why I noticed this. But, that does not matter much as I will rebase stuff locally anyway, so I am not blaming you ^_^ In Git `revert` does not do the same as the undo function in many desktop applications. Instead, it records a new commit that does the opposite of the commit you wanted to revert. So instead of *one* unwanted commit in your history, you now have *two*. If you like I can show you how to fix this before sending the PR (open an issue in my fork of this repo).

I was just wondering whether it might be feasible for Roll20 to do something to prevent this in the future. Maybe some sort of automated checking? GitHub started offering so-called "checks" some time ago. While I haven't look at that, it sounds like a good option to me. :)

PR #4471, opened 2018-12-03
Update to Ars_Magica_5th.html

Medieve:
- Added default values to combat macros to fix formulas not adding in wound and fatigue penalties. 
- Minor change, ordering of numbers to make it easier to check the numbers on float over the result.


Medieve:
Not sure what went weird but I guess my changes were being picked up as a conflict? I resolved it tho.

clevett:
Thanks for fixing those.

PR #4475, opened 2018-12-04
[CWOD V20] Remade entire character sheet added a ton of features that other 20th ann. sheets have

RinVindor:
## Changes

-Added 12 tabs of info for the sheet enabling vastly more organized and indepth data about the characters to be stored
-Added a more robust dice rolling section that gives more customization

## Additional comments

This does mean that character sheets for folks will be messed up but the end result after the change is a much better experience than before

## Roll20 Requests

- If I linked every feature request I would honestly be here for a number of hours hunting them down on the Roll20 forums. But to name a few, a separate dice rolling section, a separate paths and rituals page, a note page, another for equipment and gear and of course background tabs as well as boons and general player info
- This pull request should only edit CWOD V20 if there is anything else please ignore it.
- The changes made more or less necessitated so many changes I was not able to assist players in keeping their data, but I firmly believe the changes will enhance every players issues with the previous sheet which had gone without updates for years now. With these changes I don't think we'll need to see too many more alterations to the sheet besides more Quality of Life improvements like perhaps some dice functions nested into the Paths/Rituals page and maybe the Equipment page.


clevett:
Greetings @RinVindor 

Thanks for the detailed note! I see this a pretty big change that may effect a user's data. With that we'll want to give people a heads up. Would you please make a post on the [Character Sheets](https://app.roll20.net/forum/category/277980) forum to let people know you're doing a larger haul to this character sheet and people should back up their data? Let them know it will go live on the 18th December.  :)

Thanks,
Cassie

RinVindor:
Hey Cassie,

Yeah I'll make a post now that I've seen this :)

clevett:
I saw the post. Thanks for that. :)

PR #4529, opened 2018-12-18
Kult Divinity Lost - Fixing some minor visual issues

Rats808:
In the first version of the sheet, I seem to have inappropriately referenced the src for an img tag,, and my attempt to use padding to set a table's width didn't work out.

This commit appends the appropriate src, and adds an empty column in the table to (hopefully) fix the width.


clevett:
It takes a bit of work but replacing the `<table>` and using `<div>` will help with the width issues. Tables come with styling baggage that Divs do not have. :)

PR #4535, opened 2018-12-20
The Veil Translations

clevett:
## Changes / Comments

* Several of The Veil translation files did not generate.
* Wrapped in ES


clevett:
No longer needed.

PR #4549, opened 2018-12-26
[CWOD V20] Reverted recent massive changes to the sheet

pdusen:
## Changes

- Reverted the complete overhaul of the character sheet that was merged on December 17th

## Additional comments

I'm requesting these changes be reverted for the following reasons:
- There are some formatting issues with the altered sheet that remain unresolved
- There are some unresolved bugs with the altered sheet that remain unresolved
- The author of these changes is no longer able to help resolve either of the two above points. See his post here: https://app.roll20.net/forum/permalink/7062860/
- Making a sudden, massive overhaul to an existing template that is being actively used in games should be considered unacceptable and should not have been merged in the first place, particularly when no option is available for ongoing games to decide whether or not to opt into an update that could potentially break existing characters.
- In light of the previous point, the changes should have been implemented in the form of a new sheet instead of replacing an existing one.

## Roll20 Requests

There is no new work in this commit, only a revert to changes previously made to the CWOD V20 character sheet.


Anduh:
Those are some good points. I feel this will likely generate some discussion so I'd suggest to start a thread about in on the forums where more people could join in on the pros and cons regarding doing a revert or not. No changes to sheet will be pushed in yet another 1½ weeks so there is room to talk this out. I don't have an opinion in either direction, nor do I use this sheet, so I won't be commenting any further if possible.

@RinVindor might be interesting in responding to this.

pdusen:
Thanks @Anduh for bringing this up in that thread, I'll make a comment there summarizing my points as well.

clevett:
Greetings,

RinVindor,  the author who original made the changes, has been actively work on bug fixes. I am going to close this request as they are doing their best to get fixes in. I'm sorry for the issues this may have caused and we are using it as a case to look at ways to help us better manage large community projects such as this one. Work around for some bugs have been [posted](https://github.com/Roll20/roll20-character-sheets/issues/4565) and hopefully with support RinVindor will be able to iron them all out.

Thank you,
Cassie

con-f-use:
Not good enough. Please revert the changes for now. The sheets are unusable as is!

PR #4552, opened 2018-12-29
Correct d20 modern to allow bars

kayila:
## Changes / Comments

Modifies attr_fx-Known and attr_fx-PP to have a 'max' value instead of 'max' and 'current', thereby allowing for a PP bar on token.

## Roll20 Requests
* Renamed `attr_fx-Max_Known` to `attr_fx-Known_max`
* Renamed `attr_fx-PP_Max` to `attr_fx-PP_max`
* Renamed `attr_fx-PP-Current` to `attr_fx-PP`


chanaldebol:
Yeah... I've run into that problem. I had to have my GM go back to the previous map to remember what my current PP was supposed to be. Not having the bar on the token bound to the character sheet makes it harder to keep the number on the sheet in sync.

PR #4555, opened 2019-01-02
Celerity Sheet: Changed AC and other title alignment

Aetis:
Changed AC and other title alignment

## Changes / Comments

Changed the alignment of the AC and other titles slightly.

## Roll20 Requests

Celerity sheet.


Aetis:
Cassie, when does the merge go live in roll20?

clevett:
Greetings @Aetis 

Changes usually go live in about the next hour or so. Depends on if we run into any complications. :) Just short of catastrophic failure they will be up in 2-3 hours.

Aetis:
Great!

PR #4617, opened 2019-01-22
[CthulhuTech_1ed] Fix Agility math to correctly roll AGI Feat

KasaiMagi:
Dice were counted by raw Agi instead of `floor(Agi / 2)` like all other attribute feat rolls.

## Changes / Comments

*Provide a few comments about what you have changed.*
AGI Feat was changed to calculate number of dice equal to raw AGI. This is incorrect and does not match how the other attribute feats are calculated. Dice count should equal `floor(AGI / 2)`

@CoalPoweredPuppet was your previous AGI calculation in https://github.com/Roll20/roll20-character-sheets/pull/4042 a mistake?

## Roll20 Requests

### DIFF
--- a/CthulhuTech_1ed/CthulhuTech_1_ed.html
+++ b/CthulhuTech_1ed/CthulhuTech_1_ed.html
@@ -252,7 +252,7 @@
                                     <td>Feat</td>
                                 </tr>
                                 <tr>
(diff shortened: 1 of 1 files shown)


CoalPoweredPuppet:
...looks that way. I haven't look at it in a while.  So...yah for the fix?

CoalPoweredPuppet:
Thank you for doing this.

PR #4621, opened 2019-01-22
Enhancements to the AS&SH Character Sheet (

romelwell:
## Changes / Comments
Add attack and damage dice rollers to the Melee and Missile weapons section.  The result should show the target AC hit and damage caused.
Change Saving Throws from a single dice roller to a dice roller per saving throw type. 
Add a dice roller into the spell section. Not all spells need it, but it'd be nice if a roller could return the spell name and a damage or duration roll that is input.
Change the "Gems and Jewelry" section of the Coinage to a repeating section with fields for Type, Value, Quantity.


clevett:
Greetings @romelwell ,

Can you double check your line endings for me? You will want them to be LF (Unix). How to do this depends on your text editor but a quick google search should provide you with directions.

Thanks,
Cassie

romelwell:
> Greetings @romelwell ,
> 
> Can you double check your line endings for me? You will want them to be LF (Unix). How to do this depends on your text editor but a quick google search should provide you with directions.
> 
> Thanks,
> Cassie

Good day Cassie,

I have updated line endings for both files from CRLF to LF as requested.

clevett:
@romelwell 

I think you save the new css & html files in the wrong place. :) They need to be in their "AS&SH" folder. The new pull request has them going directly into the roll20-character-sheets folder.

romelwell:
@clevett - I'm terribly sorry (and embarrassed) by my previous mistake.  The HTML and CSS files have been updated as requested and committed to the AS&SH subdir.  Thank you for your assistance and patience.

clevett:
Not a problem. :) This looks much improved.

PR #4622, opened 2019-01-23
Splittermond: Cleaning Code & Adding attribute

ap3h3ad:
I added an attribute to allow playsers to reduce the constriction delivered by the character's armor, because of some new fancy abilities allowing this.


ap3h3ad:
In the last PR (https://github.com/Roll20/roll20-character-sheets/pull/4601) I changed the URL of the RPG Logo from an extern resource (http://) to a local ("Splittermond.png") by adding the logo directly to the repository. This doesn't seem to work. 

Old (working) code:
</img>

New (not working) code:
</img>

I don't see why this wouldn't work, except the local files are addressed differently than I thought.

Could you have a look?

Greetings

Anduh:
Sure looks like that should work, but [refering to the direct github url](https://github.com/Roll20/roll20-character-sheets/blob/0ae7a41db3ddb4f1e0ac50586c9c32d5743e69e1/CWOD-C20/C20.html#L1) is a way I know works, and nobody had commented on as improper so far.

clevett:
Greetings @ap3h3ad ,

Sorry I was so slow to respond but it looks like you added the fix  would have suggested. :)

PR #4624, opened 2019-01-24
[Blades in the Dark] Add support for Spanish translation

joesinghaus:
## Changes / Comments

* Add support for the Spanish translation.


WallaceMcGregor:
I've made some rolls to try the Spanish translation and it works fine. But it has some minor details that I don't know if can be solved.

The Engagement (Implicación) has a line break of one letter

bitd roll bug 1

The line break of Cohort Quality (Calidad de la Cohorte) overflow the text below.

bitd roll bug 2

Could be possible to show the Kristy font on the chat rolls? I have the font installed to make this rolls.

Anyway. Thanks for all the work you've made ;)

joesinghaus:
Hey @WallaceMcGregor,
I can fix the engagement roll. Cohort Quality is a bit harder, perhaps it is necessary to change the translation to "calidad" only. The font would have to be really tiny to fit here.
I can see if I can make the Kirsty font show if the toggle on the sheet is on. It's not really a good idea by default, since it will look pretty bad if Kirsty is not installed due to how differently Georgia and Kirsty look. No promises here.

WallaceMcGregor:
Okay. You know better how can be shown this. I understand the font problem so I leave it to your decision.

Do you need I change the translation of the Cohort Quality or can you do it?

Again, thanks for all ;)

clevett:
Could you do CSS just targeting the roll template when the sheet is in Spanish? Perhaps increasing the margin or decreasing the font size slightly.

WallaceMcGregor:
I have limited knowlege of CSS, but I tried this things:

- Changing the line-height of the Cohort Quality from 24px to 20px:
css bitd1

- Adding a margin-bottom to Cohort Quality of  20px:
css bitd0

- Decrease the font-size of the Engagement roll from 32px to 30px:
css bitd2


I am sure there is betters solutions.

PR #4661, opened 2019-02-04
[LANCER RPG] Added 1.8-compatible sheet, separated 1.7.5 sheet into its own folder

spingle:
## Changes / Comments

Added an overhauled LANCER 1.8-compatible character sheet. The old sheet has been re-labeled as (1.7.5) to prevent confusion. (Changes to both sheets are intentional).


spingle:
oops, fixed! thank you!

spingle:
hey, thanks for letting me know. to be entirely clear, the different versions of the sheet don't just use different code, they're also meant to be compatible with different release versions of the LANCER RPG system itself. the game is still in active development and was recently updated to version 1.8, which includes a significant amount of mechanics changes that would render the 1.7.5 version incompatible. some players have active campaigns still running off LANCER 1.7.5. i wanted to provide support for them via a legacy version of the sheet.

clevett:
Perfect. This is what we'll do. Leave the current Lancer folder as it is. Make a new folder called LANCER RPG 1.8 and submit that to me. I'll deprecate the old sheet so old games can use it but new games cannot select it in the drop down. This way we'll slowly phase it out.

Thanks,
Cassie

spingle:
changes pushed as requested. thank you for being so understanding, the LANCER community and very much appreciate it.

spingle:
pr submitted as requested. thanks for being so understanding; the LANCER community and i both appreciate it a lot.

PR #4672, opened 2019-02-05
[TheWitcher] QoL changes, minor fixes

ancallan:
## Changes / Comments

- Added a checkbox to switch between a character-creation-style skill total and default.
   - This is to help in creating characters, allowing a player to total their skill points based on cost. The checkbox allows the player to select their preferred method of totaling.
- Death state now applies at zero as well as below zero
   - Some incosistency in the rulebook on this. 
      - pg48 refers to HP as the 'number of points of damage before you go into Death State'.
      - pg43 'Tough As Nails' refers to falling to 0 to result in a Death State normally.
      - pg162 refers to a Death State as occuring when you are 'knocked below' 0.
- Current Stun is now also auto-calculated based on primary stats.
- Split Death&Stun save rolls into two buttons. Death saves uses unmodified Stun, Stun saves uses current stun.
   - This is done primarily to catch edge cases where a stun roll may be required while the character is beneath Wound Threshold, but above Death State.
- Low stats are now prevented from dropping to 0 when a death state is incurred.
- Default values for statistics and main profession skill now set to 1 by default.
   - More for the assistance in creating characters, as these values should always have at least one point invested.


Leothedino:
Sent you a Reddit message @ancallan but insane you see this first I will repeat two issues found that really should be addressed before this is pushed live:

**Current Stun:** The ability to manually change 'current' stun has been disabled. Prior to your changes this allowed players to change their current stun score based on status effects inflicted upon their characters. I'd recommend either reinstating this or please provide a query box on rolling the STUN save feature. 

**Magic & Spells Layout Error:** There is damage to the 4th focus 'show/hide' button. An arrow should be here that allows the contents of this feature to be toggled as desired. I am afraid I don't have the time to hunt down what has happened here before pull-requests are approved, I am hoping you will.

ancallan:
> I admit, it's not down to me whether these changes are approved or not, but I advise that until they are addressed and fixed, I wouldn't push these latest changes to live.

You're also the one who would likely be dealing with the reddit messages, so that's an easy change to make. ;)

ancallan:
**Current Stun:** I definitely like your feedback here, I hadn't considered modifications based on critical wounds. This is one that was requested after some playtesting, the players felt as though it was too fiddly to have to change that stat based on changing circumstances that might vary quite drastically even within a single round (such as the modifiers from a heavy bludgeon weapon). In the end, they mostly used /roll with the sheet as a reference. As I think on it, if split up as I have it here, there would need to be a need to track modifiers for Death Saves too, as they accumulate greater penalties each round. It's almost as if there's a need for a specific count for the Death Save -- though I feel like there's also a point where it might be getting over-engineered!

In the end, I feel like Stun probably needs a bit of careful nuance in how it's handled. I'll change the way it works back to how it used to be and give this one some more thought.

**Magic & Spells Layout Error:** This issue appears to have been there prior to my changes. The .css file was missing 'focus4' elements that the .html file was referencing. Good catch though, I've fixed that one!

Leothedino:
@ancallan been through the fixes, and I can confirm everything is good to go and working! **When Cassie comes to approve this, for whatever it is worth, I give it my full thumbs up.** 👍 

In regards to the Stun & Death save (and future updates in general), if there is a way to make the processes on the sheet easier to use then we should strive to do so, but only if they're tried, tested and we're certain it'll improve peoples experience. I can't insist that prior communication be made to me on Reddit before pushing that level of change in the future, but would greatly appreciate it it was. I understand the good place you're coming from with these QoL changes and the effort you put in (coding is tough, at least for an amateur like me 😄), I just want to make sure what we add/change doesn't disrupt or upset all the ongoing games being played around the world.

Thanks for the contribution, for your time and for making the fixes. I look forward to seeing this update go live.

clevett:
@Leothedino 

Helpful peer reviews are always appreciated. :)

PR #4637, opened 2019-01-29
Add a new Savage Worlds sheet

olivier-grech:
This is a new Savage Worlds sheet. The existing one was a bit too busy for my taste, so I made this one.

This sheet has no script or auto-roll buttons, as I am still inexperienced with this. However, I have taken great care in making sure it is pleasant to look at and easy to use.

I also tried to stick to the best practices described in the wiki:
- Attribute names are in lowercase
- There is only one `!important` in the CSS (I don't think it is possible to do without it)
- I didn't repeat `-sheet` for the CSS classes in the HTML file
- I used ``spellcheck="false"`` for `textareas`
- No inline styles
- I used i18n to make sure the sheet is fully translatable

What could be improved:
- Obviously add sheet workers and auto-roll buttons, though I believe given how the sheet is made, it will be tricky
- The skull icon is hosted on imgur, and is not present in the files because I am not sure how to do that
- A minimum `width` should be set

I am concerned about making this sheet as good as possible (though I don't want it to be a scripted monstrosity with 3 tons of "shortcuts" that only confuse the player), so please don't hesitate to give your feedback.

PR #4674, opened 2019-02-05
Add a new Savage Worlds sheet

olivier-grech:
Following this MR: https://github.com/Roll20/roll20-character-sheets/pull/4637

I completed my Savage Worlds sheet.

I added roll buttons for attributes, skills, and weapons.

I also added a selector that allow you to chose which setting you are using so the derived attribute specific to that setting appears on the sheet (for example if you select "Deadlands", the "Grit" attribute will appear on the sheet).

I fixed the issue with the RuneQuest images.


clevett:
Greetings @olivier-grech 

You're still off to a good start yet we still need the sheet to be capable of all the current Savage World sheet that this one is looking to replace. In particular the Skill List, Quick Macros, Derived Stats, additional inputs for settings...As the Savage Worlds system as three sheet we currently are not accepting new ones but only potential upgrades/replacements.

I'm going to close this pull request so you can keep working on your branch without it emailing notifications to all the GitHub authors. If you have any questions just let me know.

Thanks,
Cassie

PR #4700, opened 2019-02-13
Splittermond: Image colour changed

ap3h3ad:
I changed the color of the small image to calculate mana costs for better visibility


ap3h3ad:
@clevett Is it possible to request an emergence merge? I forgot to delete some debug info that now will overwrite the precious "notes" text field of the characters. I fixed it with commit 4ff6ad7a and would be very glad if you could push the change outside the regular turn.

Thanks,

ap3h3ad

clevett:
@ap3h3ad 

Lucky for you we have some hot fixes going out today so I can tag this onto them. :)

PR #4710, opened 2019-02-15
Corrected a bug in the UESTRPG sheet that caused some classes to be assigned the wrong hit dice

EnricoDandolo1:
(Hopefully.)


Anduh:
You have some conflict in your .html file

EnricoDandolo1:
Yeah, that's the bug fix. There was a leftover class name that wasn't properly changed. Is there some way for me to mark the conflict as resolved? The updated commit should replace the current master.

PR #4712, opened 2019-02-15
.gitattribute & template update

clevett:
## Changes / Comments

- Remove a line in the template
- Enforce LF line endings for specific file types

## Roll20 Requests

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,10 +2,10 @@
 text auto
 
 # Ensure LF line endings for these files
-*.css	eol =lf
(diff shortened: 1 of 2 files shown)


Anduh:
Would it be good to also add `.md` and `.markdown` too, in case people adding Readme files with their sheets?

clevett:
I've not seen a case where a .md or .markdown has caused a problem. I'm looking to mostly enforce LF endings on these extensions as our servers will some times complain if they are not LF.

PR #4727, opened 2019-02-18
Cyberpunk 2020 - i18n localization code added and Italian translation

vince-roll20:
- i18n tags added to the sheet (with help from Andreas 1223200 localization script)
- translation.json created
- italian translation included it.json (thanks to Salerno T. 2547667)
- few small changes to sheet layout to help with localization changes
- added a new roll template "&{template:general}" and updated the ability rolls so they could also be translated.


clevett:
Greetings,

If Salerno is interested in translating other sheets let me know and we'll get them setup on Crowdin. Its a bit easier to do translations there than on the .jsons. :)

vince-roll20:
I've assisted a handful of translators while creating the initial localization code and first translation for a given sheet.  Normally I create a test game on roll20 which includes an editable copy of the translation json and we work together ironing out any adjustments that may come up during the initial translation. such as missed keys, proper context for a translation, and layout.  

Of course once this initial process is completed, any future translations "should" be much easier and is often done without a sheet author's knowledge and/or assistance through crowdin.  

I'm all for making the process easier and faster.  Out of curiosity and ignorance, how exactly is the crowdin process better/easier? 

Thanks

clevett:
Crowdin has a nice GUI for the user. You can see all the character sheets that have a translation.json and what % of that sheet has been translated to your language. It will also suggest common translations for a word such as if a character sheet has **Agility** and you're translating into French it will suggest **Agilité** since this is the most commonly used translation. You can just click that option rather than doing extra typing.

Translators can also leave comments for each other and ask for Context. Context is semi useful currently since sheet authors are often not in Crowdin and will not see these message. Though I try to pass it along or answer them when I can.

Your method has merits as well since it lets you do the keys, css adjustments, and testing all in one go. With Crowdin unless sheet authors are looking for it or communicating with their translators they may not realize their sheet has been translated thus not then making appropriate CSS accommodations.

vince-roll20:
Thanks for the info.  That sounds sweet.  Salerno already has the "Translator" tag, so they could already be a Crowdin member.  I'll pass the into on just the same.  Cheers

clevett:
Perfect. Yes, likely they are already there.

PR #4731, opened 2019-02-19
Create Werewolf the Forsaken 2e Sheet

Rats808:
Adding a Werewolf the Forsaken 2e sheet.


clevett:
Greetings @Rats808 

This is sheet a replacement for the old [NWOD Werewolf](https://github.com/Roll20/roll20-character-sheets/tree/master/NWOD-Werewolf) sheet? I have looked at your  sheet  and it  appears to be a  higher quality than the current NWOD Werewolf sheet. I just want to  verify they are the same system.

Thanks,
Cassie

Rats808:
Hey, @clevett 

The current NWOD Werewolf sheet appears to have been made between first and second edition, simply adding a Beat tracker for second edition usage(as well as a couple of terms that come from oWoD Werewolf). For the most part, though, it seems best fit for first edition Werewolf the Forsaken.

I wasn't sure if I should make this as a replacement for it or not, since it's likely still in use, and I wanted to make a sheet more in line with the official sheet for second edition, thus re-arranging and adding/removing some fields. In my opinion, the current sheet could be left in, for usage in first edition games, although it does need to be cleaned up a bit overall(mostly fixing up the CSS, and removing the beat tracker if it is re-purposed for first edition).

Thanks,
Rats

clevett:
Greetings @Rats808 

When we replace a sheet we deprecate it so games currently using it will not be affected but it no longer appears in the drop down so new games cannot select it. This lets us update sheets slowly over time without effecting current games. 

We generally try to keep sheet per system down to one or two depending on popularity of the system. There are a number of options out there for NWOD Werewolf so replacing a current sheet would be preferred to adding a completely new one. Currently the Official NWOD sheet supports Werewolf The Forsaken 1sts & 2nd Edition. Then we also have the sheet I linked above. Your sheet has improved CSS styling so if it will work as a replacement 2nd Edition, I think we should replace the current community sheet. 

I'm still getting up to speed on all my NWOD editions and just wanted to verify with you that this is a potential replacement for the current community sheet. :)

Thanks,
Cassie

Rats808:
Hey, @clevett 

In that case, yes, it could replace the current sheet. I briefly forgot the Official NWOD sheet existed, but it works fine as a sheet for first edition, as far as I'm aware.

Thanks,
Rats

Rats808:
Thank you, apologies for the mistake.

clevett:
Not a problem. :) There is a lot of detail to keep track of on a first push.

PR #4740, opened 2019-02-20
Call_of_Cthulhu_7th_Ed. Updating translation.json

orphs:
## Changes / Comments

Added Translations for Headers: Bouts of Madness and Dice Rolls , 
and Placeholders: Stable and Custom.


Anduh:
Does the tags already exist on the html sheet? Because they need to be there too for this new tags to function

orphs:
Yes, the already are on the new sheet that was recently merged. But due to me thinking that I should update the en.json instead of the translation.json, the translations I commited were not essentially merged.

PR #4742, opened 2019-02-20
GURPS Character Sheet

MadCoder253:
## Changes / Comments

**files changed**
gurps.html
gurps.css
sheet.json

**Changes**
- Fixed rounding bug with flight move. Updated Air move column to hide enhanced move if there's no enhanced move.
- Added condition rolls: Stun, Knockdown, Unconscious, and Death Checks.
- [Experimental] Added default sheet options. Can only be tested after deployed.


clevett:
Greetings @MadCoder253 

> * [Experimental] Added default sheet options. Can only be tested after deployed

 What's the expected behaviors? I can test this on my local after I merge the code.

> If you are a mentor, you can use this script to automatically determine success and failure.

By no means critical but on a future update you may want to change **mentor** in the sheet.json to Pro. We renamed them sometime ago. :)

Thanks,
Cassie

PR #4766, opened 2019-02-26
Updating Wraith the Oblivion 20th Anniversary Edition

Rats808:
Updating the character sheet for Wraith the Oblivion 20th Anniversary Edition, to swap out the healthboxes(from dropdowns to radio buttons, like on the Official NWOD sheet and the Werewolf the Forsaken 2e sheet), swap Dodge out for Persuasion(I put Dodge in based on an early copy of the PDF, but the final version of it has Persuasion instead.

Part of this update, in order to keep players from losing their Dodge scores, is a red-bordered box at the top of the sheet that can be removed with a checkbox. There *is* an attribute-backed span in the text here, to preserve their old Dodge score for reference sake, but I'm unsure if I implemented it correctly.


Rats808:
I messed up with the original PR somehow and it was trying to also include the commits for the new Werewolf sheet I already submitted. I'm not sure how but I guess I fixed it, because I eliminated the conflicts.

Anduh:
If you want to work on several sheets at the same time, it's best to have separate branches so you can more easily have two different works on progress at the same time, as well as not stopping the change of rone sheet being approved of the other sheet have problems.

Rats808:
I made this PR after the Werewolf sheet's PR got merged, though, which is why I'm confused about it including the commits in this one.

Anduh:
It's likely because you haven't updated your own copy of the sheet repo after this week's sheet updates where merged.

Anduh:
https://github.com/KirstieJane/STEMMRoleModels/wiki/Syncing-your-fork-to-the-original-repository-via-the-browser if you use the browser

Rats808:
Thanks, I'll keep all that in mind for next time. As of now, it looks like my fork is up to date with the roll20 master.

PR #4770, opened 2019-02-27
[Shaan Renaissance] Magic tab positioning update

Phoxounet:
Updated css checkbox trick positioning  in magic tab for better rendering.


clevett:
Greetings @Phoxounet 

Just a note on css that would have made your task easier you can combine entries that have the exact same settings.

For example these three entries all have the same customizations. : 

```
/* Hide actual checkbox */
input.sheet-magie-portee-distance
{
    opacity: 0;
	margin: -200px 0 0 0;		
    width: 193px;
    height: 201px;
    cursor: pointer;
    z-index: 1;
}

/* Hide actual checkbox */
input.sheet-magie-portee-horizon
{
    opacity: 0;
	margin: -200px 0 0 0;		
    width: 193px;
    height: 201px;
    cursor: pointer;
    z-index: 1;
}

/* Hide actual checkbox */
input.sheet-magie-1cible
{
    opacity: 0;
	margin: -200px 0 0 0;		
    width: 193px;
    height: 201px;
    cursor: pointer;
    z-index: 1;
}
```

Rather than maintaining three blocks of code you could do this:

```
/* Hide actual checkbox */
input.sheet-magie-portee-distance,
input.sheet-magie-portee-horizon,
input.sheet-magie-1cible
{
    opacity: 0;
	margin: -200px 0 0 0;		
    width: 193px;
    height: 201px;
    cursor: pointer;
    z-index: 1;
}
```

Then if you want to change the way these elements all look at the same time you just need to update one block of code. :)

clevett:
Also on a future update please move images to Github. We've had issues in the past with images hosted outside getting removed then sheet look broken so we're requesting all images be hosted in the repository with the sheet.

Thanks,
Cassie

PR #4774, opened 2019-02-28
Earthdawn (FASA Official) character sheet v1.000 (new sheet)

ChrisDDickey:
## Roll20 Requests


ChrisDDickey:
P.S. This sheet was requested and the name Earthdawn (FASA Official) was approved by andrew@fasagames.com and several other people with @fasagames.com email addresses.  Andrew is head of tech at FASA and can be contacted if you have any questions about whether the game is FASA Official.

clevett:
Thanks for including that Chris. :)

Earthdawn (FASA Official)/sheet.json
```diff
@@ -0,0 +1,8 @@
+{
+    "html": "Earthdawn.html",
+    "css": "Earthdawn.css",
+    "authors": "Chris Dickey",
+    "roll20userid": "633707",
+    "preview": "Earthdawn.PNG",
+    "instructions": "The Roll20 Earthdawn� (FASA Official) character sheet is optimized for 4th edition, but currently has limited support for other editions./r/r For more information please see https://wiki.roll20.net/Earthdawn_(FASA_Official)"
```
Anduh:
```suggestion
    "instructions": "The Roll20 Earthdawn® (FASA Official) character sheet is optimized for 4th edition, but currently has limited support for other editions./r/r For more information please see [Earthdawn Sheet wiki page](https://wiki.roll20.net/Earthdawn_(FASA_Official))."
```
You could make the wiki page adress a hyperlink this way, the instruction section can use markdown syntax.

ChrisDDickey: Thank you. I think I did this unless I messed it up somehow. 
(I made the change and pushed a new commit - Which as far as I can tell it is considering part of the last pull request).

Anduh: Yes, any commit to your branch that you have the PR comiing from automatically updates the PR, ask you see from the new commit under this conversation. 

It's also nice that I finally managed to use the 'code suggestion' feature correctly, it's a bit wonky but has some use.

clevett:
Greeting @ChrisDDickey 

I have merged this request but in testing I was able to consistently get the [sheet workers stuck in a loop](https://gyazo.com/f53bdd2206705e64e6f0f16e097088bb).

This happened when I loaded up the character sheet on my test table and made a new character sheet then pressed the Core tab.

You may want to find what sheet workers are triggering this and disable them for the time being. If its triggering due to a change: loop you may want to use the [silent option](https://wiki.roll20.net/Sheet_Worker_Scripts#setAttrs.28values.2Coptions.2Ccallback.29_.5BAsynchronous.5D).

Let me know if you are unable to find the issue.

Thanks,
Cassie

ChrisDDickey:
Thanks @clevett 
> I was able to consistently get the [sheet workers stuck in a loop](https://gyazo.com/f53bdd2206705e64e6f0f16e097088bb).
> Let me know if you are unable to find the issue.

Thank you very much, I have been working on this, and I pretty  much knew instantly what it was doing, but can't make it happen on my own computer. 
I do have a situation where when A (anything in the AP column) is set, it sets a value into B (the value column), which in turn causes it to set a new value into C (the Base column), When C is set, it checks to see if A is compatible with C, and if not, changes A.  
The purpose is to allow the user to edit ether column A or C, and have the other values change to match.
Your video shows the values of A, B and C flipping around madly, so that is obviously the loop it was stuck in. The problem is that I can't duplicate this on my system, because like I said, it should only update A if the value in C is incompatible.
 
May I ask what browser you are using? I am using Chrome. 
The reason I ask is that while working on this I found [this](https://app.roll20.net/forum/post/7266026/sheetworker-on-change-events-not-firing-when-changes-are-made-within-a-routine-called-by-sheet-opened-event#post-7266231)
and it might be that this issue is causing different behavior depending upon browser used. 

I am also wondering if the computer you were using is remarkably fast or slow? Or is your connection to the database server unusually fast or slow? The reason I ask is the possibility that it might be some sort of timing issue where on my computer a value gets set before being read, and thus works fine, but on your's perhaps it is being read before it is well and truly set? I mean it should not be that, but I can't think of anything else because it should not be doing that and I can't make it do that on my computer. 

Anyway, I have a version that both works around the sheet:open issue, and uses the "silent" option when C changes A (with extra code to also set B and everything that uses B silently as well). The new code works fine on my system, but the problem is that the old code also worked fine on mine! I was wondering if you would be willing to repeat the test with this code. 
https://drive.google.com/file/d/1c-0HTu5qn4NaJCxyp_o7dcUeSarCZ6Yc/view?usp=sharing
If it works I will put in a new commit/pull req. 

Thanks again very much.

clevett:
Greetings,

I am using Chrome with no APIs.

Can you go ahead and put that new code in as a pull request? I will download it and test it in my local environmen and on our staging environment to see if it gives me the same behavior. I bet the silent option will fix this issue. I've done the same thing once on my Shadowrun sheet. :)

Thanks,
Cassie

PR #4781, opened 2019-03-02
Splittermond: Fake Buttons -> Action buttons

ap3h3ad:
- I replaced the input/span fake buttons with action buttons. Therefore there is a huge loss of unnecessary CSS that I removed

- I replaced a workaround from the early days of the sheet that handles the deletion of masteries with a shorter and more precise version

- I also implemented that the overview of the most important values above the sheet is getting updated automatically

- Some coding optimization here and there


clevett:
Action buttons are great! I hope you enjoy them.

PR #4787, opened 2019-03-04
RuneQuest:Rolepaying in Glorantha  version 04.03.2019

Lockbox313:
## Changes / Comments

Tweaked CSS so skill names can be read when modifying repeating sections

Added fields for coinage on the 2nd tab.  Currently this does not effect encumbrance.

I added a bunch of new hit locations including all the location tables  at the start of the bestiary. 

Added fields for individual weapons skill modifiers or occasion such as Bladesharp cast as per request  
[https://app.roll20.net/forum/post/7170823/runequest-roleplaying-in-glorantha-feature-request-weapons-skill-mod/?pageforid=7170823#post-7170823](url)

I also added checkboxes with the heading Spells Cast for Vigor, Strength  and Glamour they handle all the fudges listed under the spells,such as skill category  increases, damage die steps and max enc.So don't add the stat bonuses in the mod fields of the characteristics.The bonuses are automatically added to stat rolls.A message appears under the stats  reminding you to add the bonus when doing resistance table rolls.  There is probably better way of doing this but it would mean changing a lot of sheetworkers and/or adding a whole new column of mod fields.

Removed a lot of tables in Hit location roll templates

## Roll20 Requests

.

### DIFF
--- a/RunequestGQS/ReadMe.md
+++ b/RunequestGQS/ReadMe.md
@@ -34,6 +34,20 @@ To remove the experience checkbox from a skill add a - at the start of the name 
 
 STR and DEX minimums are not currently supported this maybe added in a future release.
 
+## Version 04.03.2019
+
+Tweaked CSS so skill names can be read when modifying repeating sections
+
+Added fields for coinage on the 2nd tab.  Currently this does not effect encumbrance.
+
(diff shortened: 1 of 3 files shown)


clevett:
Thanks for the comment. That was helpful.

PR #4810, opened 2019-03-09
Streamline Fate_Core sheet and add worker script

player-03:
I've done most of my testing in Firefox, but I have players who use Chrome and Edge, none of whom report any problems.

New features
===========
- There's a section for boosts and situation aspects.
- Checkboxes show up next to aspects and consequences, helping to track free invocations.
- Mousing over a section header shows a brief rules reminder.
- The sheet worker automatically shows the correct number of stress boxes based on how good your Physique and Will are. It also adds an extra mild consequence slot if the skill reaches +5.
- The sheet worker automatically checks off the free invocation box for any consequence you receive.

Changes to existing features
=======================
- Instead of taking up space between columns in the table, the roll buttons now only show up when you mouse over a skill.
- Grey skill slots turn white once you enter a skill.
- Stunts and extras are now repeating. However, the original textbox is still there at the top of the list, meaning players can still see what they had. This box is smaller than it was, but it's also resizable, meaning they can just resize it if they don't want to bother with the repeating boxes.
- The top three rows of the skill pyramid are hidden by default, to save space. I updated sheet.json with instructions on how to enable them.
- The third stress track is hidden by default, because most campaigns won't use it. Again, the instructions are in sheet.json.
- Replaced the out-of-date sheet preview image.
- Hid some skill boxes in order to make the skill pyramid look like a pyramid. This is only an aesthetic change, and the boxes remain fully functional.

Potential data loss
===============
- Removed negative skills entirely, to save space. In the unlikely event that someone was using these, they'll have to dig through the "Attributes & Abilities" tab to find their data, then copy it into the notes field at the bottom.
- Replaced `attr_ID` with `attr_character_name` so that Roll20 can recognize it properly. This will cause players to lose whatever they'd put down as an ID, but it'll be replaced with whatever they put down as the sheet name. If they wrote the same thing in both places, they won't even notice the difference.


player-03:
Oh yeah, and I added some links to the Fate SRD. Of course, they don't show up because Roll20 strips `<a />` tags, but they'll be there if that ever changes.

player-03:
Oops! I don't know why I changed `attr_Notes`. It's back to normal now.

As for `attr_ID`, I went through a few iterations, but eventually I decided to move it to the description. I didn't want to overwrite `attr_character_name` because I imagine a lot of people actually use that. That's what shows up in the journal tab and chat window, so dedicated players would have already gone to the trouble of setting it.

So instead, I wrote a script to add it to the description, after making a few checks. For instance, if the character name was "Superman," the description was "Man of Steel," and the ID was "Clark Kent," the character name would remain "Superman," but the description would become "Man of Steel (aka Clark Kent)." If there was no description beforehand, the worker will skip the parentheses and just put "aka Clark Kent." And if the name and ID were both "Superman," it won't change the description at all.

player-03:
Also, I didn't realize GitHub would keep a complete record of my force pushes. I was trying to keep the log clean, and instead I made it messier. Oh well.

Anduh:
You might want to check the forums, [there is some disagreement](https://app.roll20.net/forum/post/7284174/fix-the-fate-core-sheets)(warning: some are rude) on whether recent changes to a Fate Core sheet was all positive, and it seems this is the sheet that was updated latest, so they might be talking about this. If not, disregard.

player-03:
Ok, thanks for the heads-up. I'll do my best to address it.

PR #4825, opened 2019-03-13
Update Fate_Core sheet based on user feedback.

player-03:
My previous changes (#4810) [didn't go over that well with some players](https://app.roll20.net/forum/post/7284174/fix-the-fate-core-sheets/), specifically those who used the features I chose to hide.

This revision of the sheet attempts to address the concerns raised in that thread. It includes:

- Formatting improvements/fixes and better browser compatibility.
- A "high-level skills" checkbox that shows the high-level skills, rather than having to manually edit attributes.
- Roll modifiers, with an option to turn them off.
- Better auto-calculation of stress boxes.
- Stress tracks are now repeating, and you don't have to manually edit anything because they're always visible. I don't attempt to auto-calculate the stress boxes, because I realized if someone is using a custom track, they may also be using custom rules for how many stress boxes they need.
- Found some more lost data (`OpenStress1`, `OpenStress2`, and `OpenStress3`) and wrote a script to copy it over.

Possible future changes:

- Leave the "roll modifiers" box unchecked by default.
- Add some way to show/hide boxes in the repeating stress tracks, either by picking a number from a dropdown or by typing a skill into a separate textbox.
- Add some way to control the default height of stress and extras textboxes, so you don't have to manually resize them every time.


Anduh:
If you feel some less used features are in the way of your gameplay, you could consider making some parts of the sheet optional with a checkbox or something.
I've personally strived to only *increased* functionality on the sheets I've been working on, and create options for things that are more up to personal taste than strict functionality.

player-03:
That's what I ended up doing. Roll modifiers are back, but now there's a checkbox for them.

player-03:
And now there are even more options, cleanly separated into an options menu.

- You can hide the advantages and extras sections if you aren't using them. This frees up room for aspects and stunts, respectively.
- In the old version of the sheet, the stunts and extras sections had one big textbox rather than repeating small boxes. Now there's now an option to go back to that.
- Instead of a single checkbox for the top three rows of the pyramid, there's a dropdown that lets you pick the exact number of rows to display.
- Since you almost never need to add or remove aspects, stress tracks, or consequences, I moved that functionality into the options menu, freeing up space in the main sheet.
- You can change the governing skill for the physical and mental stress tracks, because one forum user pointed out that these skills vary depending on which flavor of Fate you're playing.
- I made the options as intuitive as I could, but I also included title text for all of them, just in case.

player-03:
Barring bug fixes and rewording, I think that's everything.

- I split the options menu into two sections to (hopefully) help people find what they're looking for. I mean, eleven options shouldn't be totally overwhelming, but it isn't dead simple either. I'm not 100% happy with the section names, but I can't think of anything better.
- I removed the parts of the text that describe how Fate Core "usually" does things. It just felt too prescriptive. (This only applies to text in the options menu. The skill section still contains Fate Core's list of skills.)
- You now have full control over custom skill tracks, meaning you can set governing skills for them and the sheet will track the number of boxes.
- Instead of a governing skill, you can type in a number from 2 to 4 and it'll display that many boxes. (Applies to normal and custom skill tracks.)

PR #4836, opened 2019-03-16
Call of Cthulhu 7th Ed. - Size fix

orphs:
## Changes / Comments

A couple of users have mentioned that the sheet appears tiny to them. I have not made out yet if its from personal settings or something done wrong by me, but removing the Sheet's set-width seems to fix that. As it was added by me recently, I do not see its removal creating any problems.

### DIFF
--- a/Call_of_Cthulhu_7th_Ed/coc_7th_ed.css
+++ b/Call_of_Cthulhu_7th_Ed/coc_7th_ed.css
@@ -1,7 +1,6 @@
 .charsheet {
     background-color: #888888;
     border: 2px solid black;
-    width: 840px;
     color:black;
 }
 


clevett:
Without a **width:** set resizing the page especially to smaller can make the sheet look weird. You might give that a test before you decide to remove it. I like to always set a page width and usually 850px is what I do as its generally the side of the window on open.

orphs:
I know, and it does happen after resizing to 75% or lower (buttons get minutely rearranged etc). I had recently set width to 840p (there was no width value before), and after the fact, a couple of people mentioned that the sheet appears tiny to them. 
Though this is easily fixable to each user personaly by in-game-resizing of the sheet, I found it a bit of a nuisance to have to go through it each time. Seeing that there was no problem experienced whatsoever before me implementing width, I chose to remove it for the time, until figuring out what exactly went wrong.
I tested the changes with the same players that reported the ''malfunction'', and they said it appears as normal now.

PR #4855, opened 2019-03-19
Ars Magica 5th - Combat Formula fix

Medieve:
Correcting bad combat formulas. Putting in default values for Weapon Abilities.

Swapping out imgur links for github images.


clevett:
Thanks for updating those image links. :)

PR #4860, opened 2019-03-21
Le donjon de Naheulbeuk 2.8.1

Kaias26:
Few css, HTML and spelling Fix


Kaias26:
I dont know why the heck all my previous commits are here again, they all have been already merged in roll20/master...even tought i've merged roll/master into my fork...

Only the two last comits are new > a914640 and ac234e6 

After nearly one year of work, i didn't have this issues, did you change something in the process @clevett  ? :D

clevett:
Greetings,

I did some research. It has to do with the commits being squashed and merged into master now. Your git history is different from master and the pull request is wanting to update the Roll20 repo's history. Since this PR will be squashed and merged again it will continue to happen.

If you check the Files Changed you should only see your new changes so this process is unsightly but not an issue. Using branches in Git and pushing directly from a branch to the Roll20 master can be a solution. By pushing from a branch you're not changing your fork's master commit history. Then you can sync your fork with Roll20 master it will have the same history as the Roll20 master.

- Cassie

PR #4905, opened 2019-04-03
SavageWorlds-Tabbed - French translation and fix missing translation hooks.

gronyon:
## Changes / Comments
French translation and bug fixes to enable translation where it was missing :
- Updated French translation to 3.1 of SavageWorlds Tabbed sheet
- Changed run button into a roll template to enable translation
- Fixed a few broken i18n tags
- Added i18n tags to places it was missing, and added translation entries in all the language files
(however I'm quite unsure about what to do with zu.js file)


gronyon:
Thank you Cassie for that merge !

PR #4909, opened 2019-04-06
Ars magica

Medieve:
Changing the various dice rollers to make them resilient to giving broken rolls when a player enters "+X" instead of "X" when a modifier is prompted.


Anduh:
GitHub says there exists conflicts on the html file, you need to fix that before any changes can be approved

Medieve:
Odd, not sure how I managed to conflict with myself but I think I fixed it.

PR #4924, opened 2019-04-09
v2.0 Mythras Sheet

DigitalPagan:
### Compatibility Issues
* This version removes the NPC sheet in favor of a generic Character type which can be used for NPCs or PCs.  A new Compact and Mook sheet have been added to provide the smaller sheet layout and some other features useful to NPCs.  Data from the NPC sheet will be lost due to an inability to discern which of the npc or pc values one may wish to have in the upgraded sheet.  A warning has been displayed in the NPC sheet for a number of months to copy data over to the PC sheet in preparation for this.
* In addition a large amount of the sheet has converted over to using sheet worker for auto calculations.  As a result many variable value formats have changed.
* A number of variables have been renamed to reflect a more generic approach to setting support.

### New Features
* Added the Compact view.  This uses the same values as the primary sheet but displays them in a simplified and compact form.  This allows a user to maintain basic easy to navigate sheets for NPCs without loosing the advanced features of the PC sheet.  If you need to access modifiers or other values not displayed in this view simply switch over to the more advanced views, edit and return to the compact view.
* Added the Mook view.  This is nearly identical to the Compact view but provides multiple entries for current values.  This allows GMs to have one sheet representing multiple instances of like characters.  Useful when you need to have like NPCs but wish to track the full assortment of values which won't fit into the three token bars.
* Rank is now an optional field
* Movement values are now rounded to make them more readable
* A number of sections can now be collapsed hiding them away when unneeded.
* Fatigue now uses a number based value under the hood which allows it to be set up as a token bar more easily.
* Added a toggle which switches the character between spirit and material forms.  Useful for defining spirit type characters and for shamans who often enter the spirit realm.
* Removed the tenacity for spirits option.  Now it is simply enabled when Tenacity is.
* Added a method to define separate swim and fly movement species values.  Useful for defining non-human characters.
* Added Mythra Imperative as a setting which makes the sheet fall inline with the simplified rulesets better.
* Added a sheet option to set AP to a particular value instead of being based on Characteristics.
* Added a sheet option to enable/disable weapon reach
* The encumbered toggle for skills is automatically checked/unchecked based on the base characteristics selected
* Added buttons at the sheet footer which will show a link in the chat window which serve various purposes: Report a bug in the sheet, contact the sheet author, and visit the Mythras Roll20 Forum on The Design Mechanism's forum board where we discuss beta testing, plan feature enhancements, and other development tasks.
* Firearms and Missile Weapons were merged together into Ranged Weapons
* Shields were merged into Melee Weapons
* Weapons and Armor and now primarily listed in the Equipment tab.  They still display for utility in other locations but they may only be added/removed in Equipment.
* Added a sheet option to calculate Herculean difficulty at 1/10th or 1/5th standard value.
* Added a new sheet option for the use of Linguistics instead of Language/Literacy which slightly modifies the language section to better represent how linguistics applies to multiple languages.

### Bug Fixes
* After the Vampire Wars was set as the default campaign value, it is now Default again.  Most probably never noticed because the default was reset by campaign options.
* A few bugs in the new hit location presets were resolved
* Resolved an issue which was preventing sheet upgrade functions from running when sheets were upgraded.

### Other Changes
* Quick Stats renamed to Build (Doesn't not take you through the build process, still just provides a quick way to input the fields you need when first defining a character).
* Traits were broken out of Info and into it's own section
* The Penalty field has been renamed to 101+ which will hopefully make it's purpose clearer.
* The All tab has been removed, partially because I needed to free up the space for new tabs and also because the new Compact tab can be an effective replacement for the purpose the All tab filled.
* Combat Styles is now just Combat and holds a number of things useful for combat like a handful of standard skills.  Weapons were also merged into this section.
* Roll buttons which use d100 based rolls have a hundreds d10 icon instead of a d20.

## Roll20 Requests

*Include the name of the sheet(s) you made changes to in the title.*

- Is this a bug fix? Yes but it is primarily a feature release.
- Does this add functional enhancements (new features or extending existing features) ? Yes many
- Does this add or change functional aesthetics (such as layout or color scheme) ? Yes
- Are you intentionally changing more that one sheet? If so, which ones ? No
- If changing or removing attributes, what steps have you taken, if any, to preserve player data ?
Some values are being renamed and/or converted to sheet workers.  In general a versioning worker script should convert values as needed.  We are deprecating the NPC sheet in favor of a compact character sheet view.  In this case it is not possible to discern which values the PC or NPC we wish to merge into the final product so I posted a warning message in the last sheet version months ago.  Users have been notified that the NPC sheet is going away and that they should copy the data into the PC portion of the sheet and given time to do so.
- If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ?
Not new but for the most part yes all the minimums and most of the extras are done.


clevett:
That comment was very helpful.

PR #5046, opened 2019-05-06
Pathfinder Community Sheet - updated folder/file names

vince-roll20:
- changed repo folder name to "Pathfinder Community".  Sheet is no longer representative of the Pathfinder Neceros sheet of which it was originally designed to look like, and is most commonly known by the roll20 community as the "Pathfinder Community Sheet".
- updated html, css, and preview image names as well.
- updated sheet.json to reflect new file names.
- added "Images" folder and included copies of images used on the sheet and roll templates. (sheet url's will be updated in a future update)


vince-roll20:
Thanks Cassie

PR #5071, opened 2019-05-13
Pathfinder Community v1.735 - damage_macro fix and localization list sorting (Skills, Conditions) 

vince-roll20:
- fix for damage_macro resetting to default macro when closing and re-opening the sheet
- added localization list sorting for Skills and Conditions


vince-roll20:
oops.  My bad Cassie.  That would have made a few people upset...

PR #5079, opened 2019-05-15
Forbidden Lands v1.2 - carry calc fix, pack rat option

vince-roll20:
- fixed carry calc (should use base strength, not damaged strength)
- added pack rat talent option that adjusts carry amount


clevett:
I was fiddling with this sheet this weekend. It's very nice. The die roller and template are exceptionally helpful.

vince-roll20:
> I was fiddling with this sheet this weekend. It's very nice. The die roller and template are exceptionally helpful.
Thanks.  Wes had the heart of the roller/template done and was kind enough to donate the code.  I really just put some "lipstick on the pig"  ;-)  Of course the API roller is it's own animal, but it's handy to simply roll from the sheet.

Issue #4986, opened 2019-04-19
[Low Fantasy Gaming] Initiative: suggested changes.


Dekstier:
When rolling initiative it is displayed as followed:
![image](https://user-images.githubusercontent.com/49796099/56443054-35c21e80-62f3-11e9-9cf2-f59cd4450aef.png)

There are **two** noticeable issues right now:
**1. When rolling initiative the character is not automatically added to initiative.** I have seen this functionality on other sheets. Is it possible to add the same here? It would improve the games happy flow as the GM currently has to manually add players to initiative.

**2. To determine the initiative order this system only requires a 1d20 without modifiers.**
The rules say:  "Each player rolls 1d20 for their adventurer, and the GM rolls for the monsters. Resolve Turns from Highest to Lowest. ~ p44 of rulebook".

Tagged @dedurrett who has worked on this sheet before.

Anduh:
1. Yes, it should be possible. It seems some commented out code called "Old Initiative" did have the `&{tracker}` that would add initiative rolls to the tracker. 
2. Personally it sounds odd that there exist no bonus modifiers to initiative, but if that's the way the rules are it certainly sounds it should be changed. It's always possible to make adding modifiers an option, even if it might create extra work. A sheet should certainly function by the official rules.

dedurrett:
Gents-
Currently the sheet supports the initiative procedures laid out in the Deluxe Edition of LFG, which make it a Dexterity attribute check; resolved in groups based on pass/fail of the check without individual initiative numbers. The previous method is still listed as a variant rule, but I was not able to include that option on the sheet in a way that I liked. I have made some other tweaks to the sheet and will take another look at integrating it before I submit the changes.

Dekstier:
@Anduh, you make a very good point about modifiers. For instance when a character currently has a dagger drawn he/she gets a small bonus to initiative. 

@dedurrett, thank you for explaining i did not know that initiative was different in the Deluxe Edition (I don't own them). Thanks for your hard work!

clevett:
Closing this issue as the feature will be implemented in pull request #5083 going out tomorrow.

Thanks for working together on this. :) Its heart warming to see everyone improving community sheets.

Issue #4985, opened 2019-04-19
[Low Fantasy Gaming] Crit damage: Has to be filled manually.


Dekstier:
The way attacks look right now on the sheet:
![image](https://user-images.githubusercontent.com/49796099/56442798-4625c980-62f2-11e9-830b-33f67577768a.png)

**The issue** is with the field: 'Crit Dmg'.
The rulebook says: “Critical hits automatically cause maximum damage + half your level (round up) ~ p52 of rulebook”

Right now this has to be filled manually. Is it possible to automate this field? 

Tagged @dedurrett who has worked on this sheet before.

Anduh:
> “Critical hits automatically cause maximum damage + half your level (round up) ~ p52 of rulebook”
I assume "maximum damage" means the max roll of the base dmg, so in you above example, the character is either level 1 or 2, given that  1d4 base dmg results in 4 + 6 = 6 Crit dmg.

Sure, that sounds possible to automate.

dedurrett:
I'll take a look, but given the freedom for die expressions in the base damage field it would be dangerous to make assumptions and try to parse for calculation of the max damage. The functionality provided mimics what I saw in the 4E sheet (the only other system I've played where crits result in "max" damage instead of extra damage).

Dekstier:
@Anduh, It would actually be Str mod + max roll + half level. 

@deburrett, you do make a fair point. The system gives players a lot of freedom to incorporate their own  created character improvements. A player could change the way crit damage works for example. 

Would it be better if i close this issue or do you have an idea to make this work?

clevett:
Greetings,

Per @dedurrett  suggestion in #5083 , I am going to close this issue. This suggestion, while a good one, is viable to implement at this time. If more intricate dice parsing options come along we it could be evaluated again. :)

Thanks,
Cassie

PR #5083, opened 2019-05-16
Low Fantasy Gaming changes for luck attribute

dedurrett:
## Changes / Comments

Luck is a special attribute that functions a little differently from the rest. I've taken away a few fields that are not associated with Luck, modified the roll template to account for these changes, and added a dropdown for luck (attribute) saves. New column width style associated with luck added to css as well.


dedurrett:
Still not a github expert; not sure what is up with the first two commits. One is old, and the other has no diffs? I'm trying to commit the changes to Luck Attribute diffs.

Anduh:
Is the end result what you want? Of so, doesn't matter much. The commits looks differently bc you submit them directly from your master branch.

dedurrett:
Yes, I think the end result will be the same. Thanks for taking a look; appreciate what you do!

dedurrett:
Added solution as requested per:
[Low Fantasy Gaming] Initiative: suggested changes. #4986
Initiative now has a checkbox toggle for Classic vs Deluxe roll method. Deluxe is checked by default.

In regards to:
[Low Fantasy Gaming] Crit damage: Has to be filled manually. #4985
Without a way to evaluate the maximum result of a roll expression I don't know how to parse and evaluate it myself. Suggest closing this request as I will not be able to fulfill it.

dedurrett:
added default value for hidden init toggle attribute

clevett:
@dedurrett Thanks for the detailed comments and following up on those Issues.

PR #5110, opened 2019-05-22
Ars Magica - Combat Rollers

Medieve:
Brackets and parenthesis were out of order, messing with how the modifiers were being added (or in this, not at all)


.gitignore
```diff
@@ -235,3 +235,4 @@ Call_of_Cthulhu_Spanish/Delta Green Variant
 /base.css
 /app.css
 /.vs/config/applicationhost.config
+WFRP4CharSheet/WFRP4Charsheet_beforebutcher.html
```
Anduh: I doubt this edit was mean to be included

Medieve: Oops, no its not. Uh, not sure how to remove the file from the submission.

Anduh:
```suggestion
```
maybe this works

Anduh: If you approve the above code change the problem should be solved

PR #5161, opened 2019-06-09
[L5R 5th Edition Unofficial] - Version 0.1

TheAziraphale:
## Comments - New sheet
A new take on Legends of the Five Rings 5th Edition sheet. Goal is to have a functional sheet with rolling functionality that requires no extra setup from user, requiring no API-scripts to work. Started from scratch and I'm still a bit rusty on the roll20 limitations for sheet creations. I made this after I got a request from a DM and streamer who didn't want to use the current sheets due to user issues with them.

Currently linking images from my own host, but when (if) sheet gets accepted, in version 0.2 I will link from the repo directly. All images used are uploaded already in this release for easy testing in the transition.

// Aziraphale


Anduh:
Among the [Character Sheet guidlines](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) it's mentioned to avoid using `<table>` elements for styling and instead using CSS for that purpose. Sheet have been turned down for this reason before.

TheAziraphale:
That was very sloppy of me. I removed all uses of tables for divs

PR #5166, opened 2019-06-10
CoC7 internationalization

JaegersNet:
Modifier the CoC7 Sheet for i18n and added two more fields for Pulp Cthulhu compatibility.


JaegersNet:
By this my Call_of_Cthulhu_7th_Ed - German can be removed/replaced.

PR #5189, opened 2019-06-16
Cultos Innombrables new character sheet

Lothiem:
New sheet improved from previous one (Cultos-Innombrables-Hitos) to be more true to Cultos Innombrables roleplaying game

## Changes / Comments

Global changes proposed. Some clarifications added, codes reviewed, and some topics removed that are not in CultosInnombrables based charactersheet.
- Sheet name: Cultos Innombrables
- It includes functional enhancements: new rolls, new fields, etc.
- It adds aesthetics changes and new fields to collect any necessary information during roleplaying: journal, items, etc.
- Names (codes) for attributes and characteristics have been changed, so take care if you want to change directly from previous version to this one, because many parameters may disappear.


Lothiem:
The table tags have been replaced by div and different functionality to work.
Thank you.

PR #5212, opened 2019-06-22
[Magica Logia] New character sheet

Yupe-Chang:
New character sheet of Magica Logia.


clevett:
@Yupe-Chang 

The sheet.json is invalid and this sheet was unable to go live. Always check your sheet.json with https://jsonlint.com/

- Cassie

PR #5214, opened 2019-06-24
[GURPS 1.6.0]

MadCoder253:
## Changes / Comments

- Rolltable update: Move "Ref" to the right of roll type to save space. Suggested by Mike W.
- Grimoire tab: Added checkbox to highlight spells. Added a spell level modifier column. Suggested by Mike W.
- Falling/Collision notes: Revised the default notes when the falling damage is rolled. However, these notes are only visible on new characters sheets, since the those notes can be edited on the options page.
- Bug fix: Racial  Traits Self-control roll. Success/Fail text was missing, fixed.


KlingonMike:
Correction
For: **Rolltable** update: Move "Ref" to the right of roll type to save space. Suggested by Mike W.
This should say **Roll Template** update, not ### Rolltable.

PR #5239, opened 2019-06-27
Update 5th Edition OGL by Roll20.html

geoidesic:
## Changes / Comments

- When encumbrance changes, update speed and set disadvantage appropriately

## Roll20 Requests

- [ ] Is this a bug fix?
- [x ] Does this add functional enhancements (new features or extending existing features) ?
- [ x] If changing or removing attributes, what steps have you taken, if any, to preserve player data ?
It assumes that the existing character's speed attribute data is correct when the script is added. It then adds a "max" value for speed, set to this value. Also these changes only apply if the Variant encumbrance rules are selected.
- [ x] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ?

### DIFF
--- a/5th Edition OGL by Roll20/5th Edition OGL by Roll20.html
+++ b/5th Edition OGL by Roll20/5th Edition OGL by Roll20.html
@@ -9191,13 +9191,18 @@
     const update_weight = function() {
         let update = {};
         let wtotal = 0;
-        let weight_attrs = ["cp","sp","ep","gp","pp","encumberance_setting","strength","size","carrying_capacity_mod"];
+        let weight_attrs = ["cp","sp","ep","gp","pp","encumberance_setting","strength","size","carrying_capacity_mod","speed_max","speed","advantagetoggle"];
         getSectionIDs("repeating_inventory", (idarray) => {
             _.each(idarray, (currentID, i) => {
                 weight_attrs.push(`repeating_inventory_${currentID}_itemweight`);
                 weight_attrs.push(`repeating_inventory_${currentID}_itemcount`);
             });
             getAttrs(weight_attrs, (v) => {
+                if (v["speed_max"] === undefined) {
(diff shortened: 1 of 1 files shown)


clevett:
Greetings @geoidesic 

Thanks for fixing this. :) I will need to put it through our QA process for testing. This PR may stay open for longer than typical.

- Cassie

clevett:
Greetings @geoidesic 

**Speed**
The speed update appears to be working except for one important scenario. If you use the **variant** rules and are encumbered the speed is updated as expected. However if you then toggle off the **variant** rules it needs set speed back to its original value. 

One potential solution would be setting speed_max silently before reducing it. Then if variant rule is toggle off set speed back to the max.

**Disadvantage**
The encumbrance penalty for Disadvantage only applies to the three physical skills. Adding this will turn it on for all rolls. This will cause confusion for some users. At this time we want to avoid this level of automation.

Thanks,
Cassie

geoidesic:
@clevett I'll leave that to the Roll20 devs. I'm happy with the change for my own purposes and I offer my code here as an aid to those who may benefit from it or wish to use it as a starting point.

**Speed**
The speed_max is being set silently already, so the only change required as per your comments would be to copy speed_max to speed when Variant rules are toggled off. I don't know how to check for the toggle status.

**Disadvantage**
Fair enough. Personally I think having disadvantage set across the board is a valuable reminder and the DM can always resolve any confusion. That's how I'll be running it in my campaigns anyhow.

clevett:
@geoidesic 

Thanks for the code. :) Its good to know regarding the max speed. I can look at our schedule and see about getting that added in.

clevett:
Closing this request for now. It has been added to our backlog of feature requests.

geoidesic:
Any news on when this might be implemented?

PR #5251, opened 2019-06-30
Magica Logia character sheet Bug fix html, json

Yupe-Chang:
## Changes / Comments
bug fix json, html(missing attr translationByKey)


Yupe-Chang:
Hi, @clevett 
I resolved conflicts. Is that right to do this way?
Please check this. Thank you.

PR #5260, opened 2019-07-03
(NoReturn) some improvements

Tealk:
add meditation, edit modifier
fix konstitution macro, fix rangeweapon macro
Merge branch 'master' into master
spelling corrected
fix melee dmg macro
add characteristic points
change characteristic points to text
insert weight calculation
added different fire modes

I have rebuilt the whole fork because I could not fix the error


Tealk:
The client on my second pc always wants to destroy the brench :'(

PR #5301, opened 2019-07-12
Starfinder (simple) bug fix 071219

Acyd69:


Acyd69:
@clevett I just noticed your forum post about the code push. Any chance of pushing this through? It removes 2 characters from the code. If it is too much of a hassle, I understand.

PR #5309, opened 2019-07-16
UC-253: Starfinder character mancer for testing on development server

Kurohyou:
# Known Issues
## In Charactermancer
- Some selects can lose their selection state when the charactermancer is reopened, particularly the selection of theme.
- Archetypes are selectable, but currently do not have an effect on the character
## With Applying Charactermancer selections
- Spells are not dropping on to sheet after mancer is finished
- Saves are not updating properly after mancer is finished
- some mods are not parsing properly (typically ones that require a selection, e.g. Lashunta's student ability)


clevett:
Closing ticket as the code it now outdated.

PR #5314, opened 2019-07-18
RuneQuest:Glorantha fixes and additions

Lockbox313:
## Changes / Comments

Fixed bug (reported [Here](https://app.roll20.net/forum/post/7614971/runequest-roleplaying-in-glorantha-issue-with-magic-points-and-storage-list) ) adding more than 1 magic point storage item. Another mp storage section was added. New characters created will only show the new section.  Existing characters will have red notice for the user to copy item and direct them to a configuration section to hide the old section.
	
Fixed bug in Coordination spell
	
Added support for Mobilty spell
	
Fixed.  2nd mount/alley displaying 1st mount/ally name in roll templates


Lockbox313:
Something has gone wrong with this commit  e.g. In the html the version number on line 25 has not changed, the new new check box on line 603 and  the new repeating section on 683 is not there.

PR #5324, opened 2019-07-21
Fate Core V2 Backwards compatibility

MagooChris:
I have added to the startup script a once off upgrade from the previous Fate Core sheet to this one.

I still would like some official help regarding how the Default Sheet Settings work to have a checkbox enabled by default. The way I am doing it in this sheet is not working and I haven't found a working example.

Regards,
MagooChris


Anduh:
I think you fell for the previously ambiguous wording of the [Default Settings page](https://wiki.roll20.net/Default_Sheet_Settings) , use either`value` or `default` to signify the default value of an attribute, not both.  I did cleaned up the language a bit some time ago, but as I don't know it all too well, I haven't really been able to expand it more.

The linked sheets that contains Default Settings in the article are more reliable as reference of what works and what doesn't, compared to the fairly spares article.

MagooChris:
Thanks Anduh,

I already have this:
        {
            "attribute": "run_firstTime_setup",
            "displayname": "Run First Time Setup",
            "displaytranslationkey": "RunFirstTimeSetup-Q",
            "type": "checkbox",
			"value": "1",
			"default": "1",
            "description": "Run scripts to setup the sheet for a default character.",
            "descriptiontranslationkey": "RunFirstTimeSetup-D"
		},

The value property sets what the attribute will be set to when it is checked, but what I want to say is it should be checked by default and set to 1 when it is checked...

If I can't find an answer, I might just have to change it to a selection box of two options.

Anduh:
[Darker Dungeons](https://github.com/Roll20/roll20-character-sheets/blob/master/5e%20Darker%20Dungeons/sheet.json) seems to be an example of a sheet with some check-boxes checked by default .

```
            "type": "checkbox",
            "value": "1",
            "checked" : "checked",
```

PR #5355, opened 2019-08-01
Fix folder for the new Band of Blades sheet

joesinghaus:
@clevett, could you take a look? This entry is currently referring to the wrong underlying sheet.


clevett:
Greetings,

My apologies. I did add the entry with the wrong folder name. I will ensure this change is correct after the code is live tomorrow.

joesinghaus:
Thanks Cassie! It seems to be working now.

PR #5418, opened 2019-08-27
Savage Worlds Tabbed v3.8

gronyon:
## Changes / Comments

- Ability to change linked attribute of any skill.
- Can enable/disable Soak button from the game settings.
- Made it easier to enable/disable a skill in a setting (can click the skill name, not just the checkbox, in config panel)
- Fix translations for Renaming Piloting and Climbing skills.
- Attempt to fix skills in Game Settings being all disabled by default (instead of enabled).
- Fix Show Modifiers option not working (until you unset and set it again) despite being on by default.

## Roll20 Requests

Comments are very helpful for reviewing the code changes. Please answer the relevant questions below in your comment.

- [x] Does the pull request title have the sheet name(s)? Include each sheet name.
- [x] Is this a bug fix?
- [x] Does this add functional enhancements (new features or extending existing features) ?
- [ ] Does this add or change functional aesthetics (such as layout or color scheme) ? 
- [ ] If changing or removing attributes, what steps have you taken, if any, to preserve player data ?
- [ ] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ?

If you do not know English. Please leave a comment in your native language.

PR #5439, opened 2019-09-03
Revert "Savage Worlds Tabbed v3.8"

clevett:
Reverts Roll20/roll20-character-sheets#5418

sheet.json is throwing an error when restarting the server


clevett:
@gronyon 

I had to revert this update as I cannot find where the error in your sheet.json is coming from. Can you submit the changes without the updates to the sheet.json? Are the changes in the sheet.json critical to your update?

If they can be omitted then we could break down your updates and submit them in smaller chunks to allow for easier testing.

gronyon:
You can commit without the sheet file.

Le mar. 3 sept. 2019 à 22:17, Cassie Levett <notifications@github.com> a
écrit :

PR #5495, opened 2019-09-27
Wild Talents 2nd Edition changes

Kumatei:
Added power slots and changed to dice matching. 

I don't understand why this is erroring.


Kumatei:
Why's it trying to add those commits over again, if they were already accepted? What are these conflicts, the changes I committed and am pulling are what seems to be in conflict..

Anduh:
It likely comes from the fact that you work in your master branch. It means you merged your previous changes to the sheet tries to be applied from both your master branch, as well as Roll20's master branch.

The wiki have an article on the suggested git workflow and guide to central aspects of github in the context of contributing to the Rol20 repository.
https://wiki.roll20.net/Beginner%27s_Guide_to_GitHub

PR #5584, opened 2019-10-21
Ars magica 5e

Medieve:
Some usability changes for players
-- Initiatives now output to Roll20's Tracker while selecting a token. 
-- Character sheet tabs no longer attached to an attribute, so multiple players can view the same character sheet without wrestling over tab controls.


Medieve:
Good call, edited the commit.

Anduh:
> Good call, edited the commit.

Great, that should be better. But I'm not entirely sure if it's okay to have actual line-breaks in the "instructions" section, every sheet.json I've ever seen have the things on a single line. For example, [the Official 5E sheet's instruction section](https://github.com/Roll20/roll20-character-sheets/blob/master/5th%20Edition%20OGL%20by%20Roll20/sheet.json)

It could be that it's just a habit and not an actual requirement to do so, but if you want to be on the safer side to avoid needing to fix that after getting the review later today(Monday), you could proactively change it just to be on the safer side.

PR #5589, opened 2019-10-21
Revert "[Ambition & Avarice 1E] new character sheet"

clevett:
Reverts Roll20/roll20-character-sheets#5585

Something in this PR completely breaks the VTT. I'm pulling it out for now.


clevett:
@Anduh 

Fyi, I rolled this back. Something in the CSS was majority breaking inside the VTT when the campaign loaded.

Since its not a big sheet, I'll do a little troubleshooting to try to figure out what it might be.

Anduh:
That is interesting. The sheet works when used in a custom game, and it didn't show any major things when  I now again ran the three files through validators. @clevett 

However, if I have to come up with far-fetched ideas why Roll20 doesn't like this PR, here are the clsoest I can get with any guesses:

1. A thing that I hadn't noticed before, but when a roll button have it's default dice removed, and a background-image is then applied, the button's background-image suddenly start jumping when hovering the mouse over them, a behaviour that didn't exist with default roll dice. To remove this behaviour, i set the [background position](https://github.com/Anduh/roll20-character-sheets/blob/8650ff370dc055056a2e89638f02b2ccad716293/Ambition_Avarice/ambition_avarice.css#L134) for this situation to `background-position: 0;` witch fixed it, but the roll20 css class `btn` still had some transition property applied to it, but was rendered inert from my `background-position: 0;`. It could be that there is some odd behaviour that result in conflicts, and either not using `button[type="roll"].sheet-d20` or `button[type="roll"].sheet-d6` or altering them might change this issue.
2. The commented out sheetworker. Maybe it's still partially active even if commented out, breaking things.
3. [Row 73 CSS](https://github.com/Anduh/roll20-character-sheets/blob/8650ff370dc055056a2e89638f02b2ccad716293/Ambition_Avarice/ambition_avarice.css#L73): The validator didn't consider this to be a valid style. could be removed as it's not used.
4. CSS file doesn't have an empty line at the end. Dunno but somehow I like to imagine text editors and roll20 sheets like to have the last line empty.

PR #5613, opened 2019-10-28
changed the number in input.sheet-attr width

raymondh3:
made the number 80 to fix another bug  hopefully this will be the last one needed 
some reason the skill dots were not in the correct location for some users

### DIFF
--- a/CWorldofDarkness/CWoD.css
+++ b/CWorldofDarkness/CWoD.css
@@ -511,7 +511,7 @@ table.sheet-attributes td:last-child {
 }
 
 input.sheet-attr {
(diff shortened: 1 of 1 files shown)


Anduh:
> made the number 80 to fix another bug hopefully this will be the last one needed
> some reason the skill dots were not in the correct location for some users
Chrome and Firefox doesn't always show character sheets in exactly the same way, and is the likely cause of the different experience. You could compare which browser you and those that report a problem have to see of there is a pattern.

Then again styling html/css when there are lots of `<table>` elements can get a bit hard.

PR #5638, opened 2019-10-31
Add translation tags for Lotfp

### LINKED ISSUE #5635: [Lotfp] add data-i18n tags to the sheet
Hello,
Lamentation of the Flame Princess has been translated and published in french several months ago. In my virtual table community i've had a demand to translate the sheet so I've checked if I'm able to do it but I'm not because there's no data-i18n tags in the code of this sheet.
@joesinghaus or @sirkerry or @Favashi or @jangler can you add them?
Or @Anduh can you add them via your ACSI script ?
Or someone else?
BRgds

joesinghaus:
## Changes / Comments

This should fix #5635.


Anduh:
Yep, this looks to cover everything, except the word "Bonus" used in roll buttons. That can be done with [this  method](https://wiki.roll20.net/Character_Sheet_i18n#Roll_Queries). Just note for later if this doesn't get added in this PR.

joesinghaus:
I was to lazy for that, but I should probably do it as well...

PR #5643, opened 2019-11-03
[Stars Without Number Revised] v2.5.5

joesinghaus:
* Fixed an issue with the travel cost formula when the crew field was empty
* Fixed an issue with roll buttons referencing the old "name" attribute instead
  of character_name
* Fixed the translation for Initiative


clevett:
Thank for putting this in. I was going to test that Initiative issue today when I run translations. If I find anything interesting I'll let you know. I was under the impression the way it was being 'fixed' was the correct way to do it but now I'm not so sure.

- Cassie

PR #5830, opened 2020-02-02
[Warhammer 4e Character Sheet] Banner rework & fixes

Djjus:
## Changes / Comments

- Feb 2 2020

- Banner rework, added light/dark mode to settings
- Roll public/whisper chooser moved to settings
- Fields for Career level and Career Path added.
- Perception/Stealth rolls influence from armour fixed
- Critical Hit roll fixed
- Bug fixes

## Roll20 Requests

Comments are very helpful for reviewing the code changes. Please answer the relevant questions below in your comment.

- [x] Does the pull request title have the sheet name(s)? Include each sheet name.
- [x] Is this a bug fix?
- [x] Does this add functional enhancements (new features or extending existing features) ?
- [x] Does this add or change functional aesthetics (such as layout or color scheme) ? 

If you do not know English. Please leave a comment in your native language.

PR #5838, opened 2020-02-03
Revert "[Warhammer 4e Character Sheet] Banner rework & fixes"

clevett:
Reverts Roll20/roll20-character-sheets#5830


clevett:
Reverting changes as they impacted a different character sheet.

PR #5946, opened 2020-03-15
Torg Eternity Update Notification

G-G-G:
I'm plannign a major update to this sheet, and this commit adds an notification for users about this change, and gives links to a wiki page, and forum thread where they can give feedback. No changes are made to the sheets function.


G-G-G:
I have no idea what those merge things from 2018 are, do they change anything outside of the Torg Eternity files?

PR #5988, opened 2020-03-23
Original d&d character sheet

MichaelCalleia:
## Changes / Comments

This is a character sheet designed for use with Original Dungeons & Dragons. The sheet is lacking a lot of automation in rolls and calculations to give maximum flexibility to the Dungeon Master in which Original Dungeons & Dragons rules to use. This sheet will also work with Holmes Basic D&D.

This sheet is lacking automatic roll buttons as a feature, to allow for maximum flexibility for the DM. While it is tempting to add attack rolls, this assumes the use of the Alternate Combat System rather than Chainmail. Similarly, ability checks are not added as there are multiple ways to adjudicate ability checks (3d6 roll under, a variable number of d6, roll under a d20, and more).

## Roll20 Requests

Comments are very helpful for reviewing the code changes. Please answer the relevant questions below in your comment.

- [ ] Does the pull request title have the sheet name(s)? Include each sheet name.
- [ ] Is this a bug fix?
- [ ] Does this add functional enhancements (new features or extending existing features) ?
- [ ] Does this add or change functional aesthetics (such as layout or color scheme) ? 
- [ ] If changing or removing attributes, what steps have you taken, if any, to preserve player data ?
- [X ] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ? 
Yes, see above for an explanation about why rolls were purposely left out (OD&D is just too flexible, with too many ways to adjudicate).

If you do not know English. Please leave a comment in your native language.

PR #6001, opened 2020-03-23
Revert "Original d&d character sheet"

clevett:
Reverts Roll20/roll20-character-sheets#5988

@MichaelCalleia Fyi this character sheet is being remove due to an issue in the sheet.json.


MichaelCalleia:
There are some other fixes that have been noted, will get those taken care of as soon as possible.

Will have a look at the json file as well.

PR #6056, opened 2020-03-30
[Lancer RPG] Updated LANCER Character sheet to final print-release version!

spingle:
## Changes / Comments
This sheet update includes a massive UI overhaul, scores of bugfixes, added dropdown functionality and a host of other changes designed to make playing Lancer as seamless and easy as possible.

## Roll20 Requests
Would it be possible for you to delete the legacy version of the Lancer sheet that's still in the repo? The modern, updated one is under LANCER RPG(1.8), while the old version is just called (LANCER RPG). As having both of them in the approved.yml last time caused issues, I would rather be safe than sorry. Thank you very much!


oliviakristiansen:
We will review and consider deprecating the previous version. Thanks!

PR #6115, opened 2020-04-06
French Translation of Night Witches

Loadriku:
## Changes / Comments
- Translation in French of each lines based on licensed work of Edge
- There is some mistakes on the original sheet. Like a double use of "regimental" id who define Regimental Honors title AND Regimental Commander title. The text below Command Staff is also a wrong copy of Role's legend, there is no text here on officials character sheets.

## Roll20 Requests

- [ ] Is this a bug fix?


Loadriku:
After reading the sheet code, I realized the core work isn't done. Night Witches is a PBTA and we can't get the choice on the archetype played (like Monsterhearts if you're looking to compare). So the changes between archetypes can't be made (like Hawk who get only 4 slots for Deepening Ties when Pigeon got 5).

In actual state, the sheet is unusable. It need at least some work to make it usable (adding slots and every marks). In this way, the sheet would work with a few of work by the GM. The perfect solution would be to do something like https://github.com/Roll20/roll20-character-sheets/tree/master/Monsterhearts-2.

PR #6132, opened 2020-04-08
[Black Crusade Advanced API] complete code overhaul after bug report

Pixizz:
## Changes / Comments

This is a complete overhaul of the character sheet that will break one section of the sheet (talents) for previous users - it should be easy for players to find their lost data in the character advancement of the sheet if they kept track of it - which is likely in that ruleset ; i did not take steps to saveguard the broken section in any ways.
It adds a new feature in the form of two automated table roll script. It extends scripts support to almost every possible rolls on the sheet.
It should fix all of the sheet breaking bugs peoples reported to me over the years.
It does not change the layout of the preview, and only changes two minor things in the other categories of the sheet.


Pixizz:
Hey, thanks for the heads up on the conflict it should be resolved now.
For the number of attributes brokens, it will vary wildly, it can be anywhere between 5 and 50 attributes lost. It will impact GMs more than players indeed now you say it, as it will break most of their NPCs. I sadly really don't know how to safeguard that, as it's a list change and i cannot keep the auto-generated attribute names.

In this particuliar ruleset, it is absolutely required for the players to track their "character advancement" as it will impact their characters greatly, so they should really be fine.

For the preview, it was just to precise to the R20 peoples that it's normal i don't update the old preview. I do actually play with the sheet, so i know how it looks and works (and i already have a few ameliorations/bugs to iron out).

Finally, i'm not removing the translation.json ; i added a translation.txt by mistake while trying to create a folder and that's what i removed, the translation.json was only updated.

Thanks for checking all that out :)

PR #6167, opened 2020-04-12
[Agon Official] Add official Agon sheet

joesinghaus:
## Changes / Comments

This is the first version of the official sheet for Agon by John Harper and Sean Nittner. Includes a sheet for heroes, a sheet for the Strife player will follow in a future version.

There is an older sheet for Agon in the repository, but it's for a very dated development version of the game and is not very useful for the released game. ~It would probably make sense to deprecate it in favour of this sheet~.

## Roll20 Requests

- [ ] Is this a bug fix?
- [x] Does this add or change functional aesthetics (such as layout or color scheme) ? 


Anduh:
When creating HTML&CSS with PUG/SCSS, isn't it possible to output the HTML&CSS with some form of formatting?

Considering you're doing all the development & changes in your separate repo with your own set of tools, it becomes pretty obfuscated on what changes with the sheet here. Even though you are a top sheet author and might well support the sheet for all eternity; I'd think having the HTML&CSS displayed in some form of standard formatting would be a great plus in general, without affecting your development negatively.

joesinghaus:
Well, in theory yes. In practice, due to the idiotic fact that whitespace affects layout in HTML, this is not so easy to do - pretty-printing the HTML actually changes the way the sheet looks, and leads to subtle layout bugs. For example, the neat auto-expanding textareas tend to break in some cases. There are no such problems with the Javascript or CSS code, so I could make those human-readable. The danger when doing this is someone trying to read and change this code, which for all intents and purposes is not meant to be read by humans.

The maintainability fact here is a good point, but then again many sheets are not maintained at all? If I drop off the face of the earth, someone else could pick up the code in my repository, make a human-readable version, and go from there if they don't want to use my peculiar set of tools.

Anduh:
Ah, I didn't know about the whitespace-thing, then I completely understand. I've seen the sheets made bu Roll20 & Cassie having formatted html/css files so I figures there was no problems.

joesinghaus:
It doesn't always come up, but it's pretty hard to fix unless you make the font size 0 everywhere, and that comes with problems of its own.

It's a good suggestion, I can maybe think of something to make it work that's not too much effort.

G-G-G:
I dont agree with deprecating the older version, because they are two different versions of the game and the old one is still played. I know I play it.

joesinghaus:
Okay, rescind the deprecation recommendation then!

G-G-G:
Great :)
The old agon spawned quite a few variants, most of which can use the original sheet pretty easily.

For the record, I agree with Anduh's comments above about Pug's obfuscation above. I've posted recently on the forums about this, so I wont rehash that here - just wanted to add agreement.

joesinghaus:
Yup, I managed to make this work, the sheet code is now plain human-readable, also did the same for Band of Blades and Blades in the Dark as a kind of low-hanging fruit

G-G-G:
Thats great :)

PR #6318, opened 2020-04-30
Chroniques Oubliees Contemporain: fix to avoid recent Roll20 bug

Ulty:
- Added again the change on revamped skill system that was refused last week
- Tried to program around Roll20 recent [bug on linked token attributes](https://app.roll20.net/forum/post/8427520/token-bars-not-syncing-up-with-values), since we've had no news of a fix.


Reonin:
Hi @Ulty thanks for resubmitting, sorry about the headache it may have caused. I will merge this in. It looks like someone from our team has reached out as we actively investigating this issue. Thanks.

Ulty:
That's good news. Thank you @Reonin .

PR #6321, opened 2020-04-30
[Tenra Bansho Zero - Spanish] Added character sheet

Ravencrow:
This Pull Request adds the character sheet and scenes sheet (both in different tabs) for the Tenra Bansho Zero, in spanish version.

## Roll20 Requests

About the url of the sheets, actually they are in a free hosting service, but they are included in the imgs folder. Once this sheet is added to master, I'll create another PR to change the url of the images for the ones in the Github repository.

- [x] Does the pull request title have the sheet name(s)? Include each sheet name. **Yes**
- [ ] Is this a bug fix?
- [x] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ? **Yes**


Reonin:
Hi @Ravencrow because this is a translation of an already existing sheet it would be much more efficient to handle the translation through [Crowdin](https://wiki.roll20.net/Building_Character_Sheets#Translating_Character_Sheets).

Ravencrow:
Hi @Reonin ! Thanks for the feedback.

I'm afraid that is not exactly the case. There a 2 reasons why we (I talk in the name of the company publishing the game in Spain) cannot go with the actual sheet:

a) the actual sheet is a "excel-like" sheet. It's perfectly functional but, to be honest, it a totally different experience as using the super cool character sheet the game creators designed for it. We have implemented the exact same character sheet the players can find in the rule book, so there is no impact when using the character online instead of the physical table.

b) We have taken the opportunity to include additional sheets not included in the english versions, like the scenes registry sheet. We implemented it as tabs, so if they player don't want to use it, don't have to, but the player / GMs that are interested to register the game scenes can do so right there in the second tab of the character sheet.

I totally understand you point but I hope those two reasons can make you reconsider your position about this implementation of the sheets.

Reonin:
Hi @Ravencrow that's a very good point. I looked over your changes and the ones on the previous sheet that hasn't been updated in 4 years. I will move to have this replace the legacy sheet. Thanks for the response

Ravencrow:
Thanks for the review, @Reonin !

PR #6370, opened 2020-05-04
"Sentinel Comics RPG"

the-Dissilent:
## Changes / Comments
I included 2 copies of the sheet in my repository, because I do not know whether you'd like to replace the old sheet, or consider this to be a separate one. The old sheet is years out of date, and violates your standards for character sheets in multiple ways. On the other hand, I was not the original contributor, so I don't know what you might decide.
    Edit: Having 2 created an error on the pull request, so I deleted the duplicate.

## Updated
This version of the sheet updates it to match the current version of the physical sheet. It even includes auxiliary pages for the more complicated characters, as well as a villain sheet.

## Bug Fix
The macro on the old sheet, did not respond to environment status, so I removed the automatic detection, and made status user selectable. I also put in a Macro template to improve it's appearance.

## Layout & Color Scheme
Added a CSS file to improve layout, including a background color, a few other colors throughout the sheet, and some CSS grids.

## Changing Attributes
* Replaced hero name with "attr_character_name" so that it matches the character name.
* Replaced repeating "Principle of" list with only 2 instances, as that is all that will be needed.
 ** Version Update preserves only the first 2.
* Shortened Die sizes to save space.
 ** Version Update removes "1" from beginning of each die.
* I shortened Ability Types to just an initial
 ** I put `<option value="Action">A</option>` to maintain the value (and make it clear for future translations).
* Changed Ability Icons from checkboxes to cycling images.
 ** Old data had a hyphen, and could not be preserved.
* Removed input for 5 of the 6 health range fields, and had them auto calculate off the max.
* Removed repeating_bonuses, repeating_penalties, and defense, as they are not tracked on player sheets.
* The rest of the data is transferable from old sheet.


Reonin:
Hi @the-Dissilent thanks for the explanation in your submission. I believe if you have taken steps to preserve as much player information as possible and that the sheet is up to modern standards I'm going to go ahead and merge this in as the newer sheet. Thanks.

PR #6417, opened 2020-05-09
[10 Different Sheets] change image link to point at Roll20 repo

Anduh:
## Changes / Comments
* change several sheets that had image link pointing to the creator's repo so they now point at the images' location in the Roll20 repo. Tested the unsure cases

## Roll20 Requests

- [x] Is this a bug fix?

### DIFF
--- a/AllforOne/AllforOne.html
+++ b/AllforOne/AllforOne.html
@@ -222,7 +222,7 @@
                     <span data-i18n="Motivation">Motivation</span>
                 </div>
                 <div class="sheet-img-logo">
-                    <img class="sheet-hexlogo" src="https://github.com/Lockbox313/Roll20-Testing/blob/master/AllforOne/Graphics/allforone-logo
+                    <img class="sheet-hexlogo" src="https://github.com/Roll20/Roll20-Testing/blob/master/AllforOne/Graphics/allforone-logo
 					.png?raw=true" />
                 </div>
             </div>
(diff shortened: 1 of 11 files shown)


Reonin:
Hey @Anduh I know it's a pain to do but we need each sheet to have it's own pull request. We started enforcing it in the past month to make it easier to revert any single sheet commits in case of issues. Sorry about the headache.

Anduh:
> We started enforcing it in the past month to make it easier to revert any single sheet commits in case of issues. 

As I said to Olivia when she did the first two weeks of the pulls, it would be great if you set out a statement if you make a change in policy or other similar things. But yeah this is fair, I was assuming having separate PRs for this trivial stuff needed to be split, but at least I had separate commits for each.

I'll re-submit them at some point, and the sheet author community about the stricter enforcement of "one sheet per request"-policy.

clevett:
@Anduh I went ahead and merged this. I generally recommend breaking different sheet updates into different branches even for scenarios like this. Its a best practice because if for some reason one of these fails and we need to revert the change its going to revert all of them instead of just one update.

Anduh:
> There's something wrong with the image files as Roll20 shows A LOT of broken images for the All for One sheet.

Yep, I see the problem now, I'll submit a fix right away. Roll20 does the updates weekly on Tuesdays, so that's the when the fix gets applied.

PR #5451, opened 2019-09-07
Update en translation for Shinobigami sheet

ZhugeSong:
## Changes / Comments

Update English translation for the Shinobigami sheet. A lot of it seems literally translated, presumably from the original Chinese version.

## Roll20 Requests

Comments are very helpful for reviewing the code changes. Please answer the relevant questions below in your comment.

- [X] Does the pull request title have the sheet name(s)? Include each sheet name.
- [X] Is this a bug fix? No
- [X] Does this add functional enhancements (new features or extending existing features)? No
- [X] Does this add or change functional aesthetics (such as layout or color scheme)? No
- [X] If changing or removing attributes, what steps have you taken, if any, to preserve player data? Shouldn't be applicable
- [X] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository)? Not a new sheet

If you do not know English. Please leave a comment in your native language.

PR #6455, opened 2020-05-12
Shinobigami sheet translation updates

ZhugeSong:
## Changes / Comments

It's been a few months since #5451 and I had need to use this sheet again, so I have some more changes here.

Most of the translations were pretty straightforward, but I have one question: is omitting properties on the other language translations is acceptable or if I should just fill them in with the "default" translation values?

Based on that answer, I can update the PR as needed. Even if the answer is "it's fine, so this PR is good", hold off on merging for a bit so I can see about making a few more of the strings translatable, as there are still a lot that are not.

I also rewrote the skill matrix HTML to make it display nicer, using fixed size cells instead of misaligning when you have long strings (e.g., in the English compared the original Chinese). This cuts off long skill names, but they are identifiable enough I think it is an acceptable trade-off.

## Roll20 Requests

- [x] Is this a bug fix? Kinda, if you consider some strings not being translated properly "bugs"
- [x] Does this add functional enhancements (new features or extending existing features)? Not really. Besides the translations, I refactored one of the IMO ugly table implementations to make it a little cleaner and display nicer in non-Chinese languages with longer skill names.
- [x] Does this add or change functional aesthetics (such as layout or color scheme)? Yes, the skill matrix has a fixed size for the skill names to make it look cleaner in non-Chinese languages.
- [x] If changing or removing attributes, what steps have you taken, if any, to preserve player data? N/A, no attributes changed. I took care to keep all the values and names identical when moving things around, so the worst that could happen is the wrong attribute is displayed, but no data should be able to be lost permanently.
- [x] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository)? N/A, existing sheet


ZhugeSong:
Also, I am curious if there is any way to test this without getting it merged so I can confirm everything looks good. If not, no big deal.

Anduh:
You shouldn't need to add the new translations keys to the other languages, but I think it have failed to updates sometimes.

ZhugeSong:
All right, this can be reviewed now! Going through it commit by commit might help a bit, since the only big one is the last one that rewrites the skill matrix table.

PR #6473, opened 2020-05-14
[Aborea] new creation of the character sheet

Tealk:
## Changes / Comments
This character sheet has been created from scratch, so it is not compatible with the previous character sheet. But now it follows the [Building Character Sheets standards] of Roll20.


Tealk:
@Anduh 
> you only submitted some .less code, not you actual sheet html+css code

I'm sorry, but you are wrong. the diff clearly shows my changes

> If your sheet isn't compatible with the old one, just submit it as a new,

why did you write in the issue then I should only point out that the old one will be replaced?

Anduh:
Sorry, I guess I wasn't clear on the communication there. Your sheet would **essentially** replace the older one in most sense of the word, but due to how sheet are handled, what I just describes is how things are handled, you submit a new sheet/folder, and leave the old code alone so it can exist for some time for those games that still use it, while your sheet takes it's place in the character sheet dropdown menu.

And to my eyes, the github diff only shows that you remove all the rows from the existing html+css files, without adding anything. If that isn't the case, there is something wrong with how github displays the diffs, which I haven't encountered before.

Tealk:
minified code loads faster and to keep it clear I included the .less, this one is for editing and the minified .css only for roll20

Tealk:
i close this request to create a new one for this set of rules which does not replace it.

PR #6622, opened 2020-05-26
Added new sheet called Paranoia_mongoose_oneclick - allows rolling the dice pool with a single click

DarrenHorton:
The current version of this sheet has a menu system for rolling dice that takes 3 clicks to roll each time. This new sheet has the roll button on the sheet.
Added fill from the left on all checkboxes.
Added space for new skills to be added by the player.
+ a variety of other fixes.

## Roll20 Requests

Yes
- [ ] Is this a bug fix?
No it is a new version of an existing sheet, with a new name
It has new features and increased usability of existing features
It has a brand new layout and color scheme built from scratch
I have preserved all data possible, new fill from left checkboxes cannot retain original data.
Yes


Anduh:
Any reason this wasn't just added as an option for the existing sheet?
Roll20 tries to reduce the number of duplicate sheets, so there is a chance they'll decline this PR if you don't have a goo enough reason for creating a new sheet.

DarrenHorton:
By changing the checkboxes to fill from the left, the value wouldnt carry
over to a new sheet. I didnt want to break any existing players sheet.

Also, the old sheet uses <table> so i wrote this one from scratch to
conform to the newest guidelines rather than add new features to the old
one.

Maybe i could change the old sheet with some kind of option to choose
between the two formats?

Anduh:
> By changing the checkboxes to fill from the left, the value wouldnt carry over to a new sheet. I didnt want to break any existing players sheet. 

Was this "fill from left" change purely aesthetics, or was it something you had to do when you rewrote the sheet keeping to the new sheet standards?

> Also, the old sheet uses <table> so i wrote this one from scratch to conform to the newest guidelines rather than add new features to the old one.

Oooh this is very nice.

>  Maybe i could change the old sheet with some kind of option to choose between the two formats?

Yes, it should be possible, but that would probably need some work. Did you use same attribute names for all stats, compared to the old one? If you open a character  with the new sheet, that was made with the old sheet, would you "lose" some of the stats, or would everything looks as previously?

If the "fill from left" change you did wasn't necessary and could use how that was in the old sheet, it sounds like your changes could be incorporated to the old sheet, or you could incorporate the old option to your new sheet, depending on which is more easy. Having a checkbox that toggles which way things are show should be fairly doable, and the wiki example on the [CSS Wizardry page](https://wiki.roll20.net/CSS#Show.2FHide_Areas) might help in creating this toggle.

If you want to look at a complete sheet who uses more than one toggle, my *fairly* simple [Star Wars D6 sheet](https://github.com/Roll20/roll20-character-sheets/tree/master/D6StarWars) could be something to look at. It a couple of toggles to show/hide some sheet features, as well as a switch that toggles between PC/NPC view of a sheet, which I think were based on the "Show/Hide Areas" and "Swap Visible Areas" examples from the [CSS Wizardry page](https://wiki.roll20.net/CSS#Show.2FHide_Areas)-page.

You could post about this on the Roll20 forums and ask advice, opinions or testers, especially if the "fill from left" isn't something easy to go back to to save the stats.

Looking at the old Paranoia Moongose sheet, it haven't been touched in 2 years, so it would be great if your updated mechanics, looks and code update could replace the old sheet in one form or another. It's possible Roll20 accepts this sheet, but it's good to consider the option to merge these two in some form as it seems fairly doable. I don't know the sheet nor the system, but the "3-click roll" does sound cumbersome, and possibly redundant if you added new buttons, so maybe only figuring out how to keep the "fill from left" thing working is all that's needed for overwriting the old sheet.

oliviakristiansen:
We will approve this sheet since the old seems .. old. Additionally, your Pull Request includes changes to translation files. We will not deny your Pull Request, but we would prefer sheet contributors use Crowdin for translations. Please email Cassie at cassie@roll20.net and she will get you an invite to our Crowdin within 48 hours. Please use that for future translation updates! Thanks for all you do!

oliviakristiansen:
Ah, however, I cannot approve this PR because your JSON is invalid. It is because you have an extra, trailing comma in your sheet.json at the end of the last line. Please fix this and it will be merged next week.

PR #6627, opened 2020-05-27
[Heroclix] Updating documentation to match release state

Coogrr:
## Changes / Comments

Updated the documentation to match that the sheet is released and you no longer need to use custom sheet css and html to use it in a game

### DIFF
--- a/Heroclix/sheet.json
+++ b/Heroclix/sheet.json
@@ -4,5 +4,5 @@
 	"authors": "Devin Owens",
 	"roll20userid": "1113199",
 	"preview": "clixSheetPreview.PNG",
-	"instructions": "This is a character sheet to facilitate online Heroclix play, for more complete usage instructions see the readme at: github.com/Roll20/roll20-character-sheets/tree/master/Heroclix"
+	"instructions": "This is a character sheet to facilitate playing HeroClix online. See the [Readme](https://github.com/Roll20/roll20-character-sheets/tree/master/Heroclix) for further gameplay instructions."
 }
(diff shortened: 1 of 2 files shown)


Heroclix/sheet.json
```diff
@@ -4,5 +4,5 @@
 	"authors": "Devin Owens",
 	"roll20userid": "1113199",
 	"preview": "clixSheetPreview.PNG",
-	"instructions": "This is a character sheet to facilitate online Heroclix play, for more complete usage instructions see [the ReadME](github.com/Roll20/roll20-character-sheets/tree/master/Heroclix)"
+	"instructions": "This is a character sheet to facilitate online Heroclix play, for more complete usage instructions see the readme at: github.com/Roll20/roll20-character-sheets/tree/master/Heroclix"
```
Anduh, line 7:
```suggestion
	"instructions": "This is a character sheet to facilitate playing HeroClix online. See the [Readme](https://github.com/Roll20/roll20-character-sheets/tree/master/Heroclix) for further gameplay instructions."
```
IMO it's more clean to have a markdown hyperlink than show the full url. If you have `https://` at the start the link should work. Maybe a link to wikipedia or some other page that explains the game could also be useful.

Anduh: I can also mention that updating and maintaining Gaming instructions for the game would be easier on the [Roll20 Wiki](https://wiki.roll20.net/), as you wouldn't be tied to the weekly updates of GitHub nor have to deal with this. It would also enable other to more easily contribute to gameplay advice and such. You'd then also have a pretty short url to give people for looking up Roll20 instructions: https://wiki.roll20.net/Heroclix

However, you'd need to message Roll20 to be able to upload images to the wiki, as it's not available by default.

Coogrr: Thanks, couldn't figure out why the markdown version wasn't resolving on roll20 despite working on the validator I was looking at. I'm definitely looking into a wiki page, good to know on needing to message to add images there.

PR #6765, opened 2020-06-15
Sentinel Comics RPG

the-Dissilent:
## Changes / Comments
Changed "currenthp" & "greenmax" to "health" & "health_max" to allow for healthbars on tokens.
A version update script is included to preserve the old values.


Reonin:
Hi @the-Dissilent there seems to be a conflict in this sheet. Please resolve it when you are able

the-Dissilent:
@Reonin I'm not sure why it was showing both the old and new versions of the sheet, but it's updated now.

PR #6914, opened 2020-07-05
Updatet the german translation

tom959:
The V5 German Rulebook is now aviable. I changed the translation to the new official one


samus:
Hello! Your Pull Request includes changes to translation files. We will not deny your Pull Request, but we would prefer sheet contributors use Crowdin for translations. Please email Cassie at cassie@roll20.net and she will get you an invite to our Crowdin within 48 hours. Be sure to include the languages you would like to translate, Please use that for future translation updates! Thanks for all you do!

PR #6969, opened 2020-07-13
PokeroleV2 CSS correction

Deightine:
There was an aesthetic bug in the CSS code that passed notice. It didn't harm any data, but it forced the strong tags in other areas to display incorrectly.

## Changes / Comments
- Bug fix.
- No enhancements.
- Aesthetics correction.
- No attributes changes.
- Not a new sheet.


Deightine:
Apologies to the maintainers for a second merge. Somehow this flaw slipped past the final QC check.

PR #7158, opened 2020-08-09
Vaesen RPG V1.0, includes PC and NPC sheets

starshipsandsteel:
## Changes / Comments
Submitting initial build of a character sheet for the Vaesen RPG by Free League.
Includes Tab switching PC/NPC sheets

## Roll20 Requests

- [Yes] Does the pull request title have the sheet name(s)? Include each sheet name.
- [No] Is this a bug fix?
- [Yes] Does this add functional enhancements (new features or extending existing features) ?
- [Yes] Does this add or change functional aesthetics (such as layout or color scheme) ? 
- [No Changes, new sheet ] If changing or removing attributes, what steps have you taken, if any, to preserve player data ?
- [Yes] If this is a new sheet, did you follow [Building Character Sheets standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ?


Anduh:
I noticed someone else also submitted a **Vaesen** character sheet this week. See #7160

dodgepong:
Hey there, looks like we both ended up working on Vaesen sheets independently. I don't mind there being 2 versions of the sheet as long as Roll20 is okay with it -- someone may find one version more useful than the other (such as our different approaches to handling bonus dice). Either that or we could work together on one version. I know Free League has stated that they plan on releasing their own version at some point -- obviously these are both just stop-gaps for players to use until the official sheets come out.

Just wanted to point out a couple issues I noticed in this sheet that might be worth looking into:

* Conditions are not a progression, i.e. you don't always start with Exhausted then go to Battered and then Wounded before Broken. You can be Wounded but not Exhausted, for example. The only real progression is that players only become broken when taking a fourth condition.
* Is the intent that players manually add negative bonus dice by hand to factor in conditions? If so, there might need to be a check to ensure players always roll at least 1 die on skill checks, since the book states that you can always roll at least 1 die no matter how many conditions you have.
* It would be nice to have a field for players to record a weapon's range for easier reference (e.g. `0-2`).

Reonin:
Thanks for starting the convo @dodgepong . I'm going to merge both sheets in for now. It would be awesome if you both could collab together and make one awesome Vaesen sheet for the community! :D

starshipsandsteel:
@dodgepong Hey there!  I haven't had a chance to check your sheet out.  I am new around these parts, just grabbing a pro-sub and wanting to try out sheet creation, and seeing no Vaesen and wanting to run it soon I got to work!  I added you on Twitter as well.

Conditions: I didn't really view this as a progression system, but after your comment and a quick re-read, I would be better off to change this.  I had misread this or misremembered this, as you only had 1 condition at a time, which is clearly not true.

Condition Dice, negative dice: I think I set the dice to not allow going to below 0 early in the build and only thought of this, this morning, but yes, at this time that was my intention.  It was a thing that allowed dice to be modified easily.  I'll probably end up renaming it as well.  It grew out of a way to apply equipment bonuses to a roll.  Good catch on the at least 1 die, something to look at.

Weapon Range: yes, just never got around to fitting it in as I was building the sheet, I modified the layout so I could fit more in and just ran out of time to get that added in.  

NPC sheet:  I intend to rename this to "Vaesen" as it is really aimed at representing a Vaesen, not any generic NPC, as they have slightly different stats and unique conditions.

HQ Sheet: Certainly something I intend to add, but it seemed like a nice to have and wasn't necessary for what I needed this for initially, so I had planned on adding it to the next version of the sheet.

dodgepong:
I spent a bit of time thinking about NPCs this morning, and I actually do think there's value in having a player-controllable NPC sheet, because players can create characters as upgrades to the castle (recruits, guards, and the coachman).

I admit I've avoided reading the chapters on Vaesen and NPCs in general since my GM forbade me from it 😅 So if that could be useful for some players/GMs, then it could certainly have value. I've clearly approached this more from the perspective of a player and thinking about what a player might find useful, and it sounds like you've been putting thought into what a GM might find useful.

I'm also new to creating character sheets too, and got a pro sub as part of Free RPG Day. I've played a lot of John Harper's games/FitD games on Roll20 (Blades in the Dark, Band of Blades, Agon) and always been impressed by those sheets, so I ended up taking a lot of inspiration from that.

starshipsandsteel:
Nice.  I also got the sub from Free RPG Day!  

Most of my time on roll20 has been setting up games from a free perspective since quarantine, the biggest was John Carter.  Other than that I've looked at a few (Conan 2d20 and Alien), but I didn't take much inspiration from them tbh, some syntax here and there.  

I have 100% approached this from the idea of what a GM might need to run this as a one-shot, and although I could have simply kept the files and run them custom, it seemed like a better idea to share them with the community, but I recognize I have missed things and have upgrades.  Just this morning I realized I didn't have an attack roll button!  

As for player-controlled NPC sheets, the question comes out to, what differences are there between a human NPC and a human PC stat-wise?  Can they just use the PC sheet, or is something more roster-like more useful.

dodgepong:
It looks like the primary difference between PCs and NPCs is that NPCs have physical/mental "toughness" in place of condition slots. NPC sheets could probably be simplified to name, occupation/NPC archetype, attributes, skills, physical/mental toughness, and then a text box for any additional notes. I suppose if NPCs are intended to follow rules for equipment (armor and weapons) it might be worthwhile to have sections for those, too.

starshipsandsteel:
Yes, just looking at the characters in the included adventure.  Should be able to create a pretty simple sheet to handle them.  In Truth, I would probably find it handy for what I want to run.

Anduh:
Having a simplified NPC sheet is likely a good idea. And when possible, having the NPC sheet use the same attribute names as the PC side would make it easier to switch between the two and avoid unnecessary duplicates of sheet attributes.

Great to see you two hit it off with ideas and to collaborate on a single sheet. I'll be unsubscribing from the discussion, but remember that you can stop by the [Character Sheet Forums](https://app.roll20.net/forum/category/277980) if you want input/ideas/help from other sheet authors or to copy stuff from other sheets.

PR #7195, opened 2020-08-15
[Dark Heresy] remove rounding

Tealk:
## Changes / Comments

remove rounding rolls

## Roll20 Requests

### DIFF
--- a/Dark_Heresy/DarkHeresy.html
+++ b/Dark_Heresy/DarkHeresy.html
@@ -102,7 +102,7 @@
                         <td><input name="attr_Awareness3" type="checkbox" value="10"></td>
                         <td style="width:60px"><input name="attr_Awareness4" type="number" value="0"></td>
                         <td><button name="roll_Awareness" type="roll"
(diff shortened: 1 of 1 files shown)


RobbaYaga:
Why would one remove rounding? There is absolutely NO reason to list a skill roll as having 1.2 levels of success. Please return this to the correct rounding.

Tealk:
Hallo @RobbaYaga 
Because I have not found a way to round correctly. According to the rules, rounding must always be done, unless it is a negative result, in which case it is rounded up. Unfortunately I can not set both so that negative results were always one less than they should be.

RobbaYaga:
Thank you for the explanation. That helps a lot to understand the reasoning behind the change.
RY
    On Thursday, September 3, 2020, 11:58:27 AM EDT, Tealk <notifications@github.com> wrote:  
 
 


Hallo @RobbaYaga
Because I have not found a way to round correctly. According to the rules, rounding must always be done, unless it is a negative result, in which case it is rounded up. Unfortunately I can not set both so that negative results were always one less than they should be.

PR #7298, opened 2020-08-29
[Pathfinder Community] [Pathfinder 2 Playtest by Roll20] Revert "Strike Rank" to "SR"

player-03:
# Changes / Comments

Several instances of "SR" were incorrectly replaced with "Strike Rank." This is especially bad in Pathfinder first edition, which doesn't even have the concept of a "strike rank," which has led to [some confusion](https://app.roll20.net/forum/post/9015166/pathfinder-1e-sheet-spell-resistance-now-reading-as-strike-rank).

Some instances of "Strike Rank" are correct, and I've left those alone.

Specifically: 
- For the Pathfinder Community sheet, I reverted to "SR" because the field is intended to be an abbreviation.
- For the Pathfinder 2 Playtest sheet, I reverted to "SR" to match the length and style of similar fields.

### DIFF
--- a/Pathfinder 2 Playtest by Roll20/translations/af.json
+++ b/Pathfinder 2 Playtest by Roll20/translations/af.json
@@ -266,7 +266,7 @@
 	"defensive-abilities": "DEFENSIVE ABILITIES",
 	"dr": "DR",
 	"immunities": "IMMUNITIES",
-	"sr": "Strike Rank",
+	"sr": "SR",
 	"resistances": "RESISTANCES",
 	"weaknesses": "WEAKNESSES",
 	"offense": "OFFENSE",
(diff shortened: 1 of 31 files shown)


player-03:
Facepalm. Forgot about that rule. Is it really important in this case?

vince-roll20:
Don't worry about the PF Community sheet. I went ahead and submitted changes on Crowdin to revert SR back to "SR".  Not sure why it was even changed to "Strike Rank" on Crowdin...  Hopefully this should be resolved on the community sheet this week.

player-03:
I assume some manner of find and replace was involved.

oliviakristiansen:
It looks like vince-roll20 handled these changes through Crowdin. I am going to close this PR since that seems like it is the case. Please be sure to use Crowdin in the future! Thanks!

player-03:
Sounds good to me. I wasn't familiar with Crowdin because I hadn't read the wiki page about translations, but I'll use it next time.

PR #7316, opened 2020-09-01
SWN Revised v2.4.12

### LINKED ISSUE #7117: [Stars Without Number Revised] Drone attack rolls add inteligence modifier twice
I just noticed a small bug in the Stars Without Number Revised sheet. When making attacks with a drone, the roll adds the character's Intelligence modifier twice. The first time is rolled up in the drone's attack modifier, the second time it is added separately for the roll.

I've looked into the code and I have found the two places it gets included:

Here, the worker script adds it to the pre-computed attack bonus: https://github.com/Roll20/roll20-character-sheets/blob/ecfef4f21d699291de47fcfee68823145af7d3b6/Stars_Without_Number_Revised/Source/workers.js#L3001

And here it is added to the roll separately (in addition to the attack bonus computed by the code above): https://github.com/Roll20/roll20-character-sheets/blob/ecfef4f21d699291de47fcfee68823145af7d3b6/Stars_Without_Number_Revised/Source/pug/drones.pug#L45

The simplest fix is probably to remove the separate addition of the stat bonus from the roll. All that takes is removing ` + @{intelligence_mod}[Intelligence]` from the linked line in the `drones.pug` file.

An alternative would be to not use the pre-computed bonus for the roll, but rather to reconstruct it from its parts (character attack bonus, weapon specific bonus, skill, attribute). Regular (non-drone) weapon attacks do something like this. I'm not sure how easy that would be to do for drones, since they can use either of two skills (whichever is better). I imagine similar logic exists for melee attacks that can use either of two attributes, but I haven't fully investigated the code for those rolls.

Karlinator:
## Changes / Comments

Fix an attribute (Int) being added twice to attack rolls made with drones.

Closes #7117 

## Roll20 Requests

- [x] Is this a bug fix?


Karlinator:
This closes #7117

Karlinator:
Closing due to v2.5.0 being ready to merge.

PR #7319, opened 2020-09-02
Add an Into the Odd character sheet

olivier-grech:
This is a new character sheet for the _Into the Odd_ system (https://www.drivethrurpg.com/product/145536/Into-the-Odd).

into-the-odd-preview

Regarding the Roll20 minimum requirements:
* I used CSS styling to try to make the sheet look good. There is no official Into the Odd character sheet, so I made something that was pleasing to me.
* I used proper HTML syntax.
* I did not use any `<table>` elements.
* Line endings are Unix line endings.
* I tested it on both Firefox and Chrome.

Regarding the extra requirements:
* The sheet is not standalone since it relies on some external fonts, but it is still usable in the case where the fonts fail to be fetched.
into-the-odd-preview-no-fonts
* Roll buttons are available for the "save" rolls, which are one of the two types of roll in _Into the Odd_. The other type of roll is weapon damage. I'll need to add roll buttons for that in a future update.
* Obviously inputs and text fields are available.
* The rules of the game are readily available on DriveThruRPG.
* A JSON translation sheet is available.
* I created a roll template. Here's what it looks like:
into-the-odd-roll-template


oliviakristiansen:
Thank you so much for all of the info you provided! That makes merging new sheets really helpful! :)

PR #7338, opened 2020-09-06
[Heart: The City Beneath] Some updates to mk2 sheet

joesinghaus:
## Changes

* Improves layout tidiness and makes more room for information with self-expanding text areas et cetera.
* Adds a roll template.
* Fixes various bugs, such as text fields without an underlying attribute. Canonical name format for attributes in snake case.
* All existing data is preserved.


brocktopus:
Awesome!

PR #7398, opened 2020-09-17
[Chronicles of Darkness] Normalization of translation keys

luizbgomide:
## Changes / Comments
Removing all colon from translation values, normalizing the names of translation keys, minor cosmetic fixes.

## Roll20 Requests

Yes.
- [ ] Is this a bug fix?
No.
No.
No.
No changes.
Not applicable.


wing-it:
Please clarify, did you change the names of any translation keys? If so, any existing translations for those keys would be lost.

luizbgomide:
I removed colons from translation values (now it is "Name" instead of "Name:", and therefore and removed duplicates that existed before (like Krewe and Krewe-c that differ only in having a colon), replacing the key with colon with the one without it, and adding hardcoded colons where needed to avoid having duplicate keys.

The only keys removed are the duplicates, I don't think that it can affect any translations because I was the one that provided a valid translation.json last week. And even if someone started to work on a translation in this short time on the worst case scenario some label will show two colons (like "Name::") until they fix it, nothing will break.

Just a couple keys needed to be removed, but about 20 values lost a final colon to keep uniformity. The rest are cosmetic changes here and there that changes nothing.

I'm the one providing the translation to Portuguese in Crowdin, as far as I can tell there is none in other languages. I began working on the translation.json (and CofD.html), because before last week the sheet wasn't translatable (the filename was wrong), now it is all right and fixed up.

luizbgomide:
Ah, the translation.json was also outdated, and had keys defined more than one, so I had to generate a new one that contains all the keys. I did it so keeping the order of appearance, this is why the order is different and file looks completely changed.

PR #7425, opened 2020-09-21
[Lex arcana] extra specialisation fields

eux:
## Changes / Comments
Just adds some extra specialisation fields


eux:
did something wrong sorry, there are unexpected cnflicts

PR #8560, opened 2021-03-22
Ironsworn: Scene Challenges

hituro:
Update Ironswon sheet progress elements to also support scene challenges (from Ironsworn: Delve).

Adds a "Challenge" button which shows a parallel countdown track under the progress track, using a fixed 4-box format. Also converts the existing progress track to use css grid for the formatting. The countdown track adds six attributes to the progress repeating section — four to track the countdown boxes, and two to control visibility — but does not remove or alter any existing attributes.

image

image


aureyia:
@hituro you need to include the updated built HTML and CSS from stylus and pug and replace the ones in the root of the Ironsworn directory. The build isn't handled on roll20 side.

hituro:
@aureyia Thanks! Done that now.

PR #8663, opened 2021-04-06
[Metal Adventure] Fix checkboxes

TheShinriel:
### Issue:
In character sheet, there are 6 checkboxes, that can be checked and unchecked.
However, once the checkbox is checked, the data is memorized and will always be checked when the character sheet is reopened. 

### Workaround:
If check by mistake, the only way to fix is to DELETE the character sheet an recreate a new one

I guess, the default value (uncheck) should be null
When we check, the value become "0" (from the code) 

And when we uncheck, it's still a falsy value like the check value... so nothing to save about that sheet.

### /!\ NOT RETROCOMPATIBLE /!\
All previous character sheets for Metal Adventure, will have these checkboxes uncheck
They are only use for character's evolution, and one by attribute (6 for this RPG)

### Changing
About the code, I put a value when check as "1"
We could simply remove the value attribute, if I understand the script engine but I was not sure about this behavior.

## Roll20 Requests

- [X] Is this a bug fix?
- [X] If changing or removing attributes, what steps have you taken, if any, to preserve player data ?

### DIFF
--- a/Metal Adventures/MetalAdventures.html
+++ b/Metal Adventures/MetalAdventures.html
@@ -251,7 +251,7 @@
 				<div class="column-header">Agilité</div>
 				<div class="row">
 					<div class="flex-grow1">
-						<span><input type="checkbox" name="attr_predilection_Technique" value="0">Techniques</span>
+						<span><input type="checkbox" name="attr_predilection_Technique" value="1">Techniques</span>
 					</div>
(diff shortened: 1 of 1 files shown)


mperes:
Hey there @TheShinriel, your PR has some conflicts within the sheet html file, probably due to incompatible encoding. Could you rebase against master, fix those and create a new PR?

PR #9092, opened 2021-06-14
Added new WH40k Dark Heresy 2e custom sheets

vajura:
## Changes / Comments

The readme in the folder of the new sheets goes indepth into the new sheet and the sheets can be just tested in a new sandbox ofcourse.
 Essentialy this sheet features everything in the DH 2e advanced sheet and much much more

## Roll20 Requests

- [ ] Is this a bug fix?


mperes:
Hey there @vajura, before I proceed with merging you PR, I need to ask you: Are there any major difference in the rules between your sheet and the otherwh40k / dark heresy sheets that already exist in our repo? I am asking you that because while there have been duplicate public sheets allowed in the past, going forward Roll20 wants to limit this practice.  Your collaboration with existing sheets should not be restricted to translations, in fact in the spirit of community engagement, you should be able to push all sorts of features that would make running the game a better experience. Thanks for all the work you do!

vajura:
> Hey there @vajura, before I proceed with merging you PR, I need to ask you: Are there any major difference in the rules between your sheet and the others wh40k / dark heresy sheets that already exist in our repo? I am asking you that because while there have been duplicate public sheets allowed in the past, going forward Roll20 wants to limit this practice. Your collaboration with existing sheets should not be restricted to translations, in fact in the spirit of community engagement, you should be able to push all sorts of features that would make running the game a better experience. Thanks for all the work you do!

Game rules you mean? No it follows the game rules completely but its (in my opinion) a huge upgrade to the existing sheets in terms of automation of calculation that is available here. The other changes are much more descriptive damage and hit logs as well a a big code cleanup to make it easier to continue development on the sheets.

That being said i was asking a bit around the forum and you guys have a rule that new sheets need to function without the API? Which my sheets currently dont since they call API functions for every roll they need to do, in this case i would need to update them to also allow them to be used for people without a PRO account

mperes:
Ok, in this case I will close this PR. Please consider collaborating with one of the existing sheets and implement the automations and other features you've came up with in there, in that way we keep the community healthy ;) Thanks @vajura !

vajura:
> Ok, in this case I will close this PR. Please consider collaborating with one of the existing sheets and implement the automations and other features you've came up with in there, in that way we keep the community healthy ;) Thanks @vajura !

I would like to submit it again after i done the updates so that the sheet also works without API usage. But we seem to have a misunderstand here, the old sheet can in no way be updated in this way unless we replace everything in them and the sanitization work done on the attribute name (so they are correct and more descriptive) would break it for all the people using the sheets. 

Maybe we misunderstood eachother about the "rules" i thought you mean the base game rules in which case i would think this is a requisite anyways for sheets to be submited for a game that they follow the rules in that game, thats what i meant with they are the "same", as in they follow the same base rule book.

PR #9118, opened 2021-06-17
[HarnMaster3] submitting revised "v2" version of the Harnmaster sheet

olafvantol:
## Changes / Comments
Completely reworked the other HarnMaster3 sheet, key points:
- reworked layout to CSS grid
- complete styling overhaul to be inline with paper version, and added light/dark theme toggle.
- all calculations moved to sheetworkers instead of auto-calc and added more calculations (see readme.md for full feature list)
- added a Horse section (in the system there is synergy between character and horse skills) and more rolls for convenience.

I've reached out to the original author and he's happy for me to submit this update, he doesn't have time to maintain the sheet. All the changes do mean some things break for people using the original sheet, so submitting this as V2 so people can choose to switch (migration instructions included in readme.md).

## Roll20 Requests

- [ ] Is this a bug fix? No
- [ ] Does this add functional enhancements (new features or extending existing features) ? Yes
- [ ] Does this add or change functional aesthetics (such as layout or color scheme) ?  Yes
- [ ] If changing or removing attributes, what steps have you taken, if any, to preserve player data ? Submitting a new version due to breaking changes. The fact that I introduced auto-calculation for armor layers meant I couldn't provide keeping existing player data on armor.
- [ ] If this is a new sheet, did you follow [Building Character Sheets standards] (https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) ? Not new but followed standards when rebasing to CSS grid

If you do not know English. Please leave a comment in your native language. N/A


mperes:
Hey there @olafvantol , before I proceed with merging you PR, I need to ask you: Instead of creating a new sheet, could you do some refactoring and push your changes on top of a existing one? I am asking you that because while there have been duplicate public sheets allowed in the past, going forward Roll20 wants to limit this practice. Thanks for all the work you do!

nmbradley:
Hi @olafvantol 

Unfortunately, we can't accept this submission at this time, for the reason provided by @mperes above. 

In order to overcome your problem, here are two suggestions I might make:

1. Rename the armor fields, so that they are not overwriting the existing data, and relocate them to a settings panel.
2. Use a sheetworker to copy the values to some other location before they're overwritten. 

As it stands, there is already a Harnmaster3 sheet, and we cannot accept another.

kreuvf:
> Armor in Harn can be layered. Each piece of armor provides protection against 4 different types of attacks, and has individual locations it covers. The total protection value is calculated by adding the armor value of each location together.
> The old sheet just had an input per location where you had to calculate everything yourself and enter the totals per location. My update has a complete auto-calculation where users input each armor items as per the rulebook, mark the correct locations and choose if a piece of armor is worn (this way you can easily take it off and the values get recalculated). The total values then get calculated. I couldn't figure out a way for users not to lose the manual inputs they had done while offering the new calculations. This means an overwrite/update to the sheet would mean users lose their armor info if they are not prerpared.

Here is how I am solving this kind of issue for the ''Das Schwarze Auge 4.1'' sheet:
- Add a section on the sheet with legacy data
- Move legacy data to the new section
- Explain to people where they can find their old data
- Whenever possible, try migrating old data (keep old data in that new section in case something goes wrong - it will)

olafvantol:
Perfect, thanks for the feedback, I'll check your sheet for inspiration too.

-------- Original Message --------

PR #9124, opened 2021-06-20
[AD&D 2E Revised] More Spells

DBFBlackbull:
## Changes / Comments

All the changes can be read on the forum here: [Forum post](https://app.roll20.net/forum/permalink/10177056/)

## Roll20 Requests

- [X] Is this a bug fix?
- [X] Does this add functional enhancements (new features or extending existing features) ?


Anduh:
I suggest you don't remove the sheet migrators, suspect it will break the game for any games that haven't opened the sheets since the migrators was added, so if someone have been on a break for like a year or not opened some sheets in a backup campaign in this time, they have nothing to go with.

Once a migrator, based on version is introduced, it should stay there indefinitely.

DBFBlackbull:
Hello Anduh.

Thank you for the notice. I have been adding / removing migrators to the sheet for many of my updates, and was never made aware that they should stay for all time. I make sure to leave a migrator for a couple of updates, say if one is introduced in 3.1.0, then I leave it there untill I get to 3.5.0 or something like that. I actual time this often tends to lean towards 3-6 months from introduction to removal (often the closer to the 6 months).

My major reason behind removing the migrators after a period of time, is to keep the code clean and not have any unneeded legacy code bulking up the sheet unnecessarily.

All my migrators op to this point has only been migratin very few, non essencial fields. To my estimation, a returning player missing the value in this field is not losing a lot of data that cannot easily be replicated.

If this blocks my pull request from being accepted and getting merged in, then I will go back and add the migrators back in. I will ensure to add all of them, all the way back to my first one.

mperes:
Hey @DBFBlackbull, @Anduh has a point. Even though there is no clear roll20 politics regarding this topic, from a development perspective is not a good idea to remove bits of migration code from the sheet. Even though the case raised by Anduh will not happen often, and like you said, the damage caused would not be that big, it just doesn't pay off. If you goal is cleaning this bit of the code, I would suggest extracting really old update scripts and adding them into a separate file / folder. Please le me know how you feel about this. Thanks for all the work you do!

DBFBlackbull:
Hello @mperes
I can certainly move The migration code to another file in order to Keep it clean. However I will have to make The infrastruktur to do so, which will take some time. 

Will you Block My pull request until I do so, or can this one go through, and then I can readd The migration code in another Pull Request?

DBFBlackbull:
@Anduh and @mperes 

I made another commit, adding in the migration code that was removed. I also went back through my commits all the way back to my first, and added any migrations that had been removed.

The version.js now have 3 seperate migrations. I hope this satifies your desires for backwards compatibility and returning players :)

PR #9160, opened 2021-06-27
[Imserso_ToTheLimit] First version of this new sheet

jorsermar:
## Changes / Comments

This is the sheet for the Spanish System Imserso: To The Limit, that recently has funded its second version in verkami.

## Roll20 Requests

- [ ] Is this a bug fix?


Anduh:
Few things:

- Don't worry about the failed check, that appears with all PRs for new character sheet.
- It looks great!
- I see you have commented out code in the HTML in a few places, is that intentional?
- You have the CSS code minified, is that because you used some pre-processors or frameworks to generate it? If so, committing the source files to the repo would be a good idea, making it possible for other to also work/see on the original source files. Saving the CSS prettified would also make it easier to deal with

jorsermar:
Hello, thanks for the clarification and the positive feedback. 
The comments were left behind by mistake, thanks for pointing to it. 
The css minification is part of the framework i was using yes, i've disabled it now to make easier the collaboration of other members of the community. The reason not to share the original source files is that some are part of a framework from another developer that hosts it in his own repository, and I'm not confortable sharing them, that's why I decided to share only the compiled files (besides, it involves the use of grunt, less, etc. and some developers may not feel confortable using those)

PR #9970, opened 2022-01-01
The One Ring 2e - initial release 01.00.00

bhuesemann:
## Changes / Comments
01.00.00
This is a new character sheet for The One Ring 2e build from the (basic) 2e and 1e version. It is a complete rewrite (based on css grid layout) and revamps the look&feel to the official 2e character sheets.
- it includes two sheets: player characters and adversaries (which was not in the (basic) 2e but in the 1e version)
- it has new extended support for all typical rolls (including attributes from selected tokens), including toggles for combat stance or parry/protection options
- translation is supported based on the 1e and (basic) version

- the sheet should be compatible to The One Ring Second Edition (basic). All attributes are represented in the revamp. One misspelled attribute ("treaure"->"treasure") was migration in the sheet workers.
- this is my first character sheet but I've read and followed the numerous wiki related articles including the standards for character sheets.

## Roll20 Requests

- [ ] Is this a bug fix?
- [x] Does this add or change functional aesthetics (such as layout or color scheme) ? 
- [x] If changing or removing attributes, what steps have you taken, if any, to preserve player data ?


bhuesemann:
Renamed the sheet preview image. What failed else in the character sheet validation?

bhuesemann:
Hi @nmbradley,

I understand that splitting the community (players and developers) is something that roll20 (and me too) wants to prevent.

- regarding compatibility: yes, it should be compatible to the existing sheet regarding the data, but it is a complete rewrite with partly different/enhanced logic for rolls, sheet-workers and completely new CSS. I'm pretty sure 90% of the users will switch to this sheet after some time (when it is stable), but we don't want to "force" migration of existing games to this sheet. I've discussed this with the maintainer of the existing sheet (@aragant)  and we have come to the conclusion it would be better to release it as a new sheet to give everybody the freedom to migrate at their own chosen point of time. In IT terms we want to reduce risk for users not doing a bing bang migration, integrating user feedback in the transition phase, and ideally communicate clearly the existing basic sheet is deprecated and will not be maintained in the future. We have also discussed to put the new sheet on a separate tab in the existing sheet, but this has side effects with the new sheet workers (as they implement additional, non-compatible  functionality). So data migration is possible but we can't run them in parallel. What is clear: we will not split the developer community as we are all on the same page from the beginning and I was in close contact with Aragent before I started working on the changes.
- Regarding the translations: I could remove all existing translations of the basic 2e version of the character sheet. But why? I guess they have been created already  by the translation service and should work right out of the box with new version of the charactersheet. I've paid close attention to reuse all existing tags in the UI of  the translation.json and added new entries to the end of the file. I could add all new entries in translation.json to all language files in the subdirectory if this is required. Deleting the files will delete all the existing work and raise the need to retranslate everything again. This would be painful for users that want to migrate to the new sheet, as all existing translations they already know from the basic version would be gone.

- Last point: the name of the existing sheet (The One Ring Second Edition (Basic)) was chosen to indicate that the sheet was meant as a fast version to provide early functionality when the game was still in its beta phase and not released. The new sheet is not "basic" anymore thus changed to a new sheet name. 
Best regards,
Bodo

bhuesemann:
I've just realized a different author has announced a new The One Ring 2e character sheet in the Roll20 forum.

https://app.roll20.net/forum/post/10558357/the-one-ring-2nd-edition-sheet-version-2-dot-00-pre-announcement

I've tried to contact the author via Roll20 direct message. I hope we can find a way to combine our efforts into one charactersheet. That would be the best for the community.

PR #10095, opened 2022-01-27
Sigil-Shadow - Fix CRLF -> LF

jaculpiii:
Fix CRLF changes back to LF


nmbradley:
@jaculpiii This branch has conflicts caused by merging the author back in. If you don't mind and could resolve this, I'd appreciate it. Thanks for fixing the line endings.

jaculpiii:
> @jaculpiii This branch has conflicts caused by merging the author back in. If you don't mind and could resolve this, I'd appreciate it. Thanks for fixing the line endings.

I think I got it, I don't think the author changed the .html file but the git comparison tool thinks the whole thing's a change due to the CRLF/LF issue.  I hope I didn't undo something they did.

PR #10473, opened 2022-04-12
[Warhammer 4e Character Sheet ] Imp. Conditions/Add Size & Fixes

Djjus:

# Changes / Description

- New integrated Size system - Size is selectable in Background tab, and Vs Size in the Melee Assister. Applies Size Combat Modifiers rules from the core book, automatically interacts with Race selection and effects sheet Wounds/Carry weight and modifies melee combat rolls. Showing larger/smaller bonuses as appropriate to a scale of 7 sizes. Only works with talent integration.
- Separated Mele Attack and Opposed modifier, only Attack modifier is shown in Melee Assist window.  So Opposed Parry rolls do not apply Attack only modifiers, and vice versa. 
- Improved Conditions calculation, they now cap at -30 modifier. Add Drunk as a special condition which can stack with other condition modifiers upto max -30. Penalties like Offhand and Armor are now considered in the cap.
- Fixed Carouser RT results not displaying properly
- Fixed Distract RT text not displaying properly
- Fixed issue with Impact +Dmg not displaying when Tiring was also selected and Charing
- Fixed issue with Encumbrance Penalty when Agility was exactly 30 it would result in frozen target display
- Added talents Beat Blade, Beneath Notice, Large/Small, Suffused with Uglu, Super Numerate & Supportive to talent integration
- Added Beat Blade to melee assist actions, and reordered them so the layout cascade better on sheet
- Activating Stun to Strike will now with also activate Melee Called Shot.


Robak132:
@Djjus can you also fix Super Numerate visibility, it has simple integration +1 SL to Gamble and Evaluate.

Djjus:
I did have to add some of the localization tags back, mostly Skills names without the "-" because they are used in the roll template. Please leave those for now.

Robak132:
> I did have to add some of the localization tags back, mostly Skills names without the "-" because they are used in the roll template. Please leave those for now. Psychology wasn't add for some strange reason your last PR just hasn't been applied, its included in this update too though. See what happens.

Maybe we should remove skill names with "-" so they will work for templates and interface?

Robak132:
Also how to recognize localization keys in roll templates. They are not capitalized and they don't use "data-i18n=" attribute. I want to check if any removed key should be reverted.

Robak132:
Maybe we can use size system instead of race in main tab, that's way we could make HP calculations use it.

Djjus:
They interact together Size and Race, so shouldn't be a problem. All HP is calculated automatically.
So selecting race will automatically select correct size/hp, but selecting size will overwrite racial size.

Also you want to do it completely custom too, so Monstrous Human, Tiny Dwarf etc.

Robak132:
By the way, can I enable default global settings for Critical Rules and talents integration from the roll20 campaign page?

Djjus:
Done, also Talents too.

PR #10886, opened 2022-08-04
[BESM4eExtras] v1.8.7 - Bug Fixes

johndav2:
# Submission Checklist
## Required

# Changes / Description

v1.8.7
* Fixed bug where non-weapon attributes would fail to activate.

* Attributes with specialized configurations will now update when specialization is set.  Previously only would set when level was adjusted.

* Ranged Attack will now reflect which weapon it is specialized for.  Would still apply to weapon it was configured for correctly, the weapon configuration just didn't display.

* When Targeted limiter is added to Massive Damage, now shows <not configured> if an enemy has not been defined.

* The function setCombatRoll() no longer crashes if a maneuver is not selected. Previously this would cause some attributes not to activate until a maneuver was selected.


johndav2:
Thank you, David!

John

PR #10910, opened 2022-08-11
AGE System - bug fix & large update

the-Dissilent:
# Submission Checklist
## Required

# Changes / Description

The way the dice were being rolled was interfering with 3D-dice, so I rewrote it with Custom Roll Parsing.
I had to go through hundreds of lines of code, but I made it more streamlined to make future updates easier.
While doing so, I improved the roller to more accurately send information to the chat.
One sheet attribute (in a repeating field) had to be changed, but I wrote a script to preserve user information.
I also added a hidden feature for users to be able to change the headings of fields on the Persona Tab.


LombRicc:
Hi the-Dissilent and nmbradley thank you for your contribution. I noticed earlier that 3D dice was not working with the age system while now is perfect. However I've noticed an odd bug which I'm pretty sure was introduced with this update:
while using the dragon age sheet the spell list in the character tab doesn't show the spell's name and so doesn't allow to press the button and cast it. Oddly when changing the sheet to fantasy age or others they start working as intended again but this introduces a lot of differences with the dragon age sheet to simply switch to them (like adding/changing abilities and so on). I think it's something related to the buttons not showing up when selecting the dragon age sheet. Of course they do work as intended from the magic tab, but it was simply much more convenient to have everything under control in the first tab.
I've tried adding another spell, creating a blank character and gave them a spell, I've even created a blank game just to see if I did mess something up in my game and the result was the same. Since they were working as intended earlier my only other option right now is that a recent update to the system may have broke this feature, which is easily overlooked considering it only affects one of the many possible character sheets under the same age system.
Can you please check this out as soon as you can?
At least if everything is alright to you I can try to find some other solution.

Thanks again for your work and for your attention.

image

the-Dissilent:
Fascinating!
I'm seeing the same error, now that I'm looking for it.
Thanks for pointing it out.
I'll take a look, and it should be fixed with the next update in a week or 2.

LombRicc:
Glad to be of help!

PR #11339, opened 2022-12-31
TheWitcher - Bugfix: Character sheet it's now calculating from true value ins…

gitdopedrodantas:
…tead of current value, as it should be (corebook Physycal Table - pg. 48)

# Submission Checklist
## Required

- [X] The pull request title clearly contains the name of the sheet I am editing.
- [X] The pull request title clearly states the type of change I am submitting (New Sheet/New Feature/Bugfix/etc.).
- [X] The pull request makes changes to files in only one sub-folder.
- [X] The pull request does not contain changes to any json files in the translations folder (translation.json is permitted)

- [X] This sheet has been made on behalf of, or by, the game's publisher.

# Changes / Description

According to the corebook, pg. 48, the Physical Table and other derivate from then (as Enc, Rec, Focus, etc.), the values used to calculate should be setted from the original Statistc values, and not the current values. I change a bunch of then, maybe all. Will bring future updates if I find anything else!

### DIFF
--- a/TheWitcher/witchersheet.html
+++ b/TheWitcher/witchersheet.html
@@ -1897,14 +1897,14 @@
 				            <span class='blue1-skill-span'><input class='prof-main-name' type='text' name='attr_prof_blue1_name' value='1st Skill Name...'>
 				            <select class='prof-skill-choice' name='attr_prof_blue1_choice'>
 				                <option value='0'>N/A</option>
-				                <option value='@{int}' data-i18n="int-u" >INT</option>
-				                <option value='@{ref}' data-i18n="ref-u" >REF</option>
-				                <option value='@{dex}' data-i18n="dex-u" >DEX</option>
-				                <option value='@{body}' data-i18n="body-u" >BODY</option>
-				                <option value='@{spd}' data-i18n="spd-u" >SPD</option>
-				                <option value='@{emp}' data-i18n="emp-u" >EMP</option>
-				                <option value='@{cra}' data-i18n="cra-u" >CRA</option>
-				                <option value='@{will}' data-i18n="will-u" >WILL</option>
+				                <option value='@{total_int}' data-i18n="int-u" >INT</option>
+				                <option value='@{total_ref}' data-i18n="ref-u" >REF</option>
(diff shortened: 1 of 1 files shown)


Stexinator:
I looked at page 48 in CRB and there is nothing written about only taking base stats without improvements or temporary buffs

gitdopedrodantas:
@Stexinator > I looked at page 48 in CRB and there is nothing written about only taking base stats without improvements or temporary buffs

It's not about the buffs! It's considering only the buffs and not the actual attributes! 

Maybe I've misunderstood something, but take a look:

Page 48 has a Physical Table where you can calculate your HP and STA (and others). A Witcher with Body 10 and Will 9 (both "incredible") are currently with only 15 HP, the second lowest score possible. While they should have 45 HP -> (10+9)/2 = 19/2 = 9.5 rounds to 9. A HP of 15 represents a person that would probably die of a single sword attack, maybe two, while a HP of 45 it's more realistic when we're talking about a trained witcher as in the previous example.

Later, on page 57, in the first "Skill Resolution" section, the book describes Geralt facing an opponent. In the example roll, they say Geralt has a Reflex 13, showing us to consider the "max" value, not the "current" one. If we are supposed to consider the "current" one, Geralt would need a max Reflex of 40 to represent the current value of 13! If 13 is "superheroic", then 40 it's absolutely god.

Stexinator:
@gitdopedrodantas the sheet starts with 0 hp. Thus giving you only a third of your real attributes until you put your hp over 1. This lead to some confusions already, so maybe the bugfix should be to set the starting hp to at least 1.

Please check your findings again with a higher hp count :)

Stexinator:
@gitdopedrodantas the sheet starts with 0 hp. Thus giving you only a third of your real attributes until you put your hp over 1. This lead to some confusions already, so maybe the bugfix should be to set the starting hp to at least 1.

Please check your findings again with a higher hp count :)

gitdopedrodantas:
@Stexinator OMG... Bro, you're absolutely right! I'm feeling so dumb right now hahaha really sorry, and happy new year!

PR #11367, opened 2023-01-11
New character sheet for Tesoro y gloria

agviegas:
# Submission Checklist
## Required

## New Sheet Details

Not a new sheet: this PR updates an existing one.

# Changes / Description

Makes the Tesoro y Gloria's character sheet more appealing.


nmbradley:
Hi @agviegas,

Unfortunately, I'm going to have to reject this pull request for the following reasons: 

Firstly, it is only permitted to modify one subfolder when creating a pull request.

Secondly, it is not permitted to delete a sheet from the repository in the way you have for the `Tesoro y Gloria`. It will break the games of the users who are using this sheet.

Please create a new pull request modifying the files of a single folder. If there's a good reason to remove the other sheet, let me know and we can deprecate it gracefully so that it doesn't affect the sheet's users.

agviegas:
Hey @nmbradley thanks a lot for your response, it makes sense. I'm the first dev helping the `Tesoro y Gloria` community, and I noticed that in the past they accidentally created an extra folder that no one was using, so I thought that it would be a good idea to clean that up. If you could deprecate it, will it get automatically removed in the future, or should we do that manually? 

I'll create another pull request. Thanks again. 🙂

PR #11673, opened 2023-03-30
Coriolis Alternative - Bugfix - Fix fonts in CSS

Sh4dd0w:
# Submission Checklist
## Required

# Changes / Description

Changed typeface to sans-serif in some classes that missing it so there's no mismatch in fonts throughout the sheet. While the original version of the sheet is displayed correctly, the mismatch can be clearly seen when adding translation. Probably because the font doesn't have cyrillic alphabet, so instead a default font is used that has different typefaces, unlike the Exo font that is sans serif in itself.

An example:
image

Font on the left:
image

Font on the right:
image

### DIFF
--- a/Coriolis Alternative/CoriolisAlternative.css
+++ b/Coriolis Alternative/CoriolisAlternative.css
@@ -316,7 +316,7 @@ div.sheet-page {
 }
 
 .sheet-card__panel span{
-    font-family:"Exo";
+    font-family:"Exo", sans-serif;
 }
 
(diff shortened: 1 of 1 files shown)


Sh4dd0w:
@mperes Hi there! Sorry to bother you. Thanks for quick approval. I was wondering, when will the new translation I published on Crowdin become available? It's just that most of the translations show they were updated 3 years ago and I'm worried if it's updated automatically ^^

PR #11699, opened 2023-04-09
[Abstract Dungeon] bugfixes for encoding issues

edgeworthe:
# Submission Checklist
## Required

# Changes / Description

Three main things changing here:

1) The character encodings on some of the extended character set
   letters weren't unicode, so display strangely on Roll20. I
   believe I updated them all correctly to unicode versions.
2) Underscores in Markdown set italics, so the underscore in the
   "NPC_card" section was being read as "switch to italics".
   Adding a backslash fixes that, but due to this being in JSON,
   we have to add a backslash for the backslash to ensure it
   makes it through to the Markdown interpreter.
3) English translations for some strings had spaces before
   colons, which is not standard practice in English, so I
   removed them. I left the spaces in the French strings since
   it's apparently standard practice:
   https://mashable.com/article/leaving-space-before-end-punctuation-texting


edgeworthe:
Tagging @Pantoufle167 as the original creator of this character sheet

Abstract Dungeon (FR base)/sheet.json
```diff
@@ -4,116 +4,116 @@
...
     {
       "attribute": "desc2",
-      "displayname": "Description CaractÃ©ristique 2 :",
-      "displaytranslationkey": "attribute description2 :",
+      "displayname": "Description Caractéristique 2 :",
+      "displaytranslationkey": "attribute description2:",
       "type": "text",
-      "default": "Esquiver, Parer, Equilibrer, Basculer, Rouler, Viser, Danser, Attraper, Ramasser, DÃ©sarmer, Nouer, S'Ã©chapper, DÃ©vier, Transpercer, Se retourner",
-      "description": "Change les descriptions pour correspondre Ã  votre type de jeu",
+      "default": "Esquiver, Parer, Equilibrer, Basculer, Rouler, Viser, Danser, Attraper, Ramasser, Désarmer, Nouer, S'échapper, Dévier, Transpercer, Se retourner",
```
edgeworthe, line 95: The apostrophe in this string is currently causing the string to be truncated in the campaignsettings page on Roll20

I think this is a bug on Roll20's side; I tried putting backslashes before the apostrophe, but that didn't fix anything

I don't know of a way to phrase this en français without an apostrophe, though, so I left it unchanged for now

edgeworthe:
Leaving as a draft until I can get in touch with Pantoufle; also messaged them on Roll20

Abstract Dungeon (FR base)/sheet.json
mperes: @wing-it could you weight in on this issue?

nmbradley:
@edgeworthe We're investigating the apostrophe issue.

edgeworthe:
@nmbradley: Thanks for looking into it! It's not a blocker for this PR since the behavior is the same either way, but would be good to at least have a workaround available for it

PR #11802, opened 2023-05-12
HeroSystem6eHeroic - New Feature

nightwalker450:
# Submission Checklist
## Required

# Changes / Description

* Adding Tooltips / Labels to Rolls to assist in macro creation
* Refactoring to simplify updates and calculation handling
* Added Versioning, to handle updating characters as features are added
* Changes to Attack layouts w/ more detail
* Added Shield Bash ability
* Updated Roll Output to color 1's as crit success, and 6's as crit failure to match Hero System
* There should not be any issues with existing characters
  * Versioning is just used to make sure new attributes are updated to what they should be based on existing character


Villain1nGlasses (commented):
I've found the following issues so far:

Armor dialog: The character's PD and ED are not added to the armor's stats in the "total defense" line. Instead the results are equal to zero.

Attack dialog: I think you mean to display currentEND, not endurance. I believe this is a carryover from my previous issue with health-related attributes.

Attack dialog: The STUN multiple shown for killing attacks probably should show (1x, 2x, or 3x). Meaning you can drop the "included" part, but also show "1x" when that is the result of 1d3. Or maybe say something like "(BODY x1)." Kind of a nitpick.

Martial Maneuvers: The "show" buttons don't display the maneuver's details anymore, only the default values.

No issues that would indicate CSS problems.

Villain1nGlasses (approved):

PR #12208, opened 2023-10-02
Nechronica: Updated with translation strings and images

mortaine:
# Submission Checklist
## Required

- [X] The pull request title clearly contains the name of the sheet I am editing.
- [X] The pull request title clearly states the type of change I am submitting (New Sheet/New Feature/Bugfix/etc.).
- [X] The pull request makes changes to files in only one sub-folder.
- [X] The pull request does not contain changes to any json files in the translations folder (translation.json is permitted)

# Changes / Description

Updated to have i18n translation strings for all fields, labels, and images.

`translation.json` contains English translation. Original sheet had Korean language. To conform to crowdin source language requirements (and per discussion with Roll20 Support-- see request 151311 in Helpdesk, or ask Miles), I changed the language to English.

Added ko.json, which contains the original Korean language strings. **PLEASE PUT THIS IN THE** `translations/` **FOLDER** once this sheet has been added to crowdin so that Korean-language users will not have their experience disrupted.

Also moved images from imgur to github and links to images to github.

Did some typo fixes and cleanup, but did not change any attribute names.


mortaine:
Updated sheet.json. Please merge

NorWhal:
@mortaine, your sheet.json line should include all the authors' names separated with commas in one string. You currently have three separate strings, which is invalid. I'm also unsure if we can merge the sheet with your `ko.json`. Our policy says that translation needs to go through crowdin, but that policy is from before a recent upgrade to our translation system. I will confirm for you whether that file can be included in this PR or not.

Once those two things are cleared up, this sheet is good to go!

mortaine:
Thanks. I have fixed the sheet.json file.

For the ko.json translation-- I know translations need to go through Crowdin normally. This sheet was originally submitted in Korean. I do not speak Korean, but had someone translate it into English when I added the translation tags. I put the English into the translation.json file and added the ko.json with the original strings so that existing Korean-language users of this sheet will not have it abruptly change to English on them.

Since I do not speak Korean, I don't think it would be appropriate for me to go through Crowdin and become a Korean translator just so I could submit this one translation file. When I contacted Roll20 Support and hashed it out with Miles, Miles agreed with me that submitting the file separately in this case would be the right way to go about it and the ko.json file would need to be brought into the translations folder manually.

NorWhal:
Thanks for the wait, @mortaine! I had to get my own crowdin permissions set up. 

Since it's the source language and this update was you initializing translations for this sheet, I can take your korean file and upload it to crowdin manually. The reason we don't typically allow uploads of two-letter language files in PRs is that crowdin will overwrite them whenever anyone makes a change to any language for the sheet. I just didn't want your file to get lost and have Korean users suddenly get that same sudden change to English you were worried about.

NorWhal:
@mortaine and it's uploaded!
There are six strings that don't have translations in the file you provided:
- Madness
- Attack
- Add 1 Madness point
- Madness Check
- Character Information
- Roll to attack

NorWhal:
If you just want to provide me the strings for those items in a comment here, I can add them manually

mortaine:
Hi, Nora!

Thanks for merging and reaching out for the translation strings. Three of
these should stay as-is, and three of them can be fixed by renaming the
strings in the ko.json file, as follows:

The Character Information string is embedded in an image in the sheet. The
actual text is only shown as an image title; if/when this gets translated
by a Korean translator, overwriting it shouldn't be a problem, and in the
meantime, there is no loss in functionality (the original sheet has no
image titles).
character-info: Character Information

These two strings do not have a Korean equivalent on the sheet. "Attack" is
literally shown in English on the original sheet. "Roll to Attack" is a
title used on a button image-- again, it should be no problem for this to
be overwritten by Crowdin, and there is again no loss in functionality.
attack: Attack
attack-roll: Roll to Attack

These three are due to renaming the string from "insanity" to "madness" in
keeping with the intent of the game-- I must have missed renaming them in
the ko.json file when I did that rename. The strings are in the ko.json
file, but are incorrectly named. These can be fixed this way:
Rename line 15 insanity to madness
Rename line 45 faildescription1insanity to faildescription1madness
Rename line 61 determineinsanity to determinemadness

Thanks!
--Stephanie

PR #12382, opened 2023-12-14
[Dirty System] Wrapper adjustement

ctotone:
Adjusted the tab layout to increase the length of the text which was a bit short.

### DIFF
--- a/Dirty System/dirty.css
+++ b/Dirty System/dirty.css
@@ -64,7 +64,7 @@ body.sheet-darkmode .selector {
 .charsheet {
 	background-image: url(https://raw.githubusercontent.com/Roll20/roll20-character-sheets/master/Dirty%20System/img/background.webp);
 	background-size: cover;
-	width: 740px;
+	width: 815px;
     border: 2px solid;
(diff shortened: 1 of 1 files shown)


ctotone:
Thx @NorWhal for the validation.

Just a little question. 
Are we allowed to update a local language file in the translation subfolder via a pull request?

I'm asking this because during pull request [#12374](https://github.com/Roll20/roll20-character-sheets/pull/12374) I attached [a gits](https://gist.github.com/ctotone/ec08512db7e8a4c82c69a915620b69a6) which seems not to have been integrated into the crowndin update.

If ever I am allowed to propose a pull request modifying the fr.json then I will take care of that. Otherwise, how should I proceed to request the update of fr.json?

Thanks in advance.

NorWhal:
I just checked and the translations are showing up for French users, though there are a few items that aren't being translated. We're currently undergoing an upgrade to our translation service, so the usual methods for adding new translators are not operational. Until we can get those fixed, you can create a PR with _only_ the `fr.json` file in it, and we will update it manually.

PR #12682, opened 2024-03-06
[Paladin] Overhaul continuation

Greyblade:
Change list 03/12/24 - following pull request blocked due to attribute conflict: 

Corrected Age button
Fixing repeating container bug
Ideals - adding missing romance ideal field
Ideals - changed descriptions to reflect current rules
Squire section was wrongly pointing to the checkbox rather than the actual text field
Annual glory - added missing total, fixed sum calculator
"Knights" attribute (_attributes.knights.forEach(attr => {_)was doubled, changed for "annualglory" attribute that was missing
Fixed sum calculator syntax
Dropped change from attribute "name" to "character_name" - will use sheetworker at a later stage (thanks for the tip)
Modified calculator for derived attribute damage (rounding down as per rules, instead of rounding up)
Added my name as contributor, updated sheet version

### DIFF
--- a/Paladin/paladin.html
+++ b/Paladin/paladin.html
@@ -2366,7 +2366,7 @@ data-i18n="courtesy" name="roll_courtesy" type="roll" value="&{template:rolls} {
 	movement: ['strength', 'dexterity'],
 	hp: ['size', 'constitution'],
 	unconcious: ['total_hit_points'],
-	knights: ['old_knights', 'middle_aged_knights', 'young_knights']
+	knights: ['old_knights', 'middle_aged_knights', 'young_knights'],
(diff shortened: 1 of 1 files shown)


kfroll20:
For existing games, if the names of attributes are changed, the game will lose the data. Please use sheet workers to transfer data if an attribute name needs to be changed.

Greyblade:
Hi Kayla

The attribute "knight" is a mistake it is provoking a bug in the sheet

"annualglory" is the one that is correct.

Is it possible to approve this pull request?

Greyblade:
Hi @kfroll20 & team, see all my comments & edits in the pull request, as per your advice
Thanks a million for your help!

kfroll20:
Hello @Greyblade! 

I see that there is still an attribute name change for `squire_age `. Once that is fixed, we should be able to get this merged :)

Greyblade:
Hi @kfroll20 the squire_age is actually the correct one, the one with checkbox is actually a mistake causing the whole line not to act properly :) there is no checkbox anymore :) 

Can you accept the pull request

PR #12723, opened 2024-03-18
[Paladin] Automating ideals calculation + squire skills default value

Greyblade:
All three ideals for knights (Romance, chivalry, religion) consist in adding up the values of six relevant personality traits.

Rather than doing it manually, I propose this change to automate the calculations and thus making it easier/faster for players.

**Current version (in French)**

image

**Proposed change**

image

Also, adding 0 as default value to all 4 squire basic skills

### DIFF
--- a/Paladin/paladin.html
+++ b/Paladin/paladin.html
@@ -149,7 +149,8 @@
 <div class="row 2autocolumn woman">
 <h2 data-i18n="Gentle Woman Bonus[*](total=80+)"></h2>
 <label data-i18n-title="enter gentle woman bonus" title="enter gentle woman bonus">
-<span class="text-center display" name="attr_gentle_woman_bonus"></span>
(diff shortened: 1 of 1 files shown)


BronsonHall:
Hey @Greyblade 

Definitely see the appeal of this change - I'm not too familiar with the game though, so wanted to double check one quick thing before I merge this.  Is there any way that these three values would wind up being different than their derived calculated sums, even temporarily?  
Reason I ask is that, ofc, automating these fields and turning the inputs to spans will lose player data for anyone who has input totals that break the rules.  If that's not reasonably possible within the limits of the system, then it's not an issue.  Otherwise, you might want to consider implementing an optional setting to auto-calculate these fields, that can be disabled to restore the manual inputs.  

Thanks!

Greyblade:
Hi @BronsonHall , thanks for your comment.

You add up the values to tally 90 points in each, and you also need another independent value (ex : honor at 16+) to qualify for the bonus.

So it's always 90+ with those 6 specific traits for each ideal that you need :)

Hope this clarifies the rules :)

BronsonHall:
Ah!  I see now, thanks for clarifying.  Sound reasoning, I'll merge this in!

PR #12729, opened 2024-03-19
[Paladin] - Gentlewoman autocalc + autoupdate name/character_name + correcting Size mistake in NPC Feast tab

Greyblade:
**ADDING GENTLEWOMAN IDEAL AUTOCALC**
After adding autocalc on ideals for Chivalry, Romantic & Religious, finishing up with the Gentlewoman ideal autocalc, specific to lady characters

**AUTOUPDATE NAME/CHARACTER_NAME**
In order to update name & character_name automatically, without losing the values entered in the @name field in already existing character sheets, I propose this change. This is based on a new hidden attribute called sheet_version. It is functional and changes are done automatically when a player closes & reopens the sheet. Thanks @kfroll20 for the tip!

**FEAST SIZE vs SIZE**
Size is wrongly used in this NPC tab as it is a character attribute, thus updating the value for the character whenever a value is entered in the field, while it's specific to feast size (an event), which is correctly used in the PC feast tab.

Obviously a mistake or omission from the original dev

**"Size" in the NPC feast tab - should be "feast_size":**
image

**See PC feast tab in comparison :** 
image

# Submission Checklist
## Required

## New Sheet Details

- The name of this game is: <   >  _(i.e. Dungeons & Dragons 5th Edition, The Dresden Files RPG)_
- The publisher of this game is: <   > _(i.e. Wizards of the Coast, Evil Hat)_
- The name of this game system/family is: <   > _(i.e. Dungeons & Dragons, FATE)_

- [ ] This sheet has been made on behalf of, or by, the game's publisher.
- [x] This sheet is for an unofficial fan game, modification to an existing game, or a homebrew system.

# Changes / Description

### DIFF
--- a/Paladin/paladin.html
+++ b/Paladin/paladin.html
@@ -149,9 +149,7 @@
 </div>
 <div class="row 2autocolumn woman">
 <h2 data-i18n="Gentle Woman Bonus[*](total=80+)"></h2>
-<label data-i18n-title="enter gentle woman bonus" title="enter gentle woman bonus">
-<input data-i18n-placeholder="gentle woman bonus" name="attr_gentle_woman_bonus" placeholder="gentle woman bonus" title="@{gentle_woman_bonus}" type="text" value=""/>
-</label>
+<span class="text-center display" name="attr_gentle_woman_bonus"></span>
 </div>
 <div class="row 2autocolumn">
 <h2 data-i18n="Religious Bonus(total=90+, Love (God) 16+)"></h2>
@@ -2365,7 +2363,8 @@ data-i18n="courtesy" name="roll_courtesy" type="roll" value="&{template:rolls} {
 	annualglory: ['annual_glory_rewards_traits', 'annual_glory_rewards_passions', 'annual_glory_rewards_skills', 'annual_glory_rewards_attributes', 'annual_glory_rewards_attitudes', 'annual_glory_rewards_chivalry', 'annual_glory_rewards_romance', 'annual_glory_rewards_religion', 'annual_glory_rewards_maintenance', 'annual_glory_rewards_luxury_spendings', 'annual_glory_rewards_holdings', 'annual_glory_rewards_enchanted_items'],
 	chivalry: ['energetic', 'generous', 'just', 'merciful', 'modest', 'valorous'],
 	religious: ['chaste', 'forgiving', 'merciful', 'modest', 'temperate', 'trusting'],
-	romantic: ['chaste', 'forgiving', 'generous', 'honest', 'prudent', 'trusting']
+	romantic: ['chaste', 'forgiving', 'generous', 'honest', 'prudent', 'trusting'],
+	gentlewoman: ['chaste', 'energetic', 'honest', 'modest', 'prudent', 'temperate']
 }
 
 
@@ -2426,6 +2425,12 @@ attributes.religious.forEach(attr => {
 attributes.romantic.forEach(attr => {
 	on(`change:${attr}`, (eventinfo) => {
 		sumOfCalculator(attributes.romantic, 'romance_bonus');
+	});
+})
+
+attributes.gentlewoman.forEach(attr => {
+	on(`change:${attr}`, (eventinfo) => {
+		sumOfCalculator(attributes.gentlewoman, 'gentle_woman_bonus');
 	});
 })
 
--- a/Paladin/paladin.html
+++ b/Paladin/paladin.html
@@ -1,5 +1,4 @@
 <input name="attr_version" type="hidden" value="2"/>
-<input name="attr_sheet_version" type="hidden" value="1">
 <input name="attr_sheet_type" type="hidden" value="character"/>
 <input name="attr_character_type" type="hidden" value="knight"/>
 <input name="attr_feast_type" type="hidden" value="feast"/>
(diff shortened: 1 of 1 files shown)


BronsonHall:
Hey @Greyblade

This is absolutely the right idea for sheet versioning - but there's no need to add a new "attr_sheet_version" attribute, you can totally use the "attr_version" attribute that's right above it! You can link the worker to both fetch and compare "version" instead of "sheet_version".  
Next, you'll want to adjust that worker slightly.  For all users currently, their sheet has an "attr_version" attribute with value "2".  You can use that (alongside the existence of an "attr_name") to check if it's a sheet that was created before you made this update, and thus needs the transposition. 
But as written, this will run for every user every time.  You'll want to, in that "sheet:open", also bump the "attr_version" to a higher value - the new version of the sheet - regardless of whether the sheet had a name attribute or not.  This will ensure that the sheet knows it's been updated to that version, and won't bother with updates it's already made before.   It also lets you make that worker a chain of consecutive updates, to make old sheets compatible step by step.

Let me know if you have any questions implementing this!

Greyblade:
Hi @BronsonHall , that's actually a good idea!

Let me play around with the code, I'll update the request when it's ready.

No rush so it can wait until next Tuesday :)

Greyblade:
Hi @BronsonHall ,

I updated the code, dropped the "attr_sheet_version" for "attr_version", as you recommended. Bumped the attribute value to 2 in the worker, to reflect the version of the sheet. If I bump it higher, it doesn't work however.

BronsonHall:
@Greyblade 

Looks great!  To clarify, you'll want to bump it higher in `setObj`, so that the new version number is set on the sheet in the setAttrs.  You want the update to run once for each sheet, and not run again - this'll make it so that once it runs the first time, it won't re-run the next time the sheet's opened.  
I'd suggest setting it independently of the check for attributes.name.  If they have an outdated sheet, you want to update the number, regardless of whether you need to move attributes around.  
Ex.
````
on('sheet:opened', async (info) => {
  getAttrs(['version','name'],(attributes) => {
    const setObj = {};
    if(attributes.version === '2'){
      if (attributes.name) setObj.character_name = attributes.name;
      setObj.version = '2.1';
    }
    setAttrs(setObj,{silent:true});
  });
})
````

You could also implement this in this way, to allow for future chained updates:

````
on('sheet:opened', async (info) => {
  getAttrs(['version','name'],(attributes) => {
    const setObj = { version: parseInt(attributes.version || '2') };
    if(setObj.version < 2.1) {
      if (attributes.name) setObj.character_name = attributes.name;
      setObj.version = 2.1;
    }
    // Next update --
    if(setObj.version < 2.2) {
      // Perform next update
      setObj.version = 2.2;
    }
    setAttrs(setObj,{silent:true});
  });
})
````

**Comment by Greyblade** [comment]
Hi again @BronsonHall ,

I tried the code you posted in my test game, alas it doesn't seem to work. 
However, to clarify, I want the update to be done whenever someone types in a new name and then closes/reopens the sheet.
Ideally, I'd even like the change to be instant, but alas I found no other way but using character_name

**Comment by Greyblade** [comment]
Hi @BronsonHall,

Happy to answer questions you may have about the update :)

**Comment by BronsonHall** [comment]
Hey @Greyblade 

Ah!  I think I understand now.  

There are two solutions to this problem, and you're using a mixture of both right now.  

- First, you _can_ in fact change that input to use attr_character_name instead of attr_name - provided you make sure that the old data in attr_name gets moved over.  This only has to happen once for each sheet - from then on it'll just use attr_character_name as it should - updates will be instant!
    - This solution is the one that needs sheet versioning, and runs on sheet:open.  You'll check if the sheet is old (with the version attribute), and, if so, set attr_character_name to what's currently in attr_name.  Then, even though there's no longer an input for attr_name anywhere on the sheet, its data will be preserved in the new attribute!
    - This is what I wrote my code for - the piece missing is that the input on the sheet should be for character_name.  (also my `parseInt` should be a `parseFloat` - whoops!)

- However, if you'd like to keep attr_name the same, you can add an `on("change:name")` sheet worker to update the character_name instantly when the name changes

    ```
    on ("change:name", (eventinfo) => {
	setAttrs({ "character_name": eventinfo.newValue || ""})
    })
    ```

    - The only downside here is that this runs just when the name attribute changes, and not inherently when the sheet opens.  
    - So the journal title and the in-sheet value will still be different, until the in-sheet value is changed.  It also doesn't update the opposite way, i.e. updating the journal won't update the sheet.  Thus you'd need the same sheet:open behavior that you've got already, just removing the versioning bit entirely.  

Most sheets that use a name attribute that's not "character_name" do so because they want the possibility of this name differing from the one displayed in the journal.  If that's not a requirement of this sheet (it doesn't seem to be), I'd suggest going with the first solution.  It's the least sheet_workers required to facilitate updates, and will lay the groundwork for versioning in the future.

To implement this, you should be able to use my code, changing "parseInt to parseFloat" in the second example ( you can remove the next update section as well, keeping the setAttrs), and changing the attr_name input on the sheet to attr_character_name.  

```
on('sheet:opened', async (info) => {
  getAttrs(['version','name'],(attributes) => {
    const setObj = { version: parseFloat(attributes.version || '2') };
    if(setObj.version < 2.1) {
      if (attributes.name) setObj.character_name = attributes.name;
      setObj.version = 2.1;
    }
    setAttrs(setObj,{silent:true});
  });
})
```
and
```
<input data-i18n-placeholder="name" name="attr_name"...
```
to 
```
<input data-i18n-placeholder="name" name="attr_character_name"...
```

If that doesn't work for you or doesn't cover what you're intending, let me know - and perhaps take a look at `on(change:` if you want instant updates without losing the attr_name input.

Greyblade:
Hi @BronsonHall ,

Many thanks for your reply, I hadn't had the chance to play with the code today, but I'll run some tests in my test game & see how I fare. 

No rush :) And many thanks again for your help

Greyblade:
Hi @BronsonHall ,

The solution you proposed is working nicely - I made the changes & updated the pull request.

Have a look & let me know if it's fine :)

Many thanks for your help with this :)

BronsonHall:
Hey @Greyblade 

Looks good to go!  Merging now.

PR #12822, opened 2024-04-23
Create a Shaped character sheet for The Wildsea RPG

austinmh12:
# Submission Checklist
## Required

## New Sheet Details

- The name of this game is: The Wildsea RPG  _(i.e. Dungeons & Dragons 5th Edition, The Dresden Files RPG)_
- The publisher of this game is: Quillhound Studios _(i.e. Wizards of the Coast, Evil Hat)_
- The name of this game system/family is: 13th Age _(i.e. Dungeons & Dragons, FATE)_

_Somewhat_, I know a Wildsea RPG sheet already exists, which goes against the ask *Ideally, don't submit sheets for games that already has a sheet.* but it seemed like a better idea to do this than to rewrite the entirety of the other sheet.

- [ ] I have authorization from the game's publisher to make this an official sheet on Roll20 with their name attached.

# Changes / Description

Created a custom sheet from scratch to closely mimic the shape of the Wildsea official character and ship sheets using the new sheet styling and available jquery ( $20 ) to handle updating attributes. The only unimplemented thing is the ability to "burn" an aspect/crew track as the roll20query does not allow for the updating of a class in a repeating item in the way that I was attempting.


BronsonHall:
Hey @austinmh12 

  Yep, the check's failing because it can't find the sheet folder in the `approved.yaml`; but you don't need to worry about that, we add each new sheet to the `approved.yaml` shortly before we merge them with proper folder, game & system names.  That check always fails for new sheets until we set it up on our end - really just a reminder for us.

  That said though, as you did mention in this PR, there is another existing community sheet for this game, which goes against our usual policy.  We traditionally bend this rule in a couple cases, namely when a publisher submits a sheet for their own game, or when a sheet is submitted for a 'version' of the base game that is mechanically incompatible.  (example here - a newer edition of a system is released that changes some rules, but keeps the same title.  We might allow both an existing sheet and a new sheet to exist in this case, as active campaigns may want to keep using the old ruleset.)

  This instance is a bit more tricky - this sheet has significant functionality boosts and visual improvements over the existing sheet, but is made for the same ruleset, mechanically.  The 'one game - one sheet' rule was originally made in the interest of encouraging collaboration and simplifying sheet selection.  But it's been clear in the past that, in cases like these, it can often be _a lot_ more work to make a sheet into an upgrade rather than a stand-alone alternative, due to the added requirement of maintaining existing user data from the old sheet.  Nonetheless, that's definitely the ideal scenario, insofar as the repo is concerned.

  If you wouldn't mind, I'd ask that you take a quick look, and just scope out how much work it'd be to make this an update instead.

  You could effectively replace the whole old codebase with the new, provided that - crucially - all attributes are kept, either being transferred to a new attribute name via an update sheet worker or given the same name on the new sheet.  This includes repeating sections if possible.  We'd also then give a courtesy period for the old author to review and approve the changes.

  If it looks like it'd be too much, or if the two sheets just functionally do not line up, let me know.  Either way, I'm going to discuss this all with the team tomorrow and get back to you.  
  
  Thanks for your patience!

austinmh12:
Thanks for the detailed response on this, I've been doing some comparisons between both the sheets to see what it would take to adapt this version into the existing one in the least intrusive way. It doesn't seem like it would be an egregious amount of work, I would just have to adjust some free text areas to be repeating sections and adjust attribute names. I'll try to reach out to the authors of the original sheet to see what they think and try to move forward from there.

PR #12864, opened 2024-05-06
[MetalAdventure][New Feature] Additional skills to add them when asked from books

TheShinriel:
# Submission Checklist
## Required

# Changes / Description

In the MetalAdventures books, there are many skills that can be learned during the adventures.
However, the current sheet does not allow these skills to be added.
The paper sheets provide empty slots specifically for these cases.

Thus this pull request suggests adding 2 blank skills for each characteristic.

For years, we've been using extra lines for equipment, transfers and even specialities, but every time we have to think that we have a skill.
We often forget that we have an additional skill.

### DIFF
--- a/Metal Adventures/MetalAdventures.html
+++ b/Metal Adventures/MetalAdventures.html
@@ -330,6 +330,24 @@
 						<input type="number" name="attr_Technologie_max" readonly>
 					</div>
 				</div>
+				<div class="row">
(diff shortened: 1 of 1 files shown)


TheShinriel:
Just to be clear I also removed some dead code

```
change:Jeux change:Seduction"
```

PR #12930, opened 2024-05-23
Dedales - sheet.json fix / img URL

ctotone:
I modified the URLs to point to the GH Roll20.

I also modified the sheet.json because it turns out that the name of the html / css / png files listed there did not agree with the real names of the files.


ctotone:
I don't know if it was better to change the contents of the sheet.json file or to change the name of the html/css/png files

Depending on what is most practical for you, I will make the necessary modifications.

Thanks in advance.

PR #13212, opened 2024-08-16
100DOS Mythic by 100DOS Bugfix and New Features

michaelvanweelde:
# Submission Checklist
## Required

# Changes / Description
Single loading and single use weapons list a reload rate instead of single use
Skills share the same issue as above, though it is added to advancements
the Scatter button has a typo Failure instead of Failure
NPC's cannot be dragged in from the bestiary, only from existing entries.
when dragging a bestiary stat block onto an existing sheet, it doesn't update/replace the wounds
Civilians don't have abilities which causes the sheet to search for a non existent array 
Armor Variants that give bonus to wounds show #ERROR! in note portion
Melee weapons have base damage as their pierce and do not include pierce at all
Mythic TOU and AGI are swapped in drop-down menu
Vehicle Mobility Breakpoints aren’t transferred when drag-and-dropping onto a new sheet
Implemented compatibility with Devsheet JSON exporter (Armory user compendium support)


michaelvanweelde:
I believe the line endings might have reverted to CRLF instead of LF in these files though I cant seem to check, if they have I can make a new push with the correct line endings.

kfroll20:
Hello @michaelvanweelde, 
You were correct in noting that the line endings need to be fixed. We can keep an eye out for your update and get this merged then. 

Thanks!

PR #13617, opened 2024-12-17
[New sheet] Anime 5E

VJRolmasters:
# Submission Checklist
## Required

- [x] The pull request title clearly contains the name of the sheet I am editing.
- [x] The pull request title clearly states the type of change I am submitting (New Sheet/New Feature/Bugfix/etc.).
- [x] The pull request makes changes to files in only one sub-folder.
- [x] The pull request does not contain changes to any json files in the translations folder (translation.json is permitted)

## New Sheet Details

- The name of this game is: < ANIME 5E  >  
- The publisher of this game is: < DYSKAMI PUBLISHING COMPANY  > 
- The name of this game system/family is: < D20  >

- [x] I have followed the [Character Sheets Standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) when building this sheet.

- [x] I have authorization from the game's publisher to make this an official sheet on Roll20 with their name attached.

# Changes / Description

ANIME_5E

PR #13638, opened 2025-01-02
[New Sheet] - D&D 3.5 Epic 6 Sheet

BE-Arbiter:
# Submission Checklist
## Required

## New Sheet Details

- The name of this game is: <Dungeon & Dragon 3.5> 
- The publisher of this game is: <Wizards of the Coast> 
- The name of this game system/family is: <Dungeons & Dragons>

- [ ] I have authorization from the game's publisher to make this an official sheet on Roll20 with their name attached.

# Changes / Description

This is a light and modern character sheet for **Dungeon and Dragon 3.5**.
Instead of overwhelming tables full of modifiers, it uses a smart system of generic modifiers that easily adapt to various situations.
This approach keeps the interface clean and user-friendly while still offering all the functionality needed for a D&D 3.5 game.

Features : 
- i18n
- Smart Modifier
- Clean and modern interface


BronsonHall:
Hey @BE-Arbiter 

I can tell you've put a lot of work in on this, but unfortunately we'll have to reject this sheet under our [policy](https://wiki.roll20.net/Building_Character_Sheets#Minimum_Requirements) disallowing new, duplicate community sheets for identical rulesets.  Essentially (while old sheets are grandfathered in to this rule) we no longer permit multiple sheets for the same system unless they are explicitly built for alternate rulesets or variations of that system. 

Though this sheet is visually and mechanically very different from the existing 3.5e community sheet, both are built to serve the same base 3.5e ruleset, so we can't allow them to both exist as distinctive sheets on the site.  

There are a couple reasons for this rule.  First, we do aim, wherever possible, to keep the create game dropdown clean and clear when it comes to determining the correct sheet for the game you're playing. (1 sheet for the ruleset you use, or 2 if there are community/publisher versions)  This helps reduce player/gm confusion and makes it easier for groups to track supported features.  

But second, and more importantly, we try and aim for a collaborative environment in this repo, unifying developer efforts for each system under a single sheet.  Having separate codebases for different sheet layouts or designs inherently means there will be a gap between them when it comes to future improvements from sheet authors.

While we've allowed this in the past, and there are exceptions to the rule, those are generally reserved for when the sheets cover ostensibly different system versions, where groups will not just choose which sheet layout they prefer, but what ruleset they want to use in their game.  

We'd encourage you to consider applying this alternate format as an update to the existing 3.5e community sheet, under an optional user toggle.  I know that's a pretty significant task in this case, as just getting the two sheet layouts to play nicely in a shared codebase won't be easy.  It might be possible to take some elements of this layout and apply them individual as alternate views for sections of the existing sheet?  

I apologize for the inconvenience.  We'll look at our wording on that rule and try and make it clearer for authors in the future what we do/don't allow when it comes to duplicate sheets, so ideally that's easier to identify before putting in the effort to write the alternate sheet.

Please let me know if you have any questions!

BE-Arbiter:
It's not just a different layout. It's also a different vision on how to approach the sheet design.
It won't be adaptable unless I do massive destructive changes in the data of the old sheet, which the community will not be happy about. Especially since most old user are fond of their excel sheets.

There is complains about roll20 obscureness in the sheet validation, which I dissmisses as unfounded rumour; This shows that not only it is true, it is also unfair ; The I don't know how much 5e clone sheet (with small layout changes) have been (and still are) accepted because the name and logo is not the same.

BE-Arbiter:
There is not even far too look about it : 
https://github.com/Roll20/roll20-character-sheets/pull/13617
This new sheet is a clone of the 5e sheet, with less feature and nothing you can't do with the standart sheet. The only thing that changes is the layout and the logo. 

There are real functional differences from the 3.5 sheet in this PR, which justify a real impossibility to "just integrate in the actual sheet". 

@BronsonHall

BronsonHall:
I totally understand.  The very fact that both sheets differ so significantly in data and sheet workers makes merging them impossible without either a ridiculous amount of conversion workers that'd further complicate the codebase, or isolating specific, smaller elements from this sheet that can be adapted to respect the old sheet's structure.  But, as you pointed out, the sheets differ fundamentally in approach as well, far beyond a simple layout toggle.  

You're also correct - there are a number of 5e-adjacent (and even closer to clone) systems that do have specific sheets permitted on Roll20, like Anime 5e.  But the primary reason these variant sheets are allowed has everything to do with the fact that their systems themselves are different from base 5e on paper, even if these variations are not all mechanically present in the sheet itself.   

It's certainly true that a group could play Anime 5e using the existing base 5e sheet with little difficulty - or at least comparable to playing with the bespoke Anime 5e sheet.  But there are, generally speaking, differences between these variant rulesets and base 5e. (Otherwise WotC would have something to say about it)  While initially the sheets may only vary in terms of appearance or layout - and it's undeniable that the Anime 5e sheet is WAY closer to the existing base 5e sheet than your sheet is to base 3.5e - we still allow both sheets to exist, so they can be developed independently as befits their specific ruleset.

[Anime 5e](https://www.dyskami.ca/anime5e.html) for instance has a number of rebalance changes and rule variations compared to base 5e.  For those reasons, and because it's been published as a distinct product, sheet authors may see reason to implement those changes on the Anime 5e sheet, where they do not belong on the 5e sheet, or avoid future changes to the 5e sheet that are irrelevant to Anime 5e.  

That specific sheet also falls under the other exception to the rule - if any sheet is created or approved by the system's publisher themselves, it's considered official and always allowed, even if there's already a community sheet for the exact same game.  There are certainly 5e-adjacent games that have non-official sheets, and it's not a requirement, but it does underline another reason for this rule's exceptions: publisher sheets might include character builders or planned future features, where many of the larger differences between a variant ruleset and the original may be most visible.  It also allows the two sheets to use different, separate, default compendiums.  

In addition, while users may figure out they can make the existing 5e sheet work for Anime 5e, that's not immediately obvious, and a user who knows the systems are different may assume the variant ruleset is not supported on Roll20 at all.  This is understandably a concern for the publisher, who also wants to tailor the user's experience on a visual level to their intended ruleset.  

But again, the primary reason we allow these kinds of sheets is the fact that the ruleset they support is explicitly different, regardless of whether or not the sheets themselves have any major variations beyond the nominal and visual.  If this sheet was made for a variant ruleset for 3.5e, published distinctly, there would be no issue, even if the sheets were much closer in design and functionality.  

That's not to say I don't see the use case for these alternate sheets.  I know many players who might prefer playing 3.5e with a sheet that behaves and looks like this one compared to the community sheet.  However, that's a tricky line to draw - before this rule was enforced, some systems wound up with numerous identical variant sheets, all of which could be construed as the 'correct' sheet to use if you're playing that system.  While there may be good reason for a group to prefer one over another, it made it much more difficult for any of the sheets to receive proper maintenance when their player base was split.  Each sheet would grow apart from one another, until some sheets supported features that others did not, and all could be said to be missing some level of functionality.  

In truth, I don't have a great recommendation for ways to get this sheet onto the site - merging the sheets would be a headache and respecting data of existing games is a requirement.  Going forward, you can always ask us directly whether an sheet would be approved before you begin work on it - but again I think it's a must for us to clean up that language in the wiki and elsewhere to make it clear that the policy lies with the rule system, and not the sheet.  Authors shouldn't be expected to have to ask on a case by case basis or regularly go through what is clearly a extensive development process, only to have the sheet rejected.  That's on us, the guidelines for whether a sheet can exist should be made crystal clear to avoid such frustration.

You're absolutely not the first to raise this issue, it's been discussed internally many times over the years, and this isn't even the first 3.5e alternate community sheet we've had to reject.  I have raised the discussion around this rule once more, and it's possible that, as we improve our support for multiple sheets in the same VTT game, we could relax or adjust this policy at some point in the future.  But for now, this policy is enforced as it stands, and we will reject community-published sheets for identical rulesets to existing sheets.  

I sincerely apologize for the time you've lost on this sheet, and I hope you're able to get some use out of it as a custom sheet for personal games. Again, please feel free to contact us in the future for any clarifications around this or other sheet validation rules and policies.

BE-Arbiter:
So, if this sheet was for... like... Sword of sorcery's World of warcraft game (3.0/3.5 derivative), it would have been accepted because different logo and different name?
What about E6 play, which is the primary target of this sheet and differs mecanically from base 3.5 is it not a correct distinction?

BronsonHall:
@BE-Arbiter 
You're correct there - if it's for a distinctive rebranded publication that uses variant rules, and there's a suitable argument that a different sheet is necessary (or would be necessary down the line) to support those rules, it's a-ok. Warcraft RPG would be a good example.

In fact, you're on to something there with E6.  Epic 6 winds up mechanically different from base 3.5 (even if it's the same game for 6 levels), and I'd agree that the existing sheet doesn't support campaigns using those rules as well as they could.  

You would want to rebrand and rename this sheet, folder and PR somewhat, so it can be clearly presented as a sheet designed for E6 play rather than base 3.5, of course.  But that'd indeed be what'd we need to allow both to coexist!

If you want to make those changes on this PR, I'll re-open it and throw a blocked tag on it for now.  If you @ me when that's all done, I'll try and merge it then, otherwise we'll get it when we review PRs again on tuesdays/thursdays.

BE-Arbiter:
Thanks, Will do that ! :)

BE-Arbiter:
@BronsonHall The changes have been done, can you tell me if there is anything else needed?

BronsonHall:
All looks good!  I'm resolving an internal issue with our sheet management tool, but will merge today.
Greatly appreciate your flexibility and understanding here.

BE-Arbiter:
likewise here, thanks for the positive discussion and professionalism

PR #13829, opened 2025-03-04
PF Community

vince-roll20:
- Updated dev scripts. HTML is now minified for production. Minifying the HTML may give a slight performance. YMMV
- Found/fixed a few additional syntax errors that were causing HTML minification to fail.
- Fixed Attack/CMB button text from overflowing.  Added padding to roll template table category headings.

# Submission Checklist
## Required

- [x] The pull request title clearly contains the name of the sheet I am editing.
- [x] The pull request title clearly states the type of change I am submitting (New Sheet/New Feature/Bugfix/etc.).
- [x] The pull request makes changes to files in only one sub-folder.
- [x] The pull request does not contain changes to any json files in the translations folder (translation.json is permitted)

## New Sheet Details

- The name of this game is: <   >  _(i.e. Dungeons & Dragons 5th Edition, The Dresden Files RPG)_
- The publisher of this game is: <   > _(i.e. Wizards of the Coast, Evil Hat)_
- The name of this game system/family is: <   > _(i.e. Dungeons & Dragons, FATE)_

- [ ] I have followed the [Character Sheets Standards](https://wiki.roll20.net/Building_Character_Sheets#Roll20_Character_Sheets_Repository) when building this sheet.

- [ ] I have authorization from the game's publisher to make this an official sheet on Roll20 with their name attached.
- [ ] This game is not a traditionally published game, but a copy of the game rules can be purchased/downloaded/found at: <   >
- [ ] This sheet is for an unofficial fan game, modification to an existing game, or a homebrew system.

# Changes / Description

PR #13834, opened 2025-03-04
Revert "PF Community"

NorWhal:
Reverts Roll20/roll20-character-sheets#13829
Per the request of @vince-roll20


vince-roll20:
Thank you.

vince-roll20:
100% sure this was a "format on paste" setting/issue in my local vscode.  I copied my latest html from /prod/index.html (which works in my custom sandbox without issue) and pasted that into the sheet's root html file; pathfinder_community.html for the PR. Looking at what was actually committed to the repo, I can see that it was not minified html like my production code and had other issues as well.  Smashing my head now...  I'll fix my local setup and re-submit Thursday.

PR #14144, opened 2025-06-23
3DX: Add new rolls. Layout rework. Problem fixes.

Vitones91:
Added new rolls for evasion and luck, also added new styles for better consistency and fixed some layout problems. Added new inputs so players can better elaborate their character skills.

### DIFF
--- a/3DX/3dx.html
+++ b/3DX/3dx.html
@@ -828,3 +828,4 @@
         </div>
     </div>
 </rolltemplate>
+<!-- END OF SHEET -->


Vitones91:
I don't know why github is pointing a conflict and if there is really a problem, i just added 2 new roll templates at the end, and its pointing that there's a problem with it, but i've double checked it and can't really find any problems. I thought at first it was missing an empty line at the end, but even after that it states that there's a conflict.

PR #14299, opened 2025-08-31
Update Shadowdark-Unofficial sheet to add option for Auto Roll Damage and make repeater forms into dialogs and fix typos

nothingisnecessary:
# Submission Checklist

## Pull Request Title

## Pull Request Content

- [x] The pull request does not, _without express prior permission from affected parties_, include any material that could be considered to infringe on a Publisher's intellectual property rights, such as logos, images, rules text or other rules content.

# Changes / Description

- Add Auto Roll Damage option
- Update UI for Repeater item forms to use a dialog instead of pushing content around when it opens
- Fix typos in legal text
- Add sword emoji to left side of damage rolling hyperlink because some users were confused about how to roll damage
- Updated mobile property to try to mitigate the roll20 app warning about sheet not working well with mobile (it seems to work) and tested on my pixel phone: Some font issue still, but roll20 wiki has details and I might or might not try to fix later - for now it shows letter "t" but the rolling works. Also, if you enable the new Auto Roll Damage option, the rolls actually work on phone - output rolls and damage properly. (Without this option, the damage link does not work on mobile). Mobile still has known issues, but is actually fairly decent experience now on phone for Shadowdark!
- Add default options for roll type, whisper type, and automatic damage rolling

Known issues or feature ideas:
- mobile experience is still kind of lousy and doesn't load the dice font
- readability and contrast for vision impaired may not be great
- still no option for disabling automatic level/xp feature - might do that in future, but if you hate it: use a custom counter in the meantime
- a requested feature was to allow individual attack/spell to override the roll with advantage option (for wizard casting magic missile) but I may not add that because it takes too much time to test this stuff, sorry. However, if you always leave "roll with adv" on, then it is pretty easy to just ignore the high/low/last die rolled in my opinion. my design philosophy was: keep it simple and flexible, not get bogged down in individual feature requests which can easily be solved already.

**Made with <3 by CyberSasquatch**

### DIFF
--- a/Shadowdark-Unofficial/Shadowdark-Unofficial.css
+++ b/Shadowdark-Unofficial/Shadowdark-Unofficial.css
@@ -57,6 +57,12 @@ input::placeholder {
   flex: 1 0 50%;
 }
 
+.sheet-sd .sheet-form-body textarea.attr_atk_desc {
+  height: 200px;
+  min-height: 80px;
+  max-height: 300px;
+}
+
 .sd-row {
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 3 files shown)


nothingisnecessary:
Tested using sheet sandbox on desktop and mobile. Ship it! :D

Thanks!

EDIT: I forgot that the sheet.json is on front page and not in the sandbox app. I made a fix and verified and pushed

PR #14339, opened 2025-09-11
Revert "Revert "[Draw Steel] Updated: NPC sheet visual and functionality improvements""

davidroll20:
Reverts Roll20/roll20-character-sheets#14338


PrimalZed:
Thank you @davidroll20 
@Zen-Raven is in the same discord channel I was using to coordinate with others. I'm not expecting to personally continue much on the sheet.

This seems fine to me, except for the seemingly unused `""` and `"move"` keys in translation.json.

PR #14878, opened 2026-03-11
Update agon.css

akibi90:
# Changes / Description
Adjust the position of the character sheet


NorWhal:
Hi @akibi90!

Given that this is an official sheet, we need the publisher/author, @fredhicks to sign off on this before it can be merged.

Best,
Nora

PR #15376, opened 2026-08-14
[D&D 4E] New Feature: Manual power reordering

twmccart:
# Submission Checklist

> [!WARNING]
> Submission Checklist
> Failure to complete this checklist in its entirety will result in your Pull Request being dismissed. _Deleting parts of this template (except the new sheet section for existing sheets) counts as failure to complete it_. If you have any questions, please feel free to create an issue.

> [!NOTE]
> Draft Pull Requests
> If you are unclear about any of the rules regarding the creation of character sheets, or need assistance from the Roll20 team, please feel free to create a Draft PR and request feedback. We'd much rather provide assistance than reject a PR.

## Pull Request Title

Please format your pull request in the following way: `[Sheet Name] Change Type: Description`. For example: `[D&D5e] New Feature: Adding dragons to dungeons`.

## Pull Request Content

- [x] The pull request does not, _without express prior permission from affected parties_, include any material that could be considered to infringe on a Publisher's intellectual property rights, such as logos, images, rules text or other rules content.

# Changes / Description

Opening as a **Draft**: the changes have not yet been exercised in a live Roll20 game, and I would rather have that confirmed (and take any feedback on approach) before asking for a merge.

## The problem

Powers on this sheet cannot be rearranged at all. The cause is structural rather than a missing control: powers are 100 hardcoded blocks (`power-1-*` … `power-100-*`, 18 attributes each), and Roll20's native drag-to-reorder only exists on `<fieldset class="repeating_...">`. This sheet already uses repeating sections for inventory and feats, just not for powers.

Converting powers to a repeating section is the real fix, but it needs a one-time data migration and changes what external macros resolve to, so it is not something to slip into a small PR. This change makes the existing slots reorderable in place instead, with no migration and no change to any stored attribute name.

## What this adds

**Up/down buttons on every power card.** A sheet worker swaps the two slots' contents. The slot numbering never changes — only what sits in each slot — so existing characters need no migration.

The subtlety worth reviewing: three of the fields (`attack`, `damage`, `macro`) store their own slot number inside their text, e.g. `/me uses @{power-3-name}`. A naive value swap would leave a power in slot 2 rolling slot 3's attack and damage, so those three are reindexed as they move. Hand-edited macros keep their wording, and a macro that deliberately references some *other* power is left alone.

**A "Sort by Usage" button** above the cards, grouping filled powers as At-Will → Encounter → Daily → Item and pushing empty slots to the end. The sort is stable, so powers sharing a usage keep the order the player arranged by hand, and unrecognised usage values sort last rather than being dropped. It has no undo, which the label next to it says.

**A cleanup commit** removing a commented-out copy of a single power card that sat above the generated ones — 235 lines parsed by every client on every sheet load, existing only as a reference for regenerating the cards. It had already gone stale (it described a card without the new buttons, so regenerating from it would have silently dropped them). Replaced with a short comment recording what the cards are and which fields embed their slot number. No behaviour change; git history keeps the original.

## Known limitation, documented in the code and README

`power-<slot>-*` names a **slot, not a power**. A token action bound to `@{power-3-macro}` keeps resolving after a reorder, but it now rolls whichever power moved into slot 3 — so a button labelled for one power can quietly roll another. This is inherent to reordering fixed slots: while the handle is a position, moving powers changes what it points at. Players who keep token actions for their powers need to re-check them after rearranging, and the README says so.

## Testing

No Roll20 game has run this yet — hence the draft. What has been checked:

- The sheet worker logic was extracted from the sheet as it ships and exercised against a stand-in attribute store, covering: reindexing on move (including that `power-1-` never matches `power-10-`), a byte-for-byte round trip when a power is moved down and back, hand-edited macros surviving a move, sort grouping and stability, sorting an already-sorted or empty sheet writing nothing, and no power being lost or duplicated by a sort.
- Structural checks on the generated markup: all 100 cards received their buttons, power 1 has no up button and power 100 no down button, the card count and div balance are unchanged, and every new `data-i18n` key resolves in `translation.json`.

I would particularly welcome a second opinion on whether the up/down button approach is worth having at all, versus going straight to a repeating section. I have a potential conversion on a separate branch and am happy to bring that instead if maintainers would rather take the migration in one go.

## Files changed

Only `D&D_4E/`:

- `D&D_4E.html` — buttons on each card, sort control, sheet workers, dead template removed
- `D&D_4E.css` — styling for the new controls
- `translation.json` — new English keys (source file only; nothing under `translations/`)
- `README.md` — documents reordering and the slot-reference caveat


twmccart:
@scarletcs @vectorjohn @iscribble @nogileyelsew @J-Knopf @pigalot I would love your feedback or approval, especially if you are willing to test. I don't currently have a Roll20 Pro account.

nmbradley:
A word of warning: if this sheet cannot be tested, then we cannot in good conscience release this to users.

twmccart:
Yeah if i have to I'll get a month of Pro to test it first.

D&D_4E/D&D_4E.html
```diff
@@ -1926,242 +1926,24 @@
...
-                        <div class="sheet-headfix"></div>
-                        <input type="radio" name="attr_power-[number]-useage" value="At-Will" class="sheet-colgreen" checked/><label data-i18n="at-will">At-Will</label>
-                        <input type="radio" name="attr_power-[number]-useage" value="Encounter" class="sheet-colred" /><label data-i18n="encounter">Encounter</label>
-                        <input type="radio" name="attr_power-[number]-useage" value="Daily" class="sheet-colgrey" /><label data-i18n="daily">Daily</label>
-                        <div class="sheet-headcolor"></div>
-                        <div class="sheet-bodyfix"></div>
-                        <div class="sheet-statblock">
+				<div class="sheet-power-sort-row sheet-tab-powers">
+					<button type="action" name="act_sort_powers" class="sheet-power-sort" data-i18n="sort_powers_by_usage">Sort by Usage</button>
+					<span class="sheet-power-sort-note" data-i18n="sort_powers_note">Groups filled powers as At-Will, Encounter, Daily, Item. Cannot be undone.</span>
```
BronsonHall, line 1931: I'd advocate for a stronger warning here, possibly a popup or more noticeable text, carrying a warning about token actions and off-sheet macros, as per the readme, given usual sort behavior wouldn't break anything, and it can be a headache to fix once broken.

D&D_4E/D&D_4E.html
```diff
@@ -27727,6 +27707,116 @@
...
+    Three of the fields store the slot number inside their own text, so they are
+    reindexed as they move - otherwise a power moved into slot 2 would keep
+    rolling slot 3's attack and damage.
+    ======================================================*/
+
+const POWER_COUNT = 100;
+const POWER_FIELDS = ['name','action','range','level','used','useage','toggle','mod','def',
+    'weapon-attack','weapon-damage','weapon-num-dice','weapon-dice',
+    'attack-misc','damage-misc','attack','damage','macro'];
+const POWER_INDEXED = ['attack','damage','macro'];
```
BronsonHall, line 27734: Question here - do attack and damage need to be here?
Macro makes sense, as the user will almost certainly change that text themselves, but attack and damage are pretty statically just the roll formula, which doesn't need to change at all when powers are sorted - both are calculated off of other fields you are changing, so unless I'm missing something, it's going to always leave the attr exactly the same.  
I think you can safely remove attack and damage from both power fields and indexed, and just leave them alone, only moving attack-misc, the weapon fields, and the other things the user will actually change.

D&D_4E/D&D_4E.css
```diff
@@ -459,6 +459,53 @@
     margin: 0;
 }
 
+/* Reorder buttons, two across in the puny cell that opens each card. */
```
BronsonHall, line 462: Looks like these are getting kicked to two lines, putting the sort down button behind the roll, at default and smaller sheet widths:

Image
to
Image

Might want to make them just a little smaller, or move them somewhere else - thoughts on putting them to the side of the Settings button on each power?

BronsonHall (changes requested):
Hey @twmccart 

While I'm not an original author on this sheet, I'm pretty familiar with it, (and would love this for my 4e games) so I went ahead and gave it a review + test on sandbox.

I have a couple questions / some feedback, but it seems to work well enough on both auto and manual sort without losing data.  Insofar as your questions in the PR description, my two cents as a repo maintainer:
- I'd agree that a repeating section would be the ideal, but... yeah, that's a massive and much riskier update for a sheet that is not under active development - but has an existing userbase.  If you wanna go forward with it, I'm happy to review as much as I can in my free time, but imo this reorder approach is plenty for now.  There's also really no way for a repeating section conversion to not affect off-sheet macros and token actions destructively, so ship might have sailed here.
- I'm also gonna echo what Nic mentioned above, we can't support significant sheet changes that aren't tested in a sandbox by the author, or supported in development by someone who has tested them in a sandbox.  If you're planning on doing more fixes/features on this sheet or in this repo, I'd strongly advise considering the Pro account if you can't get in touch with the existing authors for a sheet - very tough to accurately test for quirks of the legacy sheet system without running it through a sandbox yourself.

I'm leaving this PR as blocked for the moment, but feel free to mark it as Ready for Review once you've got the chance to take a look at this feedback, and we'll get it merged!

PR #15406, opened 2026-08-25
HOTFIX | Symbaroum Backgrounds

nmbradley:
Supersedes #15401


litusb:
There's still one background that fails (when you roll anything, on the chat log there's still one background that shows a not found img)...
https://www.designbolts.com/wp-content/uploads/2013/02/Rough-Grey-Tilable-Pattern-For-Website-Background.jpg
The URL and the image exists but it's not loading well...
image

PR #15412, opened 2026-08-26
[GeneFunk 2090 Revised] New Sheet: Revised official sheet by publisher CRISPR Monkey Studios

crisprmonkey:
This PR adds a new, fully revised character sheet for GeneFunk 2090, submitted by the game's publisher.

I'm James Armstrong, founder of CRISPR Monkey Studios and publisher/copyright holder of GeneFunk 2090 (Roll20 user id 645619). This revised sheet was built in-house and includes a guided character builder (genomes, classes, backgrounds, hacks) - as the copyright holder, we authorize the inclusion of this character creation content - plus drag-and-drop integration with the GeneFunk 2090 Compendium.

The existing "Genefunk 2090" folder is intentionally left untouched so current campaigns are not disrupted; existing games remain on the old sheet unless their GM opts to switch. This PR only adds the new "GeneFunk 2090 Revised" folder.

A confirmation email supporting this sheet as official has been sent to licensing@roll20.net.


nmbradley:
@crisprmonkey I need to request that you strip the comments out of the published version of this code -- if you need to have a secondary working version with the comments in, that's totally fine, but the version we serve has to come under a size threshold, and this is too large for us to serve.

I'll also note that this intersects with our newly published [AI Policy](https://github.com/Roll20/roll20-character-sheets#ai-policy) that covers submissions to this repository. I'll direct your attention toward that while we wait for this code to be amended.

crisprmonkey:
@nmbradley Thanks for taking a look! The comments are stripped out now, so down to 1.68 MB.  If 1.68 MB is still over the limit, what size should I aim for? I didn't find any specific size limits when I was looking through the wiki. I've read that some sheets use minification to reduce file size, but my understanding is that it doesn't necessarily improve performance. Thoughts? 

And thanks for the AI policy heads-up, newly published indeed! As far as I can tell, we're compliant: I built the sheet with Claude Code assistance, but the design, content, and art are mine (I wrote the game), and I tested it with a group using the files as a custom sheet.

PR #15414, opened 2026-08-26
[Umbral (Eleina)] Sheet Update: equipment, armour and Encumbrance

serydjin:
> [!NOTE]
> Unsure about something? Open a **Draft PR** and ask — we'd rather help than reject.
>
> Adding a **new sheet**? Use the new-sheet template instead: run `gh pr create --template new_sheet.md`, or append `?template=new_sheet.md` to the compare URL.

## Title

Format: `[Sheet Name] Change Type: Description` — e.g. `[D&D5e] Bugfix: Fix spell slot rollover`

`[Umbral (Eleina)] Sheet Update: equipment, armour and Encumbrance`

## Checklist (all required)

- [x] Title follows the format above
- [x] Changes are limited to a single sheet folder
- [x] No changes to `translations/*.json` (the sheet's own `translation.json` is fine)
- [x] No publisher IP (logos, art, rules text) without documented permission

## Description

Brings the sheet in line with the current rules (sections 14 and 15).

- Equipment is now a repeating section with Encumbrance per item, replacing the free-text box. Carrying capacity is 4 + Vigour, computed by a sheet worker, with a gauge showing carried vs. capacity and an "Encumbered" warning when exceeded.
- Carrying gear raises that capacity by more than it occupies: two extra columns say how much it adds and whether it rides on the character or on a mount. Only the best of each counts, and the mount one needs a mount listed.
- New "Compania" repeating section for mounts, pets and helpers. They carry no Encumbrance.
- Armour gets fields for the piece and the shield, and the sheet warns in red when the chosen protection step asks for more Vigour than the character has.
- Wording fix and second tab renamed; preview image regenerated to match.

No new external resources, no API scripts, still CSE (legacy: false).


Eldritch-Oliver:
Hi @serydjin, unfortunately I am going to have to block this PR in it's current state, as there is no attempt at migrating the removed equipment textarea's value into an equipment item, resulting in users losing data that they put on the character sheet.

This migration does not need to be 100% accurate, you could do something like just making a single repeating row with the name of the equipment row as the previous textarea value, or something like that. This will ensure that users who are using your sheet do not lose their data.

serydjin:
Thanks @Eldritch-Oliver, you were right and the migration is now in.

On `sheet:opened`, a character that still carries the old `equipo` textarea gets its value copied into the new repeating section: one `repeating_equipo` row per written line, with encumbrance 0 — guessing how much each item takes up would be inventing data. If the text has no line breaks it all goes into a single row, which is the fallback you suggested.

Two details on the safe side. The old `equipo` attribute is **not** deleted: it stays on the character, invisible but intact, in case anything goes wrong. And a `version_hoja` attribute keeps the migration from running a second time. Lists longer than 20 lines are capped at 20 rows, with the remainder joined into the last one so nothing is dropped.

I ran the worker against a stub of the sheet-worker API for six cases — multi-line value, single paragraph, empty sheet, re-opening an already migrated sheet, a 35-line list, and a sheet already flagged as migrated — and it behaves in all of them. The README has a short note for existing users as well.

Ready for another look whenever you have time.

PR #15463, opened 2026-09-13
[GeneFunk 2090 Revised] Bugfix: Chat card number styling on served build + translation keys

crisprmonkey:
- [x] Title follows the format above
- [x] Changes are limited to a single sheet folder
- [x] No changes to `translations/*.json` (only the sheet's own `translation.json`)
- [x] No publisher IP without documented permission (submitted by the publisher, CRISPR Monkey Studios)

## Description

Three fixes, two of them only visible on the served build (they don't reproduce in custom-sheet or sandbox testing):

1. **Chat cards**: ability checks, attack rolls, and saves rendered without the box around their numbers and at the wrong size on the served sheet. Cause: the build pipeline drops style attributes that contain roll template conditionals ({{...}} inside style="..."). All style attributes are now pure CSS, with conditional crit sizing expressed as conditionally-emitted inner spans. Also switched numeric font-weight values to keywords and removed box-sizing, both of which are altered by the pipeline's CSS serialization.

2. **Translation keys**: translation.json now covers every data-i18n key the sheet uses — the missing `level`/`damage` keys rendered as red bracketed key names on the served build; abbreviated uses got their own keys.

3. **Sherlock genome**: now grants its Investigation skill proficiency as printed. Also removes a sheet.json instructions sentence that referenced the previous sheet.

Verified against our full test harness.


sclyonsburke (approved):
Looks good

